PR #260, opened 2014-12-05
Italian Translation + improvements

gablau:
I realized that the translations in Italian were not correct. so I began to translate and to make some code improvements: strings without translations, missing translations of form fields select / option.
Complete Italian Translation of User, Admin Modules
Partial Translation: Feed, Dashboard (missing menu) , Vis (missing multigraph javascript strings) Modules
To Do: My Electric, Node, Input Modules

Other improvements
-User Module - Language selector
user-lang-selector

-Vis Module - Object name and options translations
vis-object name and option

-Vis Module - Multigraph Name
vis-multigraph name

- Dasboard - I realize a script for build the dashboard_langjs.php file. the script (/script/dashboard_langjs_builder.php) extract the "_Tr" string from javascript files then manually copy and paste in dashboard_langjs.php

- Other
of course we would rebuild the file ".po" and re-translate all other languages ...


TrystanLea:
grazie! Gablau

PR #275, opened 2014-12-29
Umlauts in input, feed names

kanflo:
I noted that umlauts and other special characters in feed/inout names would be discarded.
Also, some of the API calls under Modules/input did not return proper JSON.


kanflo:
I noted a problem with this patch, I'll resubmit it when I fixed it.

PR #278, opened 2015-01-04
Reduced login form size and some code formating

ildemartinez:
Changed the login form deleting the labels and show it text into the input box....maybe interesting because the size is reduced and load faster but not necesary if not interested :)


TrystanLea:
if its ok I think i will keep this as it is for now, but thankyou!

TrystanLea:
Hello Ilde, I've moved this pull request to the Extended feature set branch and will merge there once I sort the conflicts, thanks

Issue #233, opened 2014-08-06
Suggestion for new process action


aleszuin:
Hello, 
after purchasing the bundle of Open Energy Monitor  i'm learning emoncms;

Meanwhile, I want to congratulate because I find it very useful and easy to configure. 

In my experience I have found a lack of use so that I explain problem and solution i found: 

There should be a different solution give me to know, or possibly 
 please take the introduction of the change in the official trunk of the project. 

I am not aware if the processing can be activated from the feed rather than inputs; 
each chain input may alter the value so it becomes no longer available; 
you can't enable parallel processing of the same input; 

I solved this way: 

need 3 parallel process:
input  + -> process1 (input altered, such as "allow negative") 
          + ---> process 2 
          + ----> process 3 

is not possible,  so: 

input + -> process1 -> restore input -> process 2 -> restore input ----> process 3 

and this is my modification to the file "process_model.php" 

(...) 
         $ list [35] = array (_ ("Last Feed value"),ProcessArg::FEEDID,"last_feed_value",1,DataType::REALTIME,"Misc",array(Engine::PHPFIWA,Engine::PHPFINA,Engine::PHPTIMESERIES)); 

(...) 
     // ------------------------------------------------ -------------------------------- 
     // Simulate paralell processing of same input: give input back so chain can restart 
     // ------------------------------------------------ -------------------------------- 
     public function last_feed_value ($ feedid, $ time, $ value) 
     {
         $ last = $ this> feed-> get_timevalue ($ feedid); 
         $ last = $ value2 ['value']; 
         return $ value2; 
     } 

i use new action "Last feed value"  to restore the input ( saved on feed  as first action )

Thank you and congratulations again 

Alessandro from Italy

chaveiro:
Can be Closed as implemented.
See https://github.com/emoncms/emoncms/pull/280#issuecomment-71182208

PR #280, opened 2015-01-23
Smart processlist, conditional procs, schedule module + Bugfixes

chaveiro:
This is a major update with much bugfixings i made while was finding them.
I fixed and enforced the same best practices used on the core emoncms development.

Changes:
New scheduler module. Info here: http://openenergymonitor.org/emon/node/10019

New process_models functions
 - if_schedule_true
 - reset2original

Added new process list configuration to support scheduler

Bug fixing on input view and input processlist javascript:
 - fix processlist.js state machine and update events trigering
 - usage of bootstrap elements instead of 'hammered' hardcoded styles
 - hard coded css dependency remmoval
  
Bug fixing on tables.js :
 - fix events trigering during edit
 - fix table not fast redrawing on group min max

Uncomented line to display dashboard list view

TODO:
Refactor node module:
  Its a copy of view input module with custom changes and pour usage of reusable code.
  Writes a long json string to node bd on every post data. Bad performance.
  Better to make this module only an API pre-processor for input engine and use a new input process to 'work' the raw data.
  
Translations for new scheduler module


aleszuin:
hi, 
reset 2 original is something similar to my post:
new process action #233  - https://github.com/emoncms/emoncms/issues/233
bye

chaveiro:
Hi, your solution with last_feed_value is fine.
But what if you dont have a feed to get the last value from? (i now have dynamic feeds written by schedule)
Use my newer reset2original process, it gets its value from current input value, not some feed.

chaveiro:
I've made another commit on top of the original with the module files i've missed and new conditional processor.
See it here: https://github.com/chaveiro/emoncms/commit/aaf906df82bdce5e3dac52d5f86c6c8f5352b73a
Please merge the last one. Sorry!

chaveiro:
I've made another commit on top of the original. This adds smart processor list selection. Only feeds available to a process are displayed.
See it here: https://github.com/chaveiro/emoncms/commit/d8014796e0a7950407ae4e84de86e759e0d86b14
Please merge the last one. Sorry!

TrystanLea:
Hello Chaverio, Is it possible to separate out the schedule module as an optional module? Perhaps you could host it on your github and I could link it from the optionals module list?

chaveiro:
Hello TrystanLea, I thought of that. The configuration part of schedules is already modularized in /Modules/schedule/. 
The usage of schedules in processlists is a new ProcessArg enum value in Libs/enum.php, processmodel.php and processlist.js had added support for it too and they are not modularizable by design. 
What i can do is to remove the dependency of files in Modules/schedule that is the configuration part of the schedules and disable usage of schedules in processlist if it's not available.

Just to clarify, you think people should not get schedule processor functionality by default or are there other concerns about it ?

TrystanLea:
Hello chaveiro, my main concern is just that I need to keep the main emoncms repository maintainable, Ideally the main emoncms package needs to be as minimal as possible with additional functionality being added as optional modules. 
Perhaps there is a way to make input processing modular too? so that optional modules can make new input processes available.

mpruessmeier:
Yes, you are thinking about kind of preprozessing. A new Module builds new dataflow for a (new) Input. The Input is then maybe only loged to a feed. 

I think this is the right way.

chaveiro:
Hello TrystanLea
I absolutely agree with you, there should be modular support for processors.
That is relatively easy to implement php-side in API workflow for the process engine. But much work mustbe done on the javascript side as we have tight coupled processlist configuration functionality between php html divs placeholders and js events handling.
We can discuss architectural design to part process_model.php and input view code for modular support on other branch thread. It is a major refactoring.
For now i suggest to merge it as is so we can have a starting base of work with what is already available and also because many functional bugs wore fixed on js that will help further developing.

chaveiro:
TrystanLea any update on this merge ?
Longer you wait, more dificult is to merge conflicts with new pulls.
If you need help on code review let me know.

TrystanLea:
Hello Chaveiro, I'd like to merge this pull request into a different branch, minimizing changes to the master branch at this point, I've actually been trying to reduce the feature set in the master branch for a while in order to make it easier for me to maintain and hopefully to make it easier to use (hence removal of options to create new feeds with mysql)

What I propose is that we create a new branch, what do you think of the name Extended? and then we clearly document that using this larger feature set branch is an option. I've added a note to all the installation guides to look at the branch list on the emoncms github front page.

I will then merge all pull requests on this extended branch after a quick check of the code (but no in use testing). If you can then help with maintaining this Extended branch, checking that the pull requests I merge work that would be a great help.

Development I think is stifled by trying to create a single version, I think we need to create multiple branches/versions at this stage so that we can try many ideas in parallel.

You may also be interested in reading my ideas here on removing input processing and the separation between inputs and feeds: http://github.com/emoncms/emonview

Il setup a branch called Extended now

aleszuin:
Very good, I think the branch extended is a good choice, i'm already using it. I converted my processes to the new "reset2original" instead of "LastFeedValue" but I encountered two problems so in some cases I had to still keep the process "LastFeedValue".
The first case is when the at the original value of the input was applied a correction factor for calibration before saving the data on feed; with Reset2Original I should reapply at any point this fix; the second case is less simple,
when I need to create 2 feed from a values:
input1 -> elab1 -+-> feed1 ---> feed2
                          +-> feed1 ---> feed3
starting from input1 would repeat it all elab1.
This because the event for beginning a process is only input, not feed .

final thing , in my branch I have also included the SteelSeries Gauges
(Http://openenergymonitor.org/emon/node/2791) i'd like to know if those would be incorporated.
Thank you

TrystanLea:
Hello Aleszuin, great! Yes Id be happy to merge the steelseries gauges into the extended branch if you guys can help with maintaining it and testing. Il merge bug fixes promptly. The gauges look great.
Thankyou Aleszuin!

PR #286, opened 2015-02-09
Fixed bug when OS and MySQL are not using the same decimal separator

jesjimher:
I found a few days ago that my kWh counter was fixed at some round value (120.0) and wasn't growing anymore, despite consuming electricity. After some debugging, I found the problem was that the update query sent to MySQL was using the system decimal separator (a comma in my case), while MySQL doesn't take locale into account and always expects a dot. The result is that MySQL was ignoring all decimal places, effectively truncating the value.

The patch here pre-formats the float value to a format suitable for MySQL (10 fixed decimal positions, dot as a decimal separator), so it works well regardless of system locale.


TrystanLea:
thanks!

PR #290, opened 2015-02-11
Reduced login form size and some code formating

TrystanLea:


chaveiro:

PR #291, opened 2015-02-11
Dashboard menu with better support for mobile devices

TrystanLea:
https://github.com/emoncms/emoncms/pull/238


chaveiro:

PR #299, opened 2015-02-26
feedwriter permissions

pb66:
Hi Trystan. minor change, I have just set the feedwriter permissions as executable to get round git being difficult when changing branches after doing chmod locally.


pb66:
Hi Trystan
I just went to submit a pull for some corrections to the docs re emonhub.conf and they have tagged on to this outstanding pull request to change permissions on the feedwriter.py. 
They are both minor I see no reason they can't go together. let me know if you think otherwise.
Paul

PR #304, opened 2015-03-21
Backport of minor fixes on Extended to Master to simplify future merge.

chaveiro:
On Modules\feed\feed_model.php:  'Explicitly set decimal separator, in case current locale doesn't use a dot (MySQL expects always a dot, regardless of locale)' 
Was removed because of possible issues  with large numbers.
See: https://github.com/emoncms/emoncms/pull/304/files#diff-ef4a46a9ff8dd8d8cc90d51b3b0f3362R644

Fix was made to Modules\user\user_model.php API language is always EN.
This ensures all API post data (arduino firmware) follows english localization style.
See: https://github.com/emoncms/emoncms/pull/304/files#diff-3a8983b5e260461fe91544f3ff97de77R56


TrystanLea:
Looks good, thanks chaveiro!

PR #308, opened 2015-03-26
VIS module bugfixing

chaveiro:
Updated visualizations:
-> Multigraph: bugfix and improve performance of javascript. Less json calls.
-> Compare: changes for a generic comparator
-> All: Show label of feeds on selectors and usage of bootstrap where required. JS events fixing where needed.


chaveiro:
After this merge, extended is labeled v8.5.0 XT.
Clean a lot of redundant lines of code from 8.5 branch, ensured best
practices.
If not needed may delete Modules: Node (deprecated), My Electric and
Schedule folders, they are modularized and will not be required if not
used.

PR #309, opened 2015-03-26
Timezone merge from Extended to Master

chaveiro:
This was only tested on Extended (untested on Main), but should be ok.
Please note it includes also some minor fixes on language selection and javascript behaviour from extended (feedlist and input modules) nothing critical.

Please note that server php.ini timezone setting must match server timezone.
Recomend to not use "GMT" or "UTC" timezones, use localized timezones like "Europe/Lisbon" that match real server location or DST changes may get wrong.

To update, apply patch then:
Go to Admin -> update Database.
Go to Account -> set your nodes timezone. (do it for all users, or change in DB user table directly)

If one user has nodes from different timezones, then a new account must be created for each node timezone used.

Then all day feed accumulator (like kwh/d) past data might need an offset fix.
If you have a database engine it's easy as add/subtract the correct value to the old data. This is to fix users reporting that day starts at 8'o clock (or not mid night).
Report issues here: http://openenergymonitor.org/emon/node/10363


TrystanLea:
Thanks a lot chaveiro! this is now merged into the v8.5 branch for testing!

PR #313, opened 2015-04-06
Changed to jquery-1.11.2 + update flot lib.

chaveiro:
Due to break of some vis in android, we need to use jquery-1.x.
Plus minor speed fixs and updated flot lib.


chaveiro:
On https://github.com/chaveiro/emoncms/commit/339782150e71514cbe2a1508714b1f5c2a285bf9 :  i've included flot lib examples folder (~120 files) in \Lib\flot\examples to give future lazy developers developer some view of what can be done. This folder can be safely removed so save space.

PR #326, opened 2015-05-12
8.5.1 XT

chaveiro:
This is a major refactoring now labeled version 8.5.1 XT with the follow changes :

- Added touch support for dashboard edit module.
- Added 2 dashboard widget options: width and height. Can define them as percentage (or pixels) so we got adaptive dashboards for different resolutions (devices).
- Added dashboard background color configuration, available in dash editor wrench icon.
- Separate vis module from dashboard module by moving dashboard files to Modules\vis\widget folder already supported by design, it's now really a module that can be removed without causing issues, also a great example of how to create new modules with its own widgets.
- Relocate dashboard widget js files to standard module widget folder Modules\dashboard\widget.
- Bug fixing here and there on dashboard usage and widgets editing.
- Speed improvement for slow cpu servers in php code of some controllers by using 'else if' statements where permitted. So that if a condition is found its not required to check other exclusive ones.
- Simplification of code and best practices enforced on every change.

To update need to do : Admin -> Update Database

Backward compatible and all tested as usual!

This is a preparation for new MyElectric/SolarPV dashboards modules as widgets components usable from dashboard GUI editor.

TODO: Define dashboard menu design for better integration on small screens.


TrystanLea:
Wow sounds amazing Chaveiro!

PR #333, opened 2015-05-23
Logging Fix

chris-y-meyers:
Match line 96 ' $log4php_configPath = 'logconfig.xml';' in default.settings.php

### DIFF
--- a/filelogger.logconfig.xml
+++ b/filelogger.logconfig.xml
@@ -1,9 +1,0 @@
-<configuration xmlns="http://logging.apache.org/log4php/">
-<appender name="myAppender" class="LoggerAppenderFile">
-<param name="file" value="emoncms.log" />
-</appender>
-<root>
-<level value="WARN" />
-<appender_ref ref="myAppender" />
-</root>
-</configuration>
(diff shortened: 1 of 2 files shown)


kumarmohan7:
i have new problem in feed data 
error  is "Error : feed cloud not created" 
how to create feed cloud
feed error

chaveiro:
On Chrome press F12, goto network tab, select record network log.
Then try to create again and give the result that appears for the line with the url : http://youhost/feed/create.json?*

kumarmohan7:
i have this msg in my browser 

error
{"success":false,"message":"feed already exists"}

chris-y-meyers:
In that case, there are 2 scenarios:
   - the feed name was used for a another EXISTING feed
   - there was a glitch when deleting a PREVIOUS feed of the same name (confirm with MySQL Workbench or PHPMyAdmin):
      * remove the entry from MySQL 'feed' table
      * if using Redis you need to remove the entry from 'feed' key

There is documentation online on how to integrate SQL and Redis commands in PHP for the above points.

---------
P.S: for these kind of questions ideally there is the Open Energy Monitor forum or Github Issues (https://github.com/emoncms/emoncms/issues)

PR #334, opened 2015-05-26
Removed modules available individualy

chaveiro:
Removed node module.
Removed myelectric module.
Updated login page to use bootstrap elements only and allow of browser
to ask for save login. (ok on IE and Firefox)
Refactored dashboard_langjs.php usage to be a php included resource, so
we dont have replicated location functions in it.


kumarmohan7:
please send modified zip file to my mail id (kumarmohan7@gmail.com)

chaveiro:
Part of the recent touch support, dashboard menu is now a dropdown
acessible from the main menu so a dropdown  list is presented instead of
broken formatation on small devices.
All Tested.

TrystanLea:
Great chaveiro!

PR #336, opened 2015-06-03
Fix Min Max processors + Merge with 8.5 latest

chaveiro:
Refactored some code from 8.5 -> Instead of if (!is_numeric($lastvalue['time'])) in feed_model.php i fixed the reason it may not be.
Also some minor fix on getstartday function on process_model.php

### DIFF
--- a/Modules/feed/engine/MysqlTimeSeries.php
+++ b/Modules/feed/engine/MysqlTimeSeries.php
@@ -128,7 +128,6 @@ class MysqlTimeSeries
 
         $result = $this->mysqli->query("SELECT time, data FROM $feedname ORDER BY time Desc LIMIT 1");
         if ($result && $row = $result->fetch_array()){
-            $row['time'] = date("Y-n-j H:i:s", $row['time']);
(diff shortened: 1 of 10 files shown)


TrystanLea:
Hello Chaveiro, Im not sure that removing the is_numeric check will fix the reason it may not be because it was to convert the output of the last time value stored in redis which would still be in date string when calling get_timevalue for the first time. At the moment in your code I think the date string will be passed back to the process_model. Or are you fixing it with the (int) in $now = DateTime::createFromFormat("U", (int)$time_now);?

chaveiro:
The fix was applied on some other lines that get the updated time from database table feeds. Search for the comment: // feeds table is date time, convert it to epoh
This ensures all 'lastupdate' variables (and redis cache) always gets the epoh time format value and not a string in datetime format that the if was used for fixing.

The fix in process_model is about other issue, the createFromFormat U was giving an error on ->settimezone if the epoh number was passed as a string type. It requires an int var type.

PR #339, opened 2015-06-08
 	Change array initialisation to be compatible with PHP pre-5.4

jlm:
See http://openenergymonitor.org/emon/node/10687


mking007:
Hi, this fixes my error as per here: http://openenergymonitor.org/emon/node/10687

PR #344, opened 2015-06-17
XT 8.5.2 | 2015.07.06 release

chaveiro:
Fix dashboard menu for older php version.

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -20,7 +20,7 @@ DirectoryIndex index.php
   RewriteCond %{REQUEST_URI} !^/?Modules.*$
   RewriteCond %{REQUEST_URI} !^/?scripts.*$
   RewriteCond %{REQUEST_URI} !^/?Theme.*$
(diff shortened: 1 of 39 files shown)


chaveiro:
XT 8.5.2 | 2015.07.02 release
Updated javascript feed selectors to be grouped by tag name in processlist, visualization and dashboard modules.
Updated last updated to be based on server time instead of client time.
Refactored feed_model.php insert_data_padding_mode function removed, padding_mode is now optional parameter of insert_data and functions comments for easy understanding.
Refactored input_model.php some checks and functions comments for easy understanding.
Added new configuration setting $fullwidth true/false, makes dashboard and vis use full screen width.
Updated engines to also support this parameters as optional :
 - MysqlTimeSeries.php : skipmissing ,limitinterval
 - PHPFina.php : padding_mode 
 
TODO: Remove dependency on dashboard widgets (3rd button on editor) of feed names, use feed id instead. Will break existing user dashboards.

PR #348, opened 2015-07-03
logger: specify log path

LudovicRousseau:
The log file full name is /var/log/emoncms.log according to logconfig.xml


chaveiro:

PR #362, opened 2015-08-03
emoncms: use full rather than short tags in update_view.php

cyclops8456:
Use full rather than short tags in update_view.php for hosts with short
tags disabled, and a few indentation changes.


chaveiro:
Already implemented on V9.

JSidrach:
(This Issue can be closed @TrystanLea)

JSidrach:
(This Issue can be closed @TrystanLea)

PR #363, opened 2015-08-04
Enable override of SMTP port and encryption settings.

julienengie:
Using a SMTP serveur which use TLS, I have modified user_model.php to be able to override port to 465 and encryption to 'ssl'.

This pull request enable the ability to override default settings from settings.php adding optionally these parameters : 'port'=>, 'encryption'=>

I have also deleted the port 26 provided when creating the Swift_SmtpTransport instance, letting the Swift Mailer library setting the default smtp port which is 25.
If you want keep a backward compatibility, you should hardcode 'port'=>26 in default.settings.php, but usually, the default smtp port is 25.

### DIFF
--- a/Modules/user/user_model.php
+++ b/Modules/user/user_model.php
@@ -349,8 +349,11 @@ class User
                         return array('success'=>false, 'message'=>"Could not find SwiftMailer - cannot proceed");
                     }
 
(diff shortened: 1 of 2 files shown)


TrystanLea:
Thanks Julienengie, this is now integrated in Chaveiro's re-factored email library in v9.0

PR #364, opened 2015-08-04
9.0

TrystanLea:
Merge of Chaveiro v9 into main emoncms repo


arjenhiemstra:
This is going to be a great release! I'm testing it right now and it seems to work very well, especially the dashboard drop down menu is great!

One thing to mention though, the "Show published dashboards as single items" added in a recent commit clutters the UI, sometimes overlapping the other UI items, why not stick to the drop down menu or make this functionality optional?

Paul-Reed:
I tend to agree. It's unnecessary as all of the dashboards can be quickly selected via the single drop down menu, and the bolding of published dashboards is another unnecessary distraction.
But, it's good to explore different options, to see what works and what doesn't - that's evolution!

PR #378, opened 2015-09-03
Support for external themes

JSidrach:
If the $theme variable in settings.php is not "basic", it looks now for the path of the external theme (previously, it was ignored).

### DIFF
--- a/default.settings.php
+++ b/default.settings.php
@@ -19,7 +19,7 @@
 
 //3 #### MQTT
     // Enable this to try out the experimental MQTT Features:
-    // - updated to feeds are published to topic: emoncms/feed/feedid    
+    // - updated to feeds are published to topic: emoncms/feed/feedid
     $mqtt_enabled = false;
     $mqtt_server = "127.0.0.1";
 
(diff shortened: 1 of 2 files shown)


chaveiro:
I think its better to reduce context of folder structure for themes to BasePath/Theme/ and move main theme to  :
    BasePath/Theme/default 
Support for others themes as sub folders,  like: 
    BasePath/Theme/other_theme_name 
Instead of just BasePath/some_theme_name.
What do you think?

JSidrach:
I was putting the external themes on an absolute path, outside of emoncms,
but your proposal seems more reasonable.

This involves changing some code of the default theme too, as some paths
(images) load from there. It also requires changing some paths somewhere (I
am thinking of the login image, for example).

Finally, we should add some lines to the gitignore so it ignores everytheme
but the default (similar to what I did with Modules).
I can work on this today, do you mind?

JSidrach:
Implemented in #379 . I have not renamed basic to default so the same default.settings.php will work without any change.

chaveiro:
Cool, nice work.
It's looks better this way and will avoid confusion with folders role in future.

PR #380, opened 2015-09-03
Fix incorrect logins

JSidrach:
I did not fixed incorrect logins in the previous pull request, and if the credentials were wrong the page would reload incorrectly. Sorry, my bad.

### DIFF
--- a/Modules/user/login_block.php
+++ b/Modules/user/login_block.php
@@ -28,7 +28,7 @@ global $path, $enable_rememberme, $enable_password_reset, $theme;
   <div class="well">
     <img src="<?php echo $path; ?>Theme/<?php echo $theme; ?>/emoncms_logo.png" alt="Emoncms" width="256" height="46" />
     <div class="login-container">
-        <form id="login-form" action="#" method="post">
+        <form id="login-form">
             <div id="loginblock">
                 <div class="form-group register-item" style="display:none">
                     <label><?php echo _('Email'); ?>
@@ -163,13 +163,13 @@ $('input[type=text]').on('keypress', function(e) {
         if ( register_open ) {
             register();
         } else {
-            login();
+            return login();
         }
     }
 });
 
 $('#login').click(function() {
-  login();
+  return login();
 });
 
 function login(){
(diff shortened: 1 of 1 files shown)


Modules/user/login_block.php
```diff
@@ -182,11 +182,12 @@ function login(){
     if (result.success)
     {
         $('#login-form').submit();
-        window.location.reload(true);
+        return true;
```
chaveiro: This will break redirect to previous page after login...

chaveiro: This will break redirect to previous page after login...

JSidrach:
Is there any way of knowing what the original page was in PHP? I'm thinking on something like:

~~~php
<form action="<?php echo $redirect_to; ?>">
</form>
~~~

That would solve the issue. Otherwise, if we don't return true/false, the form data may not be compatible with the browsers 'Save password' feature.

Any thoughts?

JSidrach:
I have made some test and I cannot reproduce what you mentioned... I type http://localhost/emoncms/input/view without being logged in, then I enter my credentials and I'm redirected to the input/view page.

Any example of breaking redirects?

chaveiro:
Ok it was on a particular test scenario i saw the issue, i had this url host/app?#myelectric 
But this is not an issue for anyone else as this url format does not exists on repo version.

JSidrach:
Should then I install the app module and see if it is ok? Or are you fine as it is now (now = the pull request version)?

chaveiro:
It's fine, this is a test i had, default apps passes apykey (for now) so it's always logged in.
Apps in future should not pass apikey nor use the # character so no problem either.
But if you can find a way to login not change the bookmark after # if it was passed on the url, it would be good also.

JSidrach:
Could you do additional check? I'm writing now the url with the # character and it does not remove the content after it. Maybe in your case you still have the action="#" in the form tag? I did remove it in commit #380, and I think it is the cause of your issue.

JSidrach:
As additional information, the action="#" was in the previous version, but it was not trigger since window.location.reload(true); prevented the form from submitting.

Issue #369, opened 2015-08-24
No accents allowed in inputs / dashboard names


arthurlutz:
It would be nice to be able to add descriptions for inputs with accents (for french speaking users for example), same goes for dashboard names and various other places in emoncms.

Jerome-github:
I thought those were fixed with the move to UTF-8.

https://github.com/emoncms/emoncms/issues/84
https://github.com/emoncms/emoncms/issues/48
https://github.com/emoncms/emoncms/issues/17

If your database was created before UTF-8 became default, see https://github.com/emoncms/emoncms/issues/85 for the transition.

euqip:
I think the problem is located in the regex filter used to clean data before storage.
example out of user_model.php
        $flds .= '  location = "'.preg_replace(REGEX_STRING_ACCENT,'',$data['location']).'"';
        $flds .= ', language = "'.preg_replace(REGEX_STRING,'',$data['language']).'"';

where 

define('REGEX_STRING','/[^\w\s-]/');
define('REGEX_STRING_ACCENT','/[^\s\p{L}0-9]-\'/u');

these are to be defined once at the strart  (and correctable or adjustable once), to be usable anywhere.
 
Effectively the database needs also to accept utf-8 characters because accents are not allowed in ASCII.

Hope it will help

Paul-Reed:
Arthur, you've currently 5 open issues, and I'm not alone in thinking that at least some of the issues may have already been resolved through a series of bug fixes as emoncms has evolved into the current stable version - v9.  
Could I ask that you consider updating your installation to the current version, otherwise we may fall into the trap of duplicating the development work which has already taken place.  
We would be grateful if you could update the open issues with the results please.

arthurlutz:
@Paul-Reed hi, I've found some time to upgrade to 9.2. I've successfully renamed my dashboard to something with accents. Closing this ticket. 

Will try to find some time this afternoon to go through my other ticket. Thanks.

Paul-Reed:
Much appreciated Arthur.

PR #382, opened 2015-09-04
UTF-8 support

JSidrach:
- Fixed regex expressions to support utf-8 characters
- Fixed a security bug where the $tag parameter on input_mode.php was not being sanitized
- Set charset to utf-8 in the database connection

It should fix the issue #369

### DIFF
--- a/Modules/dashboard/dashboard_model.php
+++ b/Modules/dashboard/dashboard_model.php
@@ -91,7 +91,7 @@ class Dashboard
 
         $result = $this->mysqli->query("SELECT * FROM dashboard WHERE userid = '$userid' AND id='$id'");
         $row = $result->fetch_array();
-        if ($row) { 
+        if ($row) {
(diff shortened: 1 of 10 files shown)


tgmaxx:
This commit caused all of my inputs that have underscores in their names to stop working. I reverted to previous commit and they are working again, so this update has some issues.

JSidrach:
Taking a look at this... sorry for the inconvenience.

tgmaxx:
To be clear, by input names, I mean input "key" names.
Thanks!

JSidrach:
I found the issue. Turns out in regexps, \w means letters+numbers+underscores... I thought it did not include underscores, so I replaced it with \p{N}\p{L} (unicode numbers and letters). It should be an easy fix, so I'll upload it as soon as possible.

JSidrach:
Should be fixed by #383. Please reply here if the issue persists. Thanks for finding out this one!

tgmaxx:
Ok no problem. Thanks for all of your work.

PR #383, opened 2015-09-04
Fix bug that broke regexp with underscores

JSidrach:
Sorry, completely my bad. I replaced \w with \p{N}\p{L} but forgot that \w actually takes underscores as well.


TrystanLea:
Thanks a lot Juan, great!

PR #385, opened 2015-09-05
EmonCMS 9 RC | 2015.09.05

chaveiro:
Merge of:
        UTF8 support
        Browser asks for remember user and password
Minor bug fixing


Modules/dashboard/dashboard_model.php
```diff
@@ -118,7 +118,7 @@ public function set($userid,$id,$fields)
         if (isset($fields->alias)) $array[] = "`alias` = '".preg_replace('/[^\p{L}_\p{N}\s-]/u','',$fields->alias)."'";
         if (isset($fields->description)) $array[] = "`description` = '".preg_replace('/[^\p{L}_\p{N}\s-]/u','',$fields->description)."'";
 
-        if (isset($fields->backgroundcolor)) $array[] = "`backgroundcolor` = '".preg_replace('/[^[0-9a-fA-F]]/','', strtolower($fields->backgroundcolor))."'";
+        if (isset($fields->backgroundcolor)) $array[] = "`backgroundcolor` = '".preg_replace('/[^[0-9A-F]]/','', strtolower($fields->backgroundcolor))."'";
```
chaveiro: Only caps letters are supported.

JSidrach: Does that line what is intended? If we first use the strtolower, and then remove all the characters that are not uppercase, would not the string be empty?

PR #387, opened 2015-09-08
EmonCMS 9 RC2 | 2015.09.09

chaveiro:
Added support for Process List Modules.


JSidrach:
Does this change the input API? Are requests now made by POST instead of GET to set the process list? I'm a bit worried the API could be inconsistent: why every feed API method is GET (even the create one) but the processlist is POST?

chaveiro:
Hi JSidrach , just the processlist SET process field was changed to post due to the possibility of a larger length string that wont fit the query string.

JSidrach:
Valid concern. I'll update [EWatcher-Users](https://github.com/jsidrach/ewatcher-users/) to use POST for the process list too. Maybe in the future something similar will need to happen on the bulk input API methods.

Nice work!

TrystanLea:
As always, thanks a lot for this!

PR #400, opened 2015-10-27
Emon CMS 9.1 | 2015.10.27

chaveiro:
- Changed validation logic for $max_node_id_limit config setting, from node id numeric to maximal number of diferent nodeids. A node id can be a string now.
- Added Lib swiftmailer and a generic email helper class to core for sending emails.
- Added 5 new processors for virtual source feed : + - * / and 1/ 'source feed'
- Added support for devicekey instead of apikey, if using device module. A device key only allows posting to the device configured node.
- DST change fix for KWh processors.


TrystanLea:
Merged via command line, now available https://github.com/emoncms/emoncms/tree/9.x

PR #402, opened 2015-10-31
Allow 3-character usernames

njh:
Is there any reason not to allow 3-character usernames?

Also passwords should probably be required to be longer than 4


TrystanLea:
thanks njh!

PR #405, opened 2015-11-18
Empty feed data

TrystanLea:


chaveiro:
Hi, empty feed data only supported on PHPFina ?

TrystanLea:
Yes ;)

TrystanLea:
For this to work, the feedwriter script needs to be stopped prior to emptying the feeds as the meta data start time is changed and the feedwriter needs to reload its cache. I cant see an easy fix for this at the moment so this feature will need a bit of work until its ready to merge I think - alongside addition to other engines.

PR #406, opened 2015-11-18
Fixed tooltip to ensure that the tooltip is visible when hovering

Quentinb:
Fixed tooltip style code to ensure that the tooltip is visible when hovering over a data point at the edges of the graph. When the x and/or y coordinate go below 2, resulting in a "-x" or "-y" value, we flip the tooltip to the other side of the cursor.


chaveiro:
Thanks, merged to 9.2.

PR #408, opened 2015-11-26
Update input.js

arjenhiemstra:
"input/process/list.json" does not refer anymore to anything in the controller 
should be "input/process/get.json"


chaveiro:
Fixed 9.2 master.

PR #409, opened 2015-11-26
Change passwordresetmessage page location

arjenhiemstra:
Place \<div id="passwordresetmessage"\> outside \<div id="passwordresetblock"\> to keep result visible after submit


arjenhiemstra:
This PR was overwritten with a later PR from Chavero, can it be added again?

Paul-Reed:
Added again. Thank you.

PR #411, opened 2015-11-27
9.x merge to master

chaveiro:


Modules/admin/admin_main_view.php
```diff
@@ -67,7 +67,7 @@ function system_information() {
         </td>
     </tr>
 <?php
-if ($log_enabled and !$allow_emonpi_update) {
```
chaveiro: This can be configured on settings.php

Issue #403, opened 2015-11-09
Only one value for timeseries feed


arthurlutz:
Hi, 

I'm having a problem when trying to graph a value by day on emoncms 8.5.2  2015.07.20

![screenshot from 2015-11-09 16 07 45](https://cloud.githubusercontent.com/assets/108437/11037125/77515a72-86fc-11e5-97cb-a120e876f0fa.png)

Here is the feeds view : 

![screenshot from 2015-11-09 16 08 05](https://cloud.githubusercontent.com/assets/108437/11037130/800b8976-86fc-11e5-810b-24b094b3b9a5.png)

and the transforms from the input 

![screenshot from 2015-11-09 16 08 36](https://cloud.githubusercontent.com/assets/108437/11037146/8f9c260c-86fc-11e5-90db-970b225bf291.png)

The "transformation" is inspired by Fred's comment that can be found here  http://hallard.me/teleinfo-emoncms/#comment-10290

```
1 -> Log to Feed – Compteur (Utile pour la phase de dessous)
2 -> Total pulse count to pulse increment – (Diff N- N-1)
3 -> Log to Feed – Elec. Wh
4 -> Accumulator – Elec. Wh Acc
5 -> X 0.001 ( vers kWh)
6 -> kWh to kWh/d (Consommation Journalière en kWh/d)
7 -> kWh to Power (Puissance instantanée)
```

I'm getting the same problem with my teleinfo electricity counter. Values seem ok but no graph for the daily kWh/d.

arthurlutz:
am getting the same problem with v9

chaveiro:
May be related to this https://github.com/emoncms/emoncms/pull/414.
Can you test with the fix?

arthurlutz:
@chaveiro #414 solves my issue. Thanks.

PR #414, opened 2015-12-03
Update PHPTimeSeries.php

arjenhiemstra:
In function get_data try to find the first valid data entry within the given timeframe and set $start to that time
This will 'sync' the start of the search through the dataset to the first real data entry. Preventing problems if $limitinterval = true and the passed &start   value is based on for example browser time.

### DIFF
--- a/Modules/feed/engine/PHPTimeSeries.php
+++ b/Modules/feed/engine/PHPTimeSeries.php
@@ -259,7 +259,7 @@ class PHPTimeSeries
             if ($limitinterval)
             {
                 $diff = abs($dptime-$time);
-                if ($diff<($interval)) {
+                if ($diff<$interval) {
                     $value = $array['value'];
                 } 
             } else {
--- a/Modules/feed/engine/PHPTimeSeries.php
+++ b/Modules/feed/engine/PHPTimeSeries.php
@@ -242,6 +242,15 @@ class PHPTimeSeries
         $time = 0; $i = 0;
         $atime = 0;
 
+        // Try to find the first valid data entry within the given timeframe and set $start to that time
+        $pos = $this->binarysearch($fh,$time,$filesize);
+        fseek($fh,$pos);
+        $d = fread($fh,9);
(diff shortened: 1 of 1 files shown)


chaveiro:
with this change you set start time to the first timestamp on the engine, independent of specified start time.
After a few months of data you will get performance issues.

arjenhiemstra:
Chaveiro, it is related to this forum post: http://openenergymonitor.org/emon/node/11720

For example vis. module zoom.php request a data set (5 years of data) with the start time of the request being the browser time. The database timestamps are based on 0:00 +- timezone offset. This leads to datapoint not being displayed.

I expected the added search to set the following search to the in data present startingpoint of the given timeframe. The following time entries in DB should then better fit the given interval and not cause the interval/2 test to fail (with as result no or limited data in graphs) 
Apparently this is not the case.

I did a small test with 3+ years of data, first without the new code, second example with. The second example shows less searches through the file on disk to find al the values in the given timeframe. I didn't test with a larger dataset so you are probably right.

No patch:
PHPTimeSeries.php|get_data feedid= 39 timestamp_in= 1291590000 timestamp_found= 1291590000 timestamp_end= 1449172805 number of searches= 1825

With patch:
PHPTimeSeries.php|get_data feedid= 39 timestamp_in= 1291590000 timestamp_found= 1353715200 timestamp_end= 1449172992 number of searches= 1106

arjenhiemstra:
I've change $time to $start and changed the pull request.
Sorry for the error.

arjenhiemstra:
Any thoughts on the proposed solutions in my forum post?
http://openenergymonitor.org/emon/node/11720

chaveiro:
I can test this, but suppose it's ok now.

arthurlutz:
am interested in this patch (for #403) but would rather prefer to get it with a new emoncms version. Any ideas on when that is planned ?

arthurlutz:
sorry, got round to take a look at the patch, was so small I modified the line on my emoncms. It solves my problem! Thanks to all who have worked on this.

PR #427, opened 2015-12-16
Update to install docs

pb66:
Just adds a nore to advise user to point emonhub at the newly installed localhost emoncms rather than the default emoncms.org whilst adding the apikey


Paul-Reed:
Thanks for the update Paul

Paul-Reed:
Paul, are you doing a similar update for the [Jessie install guide too](https://github.com/emoncms/emoncms/blob/master/docs/RaspberryPi/readme.md)

PR #428, opened 2015-12-16
Added note to prompt user to set local emoncms url

pb66:
http://openenergymonitor.org/emon/node/11809
http://openenergymonitor.org/emon/node/11456


pb66:
ref Update to install docs #427

Paul-Reed:
Thanks again Paul

Issue #430, opened 2015-12-18
Password security issues


jpsingleton:
Please increase the maximum allowed length of passwords from 30 characters to something more reasonable such as 250. The allowed length should also be fed back to the user and could even be used as a maximum length attribute on the form input field. `Password length error` is not as helpful as it could be. I had to dig into [the code](https://github.com/emoncms/emoncms/blob/stable/Modules/user/user_model.php#L170) to find what the restriction was. I also suggest increasing the minimum length from 4 to something more secure such as 10.

As the password is hashed (good work) the length is irrelevant. The hash output will always be the same size. You only need to protect against obscenely large malicious submissions.

Regarding hashes, although a couple of rounds of SHA256 with a salt (similar to an HMAC) used to be sufficient, these days it can be cracked in a trivial amount of time. It would be better to use something with a higher work factor such as bcrypt or PBKDF2 (with at least 10000 iterations). You can transparently upgrade existing hashes by using them as the input to the new hash algorithm like this pseudo-code: 

`newHash = pbkdf2(oldHash, salt, 10000)`
`passwordOk = newHash == pbkdf2(sha256(salt + sha256(password)), salt, 10000)`

Also, I've not tested this but it looks like you may have a SQL injection vulnerability in the way you are building your SQL statements. I'm not too familiar with PHP and mysqli but prepared statements and parameterized queries are safer.

elyobelyob:
Really? Who would want a password longer than 30 characters?

Jerome-github:
Maybe someone who stores his passwords in a safe place rather than remembering all of them.

I don't do that myself, but I agree there is no need limiting to 30. Especially since 

> As the password is hashed (good work) the length is irrelevant. The hash output will always be the same size. You only need to protect against obscenely large malicious submissions.

jpsingleton:
Yes, the primary use case is for password managers but also for people who use pass-phrases instead of passwords. I use a tool to generate a long unique password for each sign-up however many people still reuse credentials despite the big risk. This is why a long password and slow hash function are a good idea. 

To turn the argument around, what is the reason for wanting to restrict passwords to a short string?

pb66:
Although I would be surprised by a >30 char password, as you say there is no real reason to restrict it to 30 if users want more. But a minimum of ten is a bit extreme, 

[...]

PR #431, opened 2015-12-20
Increase minimum and maximum length of passwords

jpsingleton:
Increase the minimum length of new passwords to 10 (same length as reset generated ones).
Increase the maximum length of all passwords to 250 (enough for users but stops huge posts).
Improve the password length error message to be more helpful.

One of the issues from: https://github.com/emoncms/emoncms/issues/430


pb66:
IMO imposing a >10 length is over kill see "Password security issues #430"

Jerome-github:
I agree. Allowing long passwords is good. But allowing down to 6 chars is fine for me. No need to force > 10.

If we're concerned about password security, maybe we can (or maybe there are libs available that) compute some sort of password security indicator while the password is typed, based on length and usage of numbers, special characters,... and display an assessment like "weak", "average", "string" in red, orange, or red.

jpsingleton:
OK, one step at a time. https://github.com/emoncms/emoncms/pull/432

PR #432, opened 2015-12-20
Increase maximum length of passwords to 250 chars

jpsingleton:
Increase the maximum length of all passwords to 250 characters (enough for users but stops huge posts). https://github.com/emoncms/emoncms/issues/430


glynhudson:
Thanks

PR #448, opened 2016-01-17
Created ability to toggle lines in multigraph

software-monkey:
I often have a multigraph where most of the lines are small values, but one line has large values that dominate the graph. When that happens, I cannot get a very good idea of what the other lines look like. Here's an example:

image

So I added the ability to toggle lines by clicking on the name of the line in the legend. When you toggle the line, I have the y axis rescale so you can see the lines with smaller values.

Here I have hidden furnace:

image

When a line is hidden, you can zoom or pan the graph without restoring the line. You can also change from daily to weekly, monthly etc views without restoring the lines. You can hide as many lines as you like and restore them whenever you want by clicking on the name in the legend again.

### DIFF
--- a/Modules/vis/visualisations/multigraph/multigraph.js
+++ b/Modules/vis/visualisations/multigraph/multigraph.js
@@ -7,7 +7,7 @@
   function convert_to_plotlist(multigraph_feedlist){
    var plotlist = [];
    for (z in multigraph_feedlist){
-	var tag = (multigraph_feedlist[z]['tag']!=undefined && multigraph_feedlist[z]['tag'] !="" ? multigraph_feedlist[z]['tag']+": " : "");
+    var tag = (multigraph_feedlist[z]['tag']!=undefined && multigraph_feedlist[z]['tag'] !="" ? multigraph_feedlist[z]['tag']+": " : "");
     if (multigraph_feedlist[z]['datatype']==1){
       plotlist[z] = {
         id: multigraph_feedlist[z]['id'],
--- a/Modules/vis/visualisations/multigraph/multigraph.js
+++ b/Modules/vis/visualisations/multigraph/multigraph.js
@@ -2,24 +2,22 @@
   var timeWindowChanged = 0;
   var ajaxAsyncXdr = [];
   var event_vis_feed_data;
+  var hidden_lines = {};
   
   function convert_to_plotlist(multigraph_feedlist){
    var plotlist = [];
    for (z in multigraph_feedlist){
-    var tag = (multigraph_feedlist[z]['tag']!=undefined && multigraph_feedlist[z]['tag'] !="" ? multigraph_feedlist[z]['tag']+": " : "");
+	var tag = (multigraph_feedlist[z]['tag']!=undefined && multigraph_feedlist[z]['tag'] !="" ? multigraph_feedlist[z]['tag']+": " : "");
     if (multigraph_feedlist[z]['datatype']==1){
       plotlist[z] = {
         id: multigraph_feedlist[z]['id'],
         selected: 1,
         plot:
         {
+          idx: z,
           data: null,
+          temp_data: null,
           label: tag + multigraph_feedlist[z]['name'],
-          points: { show: true,
-                    radius: 0,
-                    lineWidth: 1, // in pixels
-                    fill: false
-          },
           lines:
           {
             show: true,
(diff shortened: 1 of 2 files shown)


Paul-Reed:
Like your last contribution, this is a very welcome addition, especially as the y-axis rescales to show greater detail.  
Would it be possible to have an indication of which lines are hidden - for example, for the Legend name to be faded or a line drawn through the name?
This would then remind us which, if any lines were hidden/shown.

Paul

glynhudson:
Nice! Good addition, thanks. Happy for this to be merged once you're ready. Any other comments tweaks before merge?

software-monkey:
Great idea on the legend color. I made the change (even kept the name of the variable colour to be consistent with the rest of the project). I'm also going to make a similar change for the timecompare module.

software-monkey:
I've added the change for timecompare to this pull request since they are somewhat related. If you'd like it in a different pull request, let me know and I can play around with branching and rebasing.

Paul-Reed:
Just tried your latest, and works fine here. Making the legend grey when the line is hidden is much better, you can see at a glance which data you are viewing. Same applies to TimeCompare.
Nice work!  
Paul

Paul-Reed:
Is [this an issue](http://openenergymonitor.org/emon/node/12021)?
Paul

software-monkey:
Crap, yes. That applies to both timecompare and multigraph. It is some leftover code from when I was debugging an early version of the code on timecompare. I cloned it to multigraph. I don't have negative values and didn't notice I'd left it in.

I created a pull request that fixes it: https://github.com/emoncms/emoncms/pull/451

software-monkey:
I just read through the code to make sure there's nothing else I left in like that. I didn't find anything.

chaveiro:
I think a plug-in should be integrated for hiding series. The way it's done is not modular, slow and dificult to implement on other graphs.
See if you can merged this with our flot: https://github.com/flot/flot/pull/848

glynhudson:
Nice addition, love being able to hide traces. Noticed that it only works
when viewing multigraph in fullscreen mode not when multigraph is displayed
on a dashboard. Not a big issue. Another thing I've noticed is tooltip
labels don't seem to be working for me anymore. Have you notices this?

Update: tooltip works when viewing multigraph embedded in a dashboard when
hiding traces does not work, when viewing multigraph full screen directly
tooltip labels don't work. This indicates that show/hiding traces is the
cause for tooltip labels stopping working.

Maybe show / hide traces could be disabled when tooltip is enabled and vice
versa is there is a compatibility issue?

On 23 January 2016 at 03:42, chaveiro <notifications@github.com> wrote:

> I think a plug-in should be integrated for hiding series. The way it's
> done is not modular, slow and dificult to implement on other graphs.
> See if you can merged this with our flot: flot/flot#848
> <https://github.com/flot/flot/pull/848>
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/emoncms/emoncms/pull/448#issuecomment-174126277>.
>



-- 
Glyn Hudson

http://megni.co.uk
http://adventuresplusnorthwales.blogspot.com

software-monkey:
Hi Glyn,

I'm looking into it and have a few questions for you

1) What browser are you using?

2) I'm able to hide traces in the visualization plugin, embedded dashboard, and full screen. Are you not able to hide traces in any of those modes? I'm not clear from your update if the problem is just with tooltips or if you have a problem hiding traces as well..

3) The tooltip works for me when viewing a multigraph embedded in a dashboard both with traces hidden and not hidden (I've used both Chrome and IE). See below for a screenshot from IE when traces are hidden. Are there any specific steps to repro or does the tooltip stop working as soon as you hide a trace?

image

4) Similar to above, when in fullscreen mode, tooltips also continue to work for me when traces are hidden. Tried both Chrome and IE (IE screenshot below)

image

That said, I've noticed that tooltips don't work in Chrome when using the visualization plugin to look at a multigraph. I reverted the toggle lines change on my server, but the tooltips still don't work in the visualization plugin. That bug is unrelated to my change.

software-monkey:
chaveiro,

Thanks for your feedback. Agree that having this logic live in flot would be preferable.

I took a look at the change you link to and I see a number of comments that there are bugs with it. That is listed as the reason it was never merged. I also noticed several comments that the yaxis doesn't scale when traces are hidden. That was the main point of my change here.

Looks like a bit more work is needed on that solution before moving to it.

glynhudson:
Hi @software-monkey ,

Thanks for getting back, here are my clarifications:

1. I'm using Chrome on Chrome OS  48.0.2564.88 beta (64-bit). Also tested on Firefox 43.0.4 with same results.

2/3/4. I've just figured out that the reason I could not show/hide traces when embedded in dashboard was my multigraph windows was resized to be so small that the button on the top RHS where overlay over the legend which obviously caused issues when trying to click the legend to show /hide (wops). I can now confirm that:

      * Viewing multi graph full screen: show / hide traces WORKS,  tool tip WORKS :+1: 
      * viewing multi-graph in vis window:  show / hide traces WORKS,  tool tip does NOT work :-1: 
      * viewing multi-graph embedded in a dashboard (assuming multigraph window is sufficiently sized):  show / hide traces WORKS,  tool tip WORKS :+1: 
      
It's not a big issue that tool tip does not work in vis window, for any in-depth analysis involving tool-tip I would expect user to go full-screen. I'm happy to close this issue if you are? Thanks again for quick response and.

software-monkey:
"viewing multi-graph in vis window: show / hide traces WORKS, tool tip does NOT work :-1:"

I was able to reproduce that on Chrome. I reverted this merge (the hide traces changes) and the tooltips still don't work. That is some other bug unrelated to this merge.

Perhaps open a separate issue for it and we can see if someone can track it down?

chaveiro:
Refactored to use a plugin at https://github.com/emoncms/emoncms/pull/459

PR #451, opened 2016-01-18
Bugfix to allow negative numbers on timecompare and multigraph

software-monkey:
Addresses http://openenergymonitor.org/emon/node/12021


Paul-Reed:
Thanks for the quick update!

PR #454, opened 2016-01-21
Fix uname and Gravatar URI

O2Graphics:
@exec('uname -a') wasn't working, and it's better to call native PHP functions

### DIFF
--- a/Lib/listjs/list.js
+++ b/Lib/listjs/list.js
@@ -135,7 +135,7 @@ var list = {
  
         'gravatar':
         {
-          'draw':function(value) { return "<img style='border: 1px solid #ccc; padding:2px;' src='http://www.gravatar.com/avatar/"+CryptoJS.MD5(value)+"'/ >" },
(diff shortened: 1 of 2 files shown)


Paul-Reed:
Does anyone really use Gravatar in emoncms?
When emoncms is run in a https environment, the gravatar link causes security warnings as they don't have a secure connection.
ssl
I'm not discrediting your pull request @O2Graphics just questioning if gravatar should be there at all.

Paul

O2Graphics:
I've no idea if anyone really use Gravatar in emoncms, but there's a Gravatar on the profile page: the first page displayed after the login.
I think it's better to avoid a security alert from the browsers.

Paul-Reed:
It's more an annoyance that a security threat, as visitors should not be able to visit and exploit the profile page anyway... and although I might be missing something, what's the point of it - to remind users what they look like - only they can see it!!

Paul

glynhudson:
I personally think the Gravatar is unnecessary. If it's causing a security
issue I would be quite happy to see it go.

- sent from my mobile device

PR #455, opened 2016-01-23
Added ability to hide tag name from multigraph legend

software-monkey:
By default the multigraph legend includes a tag for the feeds. In my case it is "Node 1". This takes up space on the graph. The buttons to switch between week month, year etc also takes up space. This makes it hard to hover over a line to see the exact reading there.

image

I added an option to the multigraph editor that lets you turn off the tag. It is enabled by default.

image

With it unchecked, the tag is not displayed and the legend is smaller.

image


glynhudson:
Excellent, thanks a lot. My multi graph legend looks a lot tidier now :+1:

PR #459, opened 2016-01-25
EmonCMS 9.3 | 2016.01.25

chaveiro:
Added:
- Flot toggle module added to multigraph and timecompare visualizations.
- Input and Feed list grouped details
- Export group

Merge of :
- Multiscreen series toggle.
- Multiscreen hide tag.


software-monkey:
Unfortunately when you hide a graph, it doesn't rescale the y-axis

If you take the starting graph

image

And hide furnace, you still cannot make out the values for the other lines, except that they are much smaller.

image

Here's what it looks like with the current hiding functionality.

image

Re-scaling the y-axis was the reason for my previous change so losing it removes all the value for me.

I did see a suggestion for how to fix this at https://github.com/flot/flot/pull/848#issuecomment-10782748.

This is more minor, but I also found that hiding does not work on the Visualizations module (tried both Chrome and IE). Clicking on the legend does nothing.

chaveiro:
It's possible to add a parameter to config if axes will rescaled, will look on that later.
I was not able to reproduce any issue with the visualization module, can you describe the steps?

chaveiro:
A quick doubt on why you want the axis to be rescaled...
I personally prefer the series to hide as a way to clean up the graph when many feeds are visible at the same time. When hided a series rescales the graph o tend to loose track of what was looking at because all the graph may change form.
For multigraph you have a nice solution, 2 y axis (left and right) to add a feed to, just ensure feeds with similar sizes are on the same axis.
(I will take a look on adding optional rescale anyway, just want to point out alternatives.)

software-monkey:
The way I've got my multigraphs set up, the values are usually similar. However, I sometimes get spikes in some values that puts the y-axis way off for the others. I tried to move feeds to different multigraphs, but there is no combination where they are the same scale all the time.

I don't like the 2 y axis: left / right because the spikes don't always happen and when they don't, I want the scale of the two lines to match exactly so that I can compare. The graph makes it look like two values are similar when there are 2 y axis, but the values are way off.

The only solution I'd found before making this change was to move each feed into a separate graph so that I can track each value. I have 24 feeds, which makes that quite annoying. Even then, I cannot compare lines when they are split out. So I'd need many more combinations of lines to get the ability to compare and handle spikes. With the change I checked in, I have a single dashboard with a few multigraphs that lets me see everything. If I get spikes, I can hide those lines and look at the data. 

Another benefit is that I can hide all lines except for one and the scale lets me look for unexpected slight variations (could be a failing device for instance). I don't need a separate raw data graph to do that (or 24 of those in my case).

When using this to clean up the graph when many lines are visible, doesn't the color and x axis help you keep track of where you were on the graph? Same spot left to right, but you might lose some of the whitespace at the top and bottom.



With the Visualization plugin, I used these steps
1) On the top bar of emoncms, select Extra->Visualization
2) Select multigraph for the visualization
3) create a new multigraph and add two feeds or select an existing one.
4) try to hide one of the lines on the preview panel on the right. Doesn't work for me (Windows 7 both Chrome and IE).

chaveiro:
Ok, so i revert to old method, simplified a little but needs refactoring to module in future.

software-monkey:
Did you actually mean to revert this? I picked up the latest code and the toggle functionality isn't working. Looked through the code and it seems to be half set up for flot toggle and half for my way, but neither is fully done so neither works.

Also looks like the line smoothing change I put in timecompare was removed.

Can we revert this?

software-monkey:
Ah, I see you made flot changes too. I'm working on the more scalable solution and didn't have those changes. I'm trying to keep everything in the flot plugin so that future core flot changes are easier to merge.

We don't need to revert for now, but the graph smoothing for timecompare will need to be orrected at some point.

chaveiro:
Hi,the graph smoothing you refer removed data points from graph that are needed to display a graph with some virtual feeds. Such as time discontinuous ones, please dont remove them.
Flot lib maintainer does not maintain it anymore, we can pretty much forget a new official version, so tweeks to flot lib are allowed as long as they follow flot generic functionality and are not risky or create incompatibilities.

software-monkey:
I'm planning on exposing npoints to the user so they can decide. Any concerns with that approach?

I've been able to get everything working by only adding only the flot legend hook. Togglelegend is also down to 100 lines and clearer imo. Rescale on toggle is now a graph option too.

One thing on flot I noticed was some code related to triggering dblclick events that was added. Doesn't appear to be related to hiding lines. What was that for? Is it needed?

Also, when you modified timecompare.js, you made it so that any navigation on the graph recomputes the compare_unit. I wanted it so that only the Day, Week, Month, and Year buttons did that. That way you can zoom in on an area (say a specific day on this week vs last week) without causing the underlying comparison to change. I'm reverting to the previous behavior.

software-monkey:
Submitted a pull for my changes. Take a look: https://github.com/emoncms/emoncms/pull/465

PR #465, opened 2016-02-01
Moved series hiding to flot + cleanup + bugfixes

software-monkey:
Made a number of changes:

1. reworked jquery.flot.togglelegend.js to simplify it and get it to support y-axis rescaling (requested in https://github.com/emoncms/emoncms/pull/459#issuecomment-177555571).
2. changed the hidden legend to include the line color so that it is easier to find the one you want to restore
3. moved timecompare and multigraph to use flot togglelegend for hiding series
4. got togglelegend to work in the visualization tool for multigraph
5. got multigraph tool tips to work in the visualization tool (had to move tool tip code to multigraph.js and did the same for timecompare.js).
6. allowed the user to smooth graphs on timecompare. Just lower the npoints (data points) number. 800 is default, 80 makes a pretty good, smooth graph. This will average values over a time period instead of showing all data
7. cleaned up the code of multigraph and timecompare for consistency (spacing, braces etc).
8. reverted a change to once again allow panning and zooming on timecompare without modifying the timeframes being compared. I like to zoom in on areas of concern to see a breakdown of the data.
9. exposed togglelegend setting for rescale to the plot options (so it can be disabled): legend: { toggle: true }, toggle: { rescale: false },

Needed testing and other notes:

1. I haven't tested this with the latest version of master. Login was failing with error "undefined" and I haven't had time to troubleshoot. I didn't want to delay this pull until I can look at it. Can someone test with the latest?
2. I'm not set up to test on a touch screen. Can someone confirm multigraph and timecompare work on touch screens
3. Some dblClick trigger logic was added for the previous flot checkin. What was this for? Doesn't seem toggle related, but was checked in with toggle changes. I've removed it, but let me know if that is an issue


chaveiro:
Looks ok, dblClick in flot is not used, so full revert is fine.
A minor issue, hidden feeds are not persisted after zoom or screen resize.

software-monkey:
Thanks, I'll take a look when I can (traveling at the moment)

software-monkey:
Take a look now.

Also, rescale as a toggle option didn't make sense as when you zoom it has to rescale to the new data even if some is hidden. So instead I call it the scale option. Values should be "visible" to scale to only visible series or "all" to scale to all series. Default is "visible"

chaveiro:
Looks ok now. Many thanks!

PR #466, opened 2016-02-03
Node variable MQTT topic structure & change of PHP MQTT library

glynhudson:
**Issue**

The old emonHub MQTT topic structure posting a CSV with all the keys for a particular node to topic  `emonhub/rx/[NODEID]/values`. This is not following MQTT recommended practices and this was making it difficult to subscribe to a particular value in other applications (e.g in openHab, nodeRED). 

MQTT best practice states that all MQTT topics should be human readable with one value per topic. 

The [Nodes Module](https://github.com/emoncms/nodes) was used to subscribe to the `emonhub/rx/#` MQTT topic and post the data to Emoncms via the Nodes module interface. 

`RFM69Pi/emonPi > emonHub > MQTT > Emoncms Nodes Module `

**Solution**

1. Modify emonHub to post data to MQTT in the format `[basetopic]/[nodename]/[keyname]` e.g. `emon/emontx/power1`. This new node variable MQTT topic structure can be turned on/off and basetopic set in emonub.conf. The new posting structure is turned on by default in the new [default.emonpi.emonhub.conf](https://github.com/openenergymonitor/emonhub/blob/emon-pi/conf/emonpi.default.emonhub.conf). The old method is implemented by default in [old.default.emonhub.conf](https://github.com/openenergymonitor/emonhub/blob/emon-pi/conf/old.default.emonhub.conf)

2. Use `phpmqtt_input.php` script to subscribe to the base MQTT topic (default `emon/#`) and post to Emoncms Inputs using MQTT topic structure as feed and key name (nodes module not required). *MQTT basetopic to subscribe to can now be set in settings.php. *

`RFM69Pi/emonPi > emonHub > MQTT > Emoncms Inputs`

e.g. `emon/emontx/power1` will result in an input from a node called `emontx` with feed `power1` set to corresponding value. This Input can then be logged to a feed. 

**More Issues!**

It was found that the [bluerhinos PHP MQTT library](https://github.com/bluerhinos/phpMQTT) that phpmqtt_input.php was using was not reliable. This library does not seem to be maintained or developed any more. 

**More Solutions!**

Switch to using  [mgdm PHP-Mosquitto Library](https://github.com/mgdm/Mosquitto-PHP). This required installing some packages and a partial rewrite of phpmqtt_input.php. This library is actively maintained and has proved to be much more reliable. See 'how to install section below'.  

**How this update will effect current users**

If a user is using `phpmqtt_input.php` running as a `mqtt_input` daemon service they will need to update following the guide below. Is is expected that this will only effect a minority of users. MQTT input service was not implemented on any of our pre-build SD card images. The user will also have to update to the latest settings.php

**How to Update**

[Follow MQTT Raspberry Pi Install guide](https://github.com/emoncms/emoncms/blob/dev-mosquitto-php/docs/RaspberryPi/MQTT.md). 

*Note: there is no reason why this solutions should not work on non-Raspberry Pi setups. Non Raspberry Pi support should be improved now the Nodes modules (that was written for emonPi) is not used any more*

**Going Forward**

We will implement this new MQTT topic structure and depreciate the Nodes module on the next emonPi / emonBase pre-build SD card image based on Raspbian Jessie Lite (currently a weeks away from beta release)

**Related Forum Topic**

http://openenergymonitor.org/emon/node/12091

### DIFF
--- a/default.emonpi.settings.php
+++ b/default.emonpi.settings.php
@@ -26,7 +26,7 @@
                           'user'     => 'emonpi',
                           'password' => 'emonpimqtt2016'
                           // Used with scripts/phpmqtt_input.php
-                          'user'=>1,
+                          'userid'=>1,
                           'basetopic'=> 'nodes'                          
                           );
 
--- a/default.emonpi.settings.php
+++ b/default.emonpi.settings.php
@@ -110,8 +110,8 @@
     $default_action = "login";
 
     // Default controller and action if none are specified and user is logged in
-    $default_controller_auth = "nodes";
-    $default_action_auth = "view";
+    $default_controller_auth = "feed";
+    $default_action_auth = "list";
 
     // Public profile functionality
     // Allows http://yourdomain.com/[username]/[dash alias] or ?id=[dash id]
--- a/default.emonpi.settings.php
+++ b/default.emonpi.settings.php
@@ -24,7 +24,11 @@
     $mqtt_server = array( 'host'     => 'localhost',
                           'port'     => 1883,
                           'user'     => 'emonpi',
-                          'password' => 'emonpimqtt2016');
+                          'password' => 'emonpimqtt2016'
+                          // Used with scripts/phpmqtt_input.php
+                          'user'=>1,
+                          'basetopic'=> 'nodes'                          
+                          );
 
 
 //4 #### Engine settings
--- a/default.emonpi.settings.php
+++ b/default.emonpi.settings.php
@@ -27,7 +27,7 @@
                           'password' => 'emonpimqtt2016',
                           // Used with scripts/phpmqtt_input.php
                           'userid'=>1,
-                          'basetopic'=> 'nodes'                          
+                          'basetopic'=> 'emon'                          
                           );
 
 
--- a/default.settings.php
+++ b/default.settings.php
@@ -26,7 +26,7 @@
                           'user'     => '',
                           'password' => '',
                           // Used with scripts/phpmqtt_input.php
-                          'user'=>1,
+                          'userid'=>1,
                           'basetopic'=> 'nodes'
                           );
 
--- a/default.settings.php
+++ b/default.settings.php
@@ -24,7 +24,11 @@
     $mqtt_server = array( 'host'     => 'localhost',
                           'port'     => 1883,
                           'user'     => '',
-                          'password' => '');
+                          'password' => '',
+                          // Used with scripts/phpmqtt_input.php
+                          'user'=>1,
+                          'basetopic'=> 'nodes'
+                          );
 
 
 //4 #### Engine settings
--- a/default.settings.php
+++ b/default.settings.php
@@ -27,7 +27,7 @@
                           'password' => '',
                           // Used with scripts/phpmqtt_input.php
                           'userid'=>1,
-                          'basetopic'=> 'emoncms'
+                          'basetopic'=> 'emon'
                           );
 
 
--- a/default.settings.php
+++ b/default.settings.php
@@ -27,7 +27,7 @@
                           'password' => '',
                           // Used with scripts/phpmqtt_input.php
                           'userid'=>1,
-                          'basetopic'=> 'nodes'
+                          'basetopic'=> 'emoncms'
                           );
 
 
--- a/docs/RaspberryPi/Low-write-mode.md
+++ b/docs/RaspberryPi/Low-write-mode.md
@@ -89,7 +89,7 @@ Save & exit:
 
 Ensure all redis databases have been removed from `/var/lib/redis` with: 
     
-    sudo rm -rf /var/lib/redis/
+   rm -rf /var/lib/redis/*
 
 #### Configure Feedwriter
 
--- a/docs/RaspberryPi/Low-write-mode.md
+++ b/docs/RaspberryPi/Low-write-mode.md
@@ -85,6 +85,10 @@ Comment out all redis saves:
 
 Save & exit:
 
+    sudo service redis-server reboot
+
+**Ensure all redis databases have been removed from `/var/lib/redis`**
+
 #### Configure Feedwriter
 
 Create a symlink to run feedwriter as a daemon and set permissions:
--- a/docs/RaspberryPi/Low-write-mode.md
+++ b/docs/RaspberryPi/Low-write-mode.md
@@ -87,7 +87,9 @@ Save & exit:
 
     sudo service redis-server reboot
 
-**Ensure all redis databases have been removed from `/var/lib/redis`**
+Ensure all redis databases have been removed from `/var/lib/redis` with: 
+    
+    sudo rm -rf /var/lib/redis/
 
 #### Configure Feedwriter
 
--- a/docs/RaspberryPi/MQTT.md
+++ b/docs/RaspberryPi/MQTT.md
@@ -1,50 +1,89 @@
-##Enabling MQTT
-MQTT is a "lightweight" messaging protocol, which enables the publishing of data from emoncms, as well as subscribing to data from
-other connected devices.
-###Preparation
+## Enabling MQTT
+MQTT is a "lightweight" messaging protocol, which enables the publishing of data from Emoncms, as well as subscribing to data from other connected devices. The emonPi uses MQTT to transfer data from emonHub to Emoncms.
 
-Before following this guide, it is essential that emoncms was initially installed by following either the [Raspbian Jessie](readme.md) or [Raspbian Wheezy](install_Wheezy.md) installation guide, or you have used git to install a working version of emoncms on your Raspberry Pi, as well as a MQTT message broker such as [mosquitto](http://mosquitto.org/) installed and running.
+### Preparation
 
-Update emoncms to current version
+Before following this guide, it is essential that emoncms was initially installed by following either the [Raspbian Jessie](readme.md) or [Raspbian Wheezy](install_Wheezy.md) installation guide, or you have used git to install a working version of Emoncms & emonHub on your Raspberry Pi, as well as a MQTT message broker such as [Mosquitto](http://mosquitto.org/) installed and running.
+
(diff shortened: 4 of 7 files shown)


glynhudson:
Pull request has been created. I will leave it open for review for 24hrs before merge

Paul-Reed:
Just read through the changes that have been [made to the Enabling MQTT guide](https://github.com/emoncms/emoncms/blob/dev-mosquitto-php/docs/RaspberryPi/MQTT.md) and wondered how the changes to emonhub.conf affect non-emonpi users (emonbase), as they do not use the OEM version of emonhub?
Has the guide been amended from an emonpi perspective, and if so, it could be confusing for emonbase users.

joyrider3774:
Hi,
I've just spend the last 2 hours trying to get the new mosquittodev branch to work on windows. unfortuantly i could not get it to work but i might look at it tommorow again.

This is what i found out:
we need the php mosquitto extentsion build specifically for windows so *.so files won't work. I managed to find (an older) version here which is prebuilt: http://windows.php.net/downloads/pecl/snaps/mosquitto/0.2.2/ i did not find a newer 0.3.0 prebuild version (on the site above in releases directory there are only logs with failed build info but no actuall builds for the 0.3.0 version)

before you can use those dll's i think pear has to be installed because php mosquitto extension has a dependcy on it, at least it says so on the pcl page here https://pecl.php.net/package/Mosquitto 
I basically used the pear 2 pyrus installer to install pear like so "php pyrus.phar install pear/PEAR-1.10.1" in a directory outside www but inside my wamp directory. 

i then copied the files from the prebuild mosquito extension to the directory "ext" from php inside wamp. added the extension=php_mosquitto.dll line to php.ini and restarted wamp. 

Wamp did not display a warning sign near the module but it seems the module still is not loaded into php for some reason. Since calling the input script gives me the following error : "Fatal error: Class 'Mosquitto\Client' not found in C:\wamp\www\emoncmsnew\scripts\phpmqtt_input.php on line 78" which basically means the class that the extension exposes is somehow not availble in php or at least i think so. i don't see any errors in php.ini either dealing with this extension so it's unclear to my as to why it's not loaded yet, will check that out tomorrow

Also i noticed that the old lib is still being used in Modules\process\process_processlist.php see line 46 actual send is done near line 678. Not sure if the process list (to post an inputs value to mqtt) is still supposed to be using the old library but i thought i'd mention it. 

If it's only the input script that's changing it's rather easy to make a seperate file for windows, basically the old one with the old lib since i had not find any "real" problems with it out except from erroring out sometimes.

glynhudson:
@Paul-Reed emonBase users use the same pre-built SD card image as the emonPi therefore there won't be a problem if they are using our pre-built images. I have created an [old.default.emonhub.conf](https://github.com/openenergymonitor/emonhub/blob/emon-pi/conf/old.default.emonhub.conf) to be used on older pre-build images (without MQTT authentication and using old emonhub CSV MQTT topic structure). I have also added the same `old.default.emonhub.conf` under the name `emonhub.conf` in the emon-pi `emonhub/conf` folder to maintain compatibility 

Of course  if users have chosen to install everything themselves from scratch and use the non emon-pi version of emonhub then there could be issues. They will have to update to the latest version of emon-pi emonhub. I'm not familiar with what the differences are between the two versions of emonhub.  I believe if a user is using a non-emonpi version of emonhub then things should continue to work in the same way after this update to emoncms unless they are using the mqtt input script. I think these users will be in the minority? 

@joyrider3774 thanks for testing php-mosquitto on windows. Probably worth opening an issue on the [php-Mosqutto github](https://github.com/mgdm/Mosquitto-PHP) regarding windows compatibility. 

Well spotted publishing to MQTT via import process [does still use the old MQTT PHP library](https://github.com/emoncms/emoncms/blob/master/Modules/process/process_processlist.php#L46). No issues have been reported using the old library to *publish* to MQTT, it was *subscribing*  to lots of topics that was cusing data to be missed. However I agree it would be best to fully switch totally over to the new library. I will discuss this with @TrystanLea

Paul-Reed:
Glyn, based upon posts in the forum, I would suspect that self-build users are far from the minority and shouldn't really be overlooked, as especially these users have contributed much to emoncms development over the years. The self build guides **have always** specified cloning emonhub from Paul Burnell's repo, and it was only when emonpi was launched that the 'emonpi' version of emonhub came into being to service the needs of emonpi. There has never (as far as I'm aware) been any suggestion that self builds should use the emonpi emonhub.
Can I ask you to consider resolving these issues first, otherwise the changes may cause problems for a good many users.

glynhudson:
@Paul-Reed @joyrider3774 as a fall-back we have added back the old version of the mqtt input script in the scripts folder renamed to `old.phpmqtt_input.php` https://github.com/emoncms/emoncms/commit/b12956a30d02674d6e1d6f751ec9fbdd31ec55ef. If a users does want to continue using the old PHP MQTT lib then they can do by renaming this file.

Paul-Reed:
I don't think that self builds will want to continue using the old version
of the script, as they will want to use the long awaited MQTT updates.
Using the old script will alienate self builds, frustrate development,
create support difficulties, and it also undermines the v9 ethos of 'one
common version'.
Surely a better approach to consider would be to delay this merge, make a
decision about which version of emonhub you want users to use, update the
installation guides and provide some guidance to users how to switch to the
preferred emonhub version.
Then roll this out so everyone can move forward together.

@Paul-Reed <https://github.com/Paul-Reed> @joyrider3774
<https://github.com/joyrider3774> as a fall-back we have added back the old
version of the mqtt input script in the scripts folder renamed to
old.phpmqtt_input.php b12956a
<https://github.com/emoncms/emoncms/commit/b12956a30d02674d6e1d6f751ec9fbdd31ec55ef>.
If a users does want to continue using the old PHP MQTT lib then they can
do by renaming this file.

—
Reply to this email directly or view it on GitHub
<https://github.com/emoncms/emoncms/pull/466#issuecomment-179844659>.

glynhudson:
Agree, it would be great to have a single version of emonHub going forward. Apart from the chance of path for emonhub.conf and the recent update to the MQTT topic structure what is the main difference between the two? I'll discuss with @TrystanLea he was more involved than I was when the fork was created.

glynhudson:
Just taken a look and spoke to @TrystanLea; it looks like the emonhub from the Dev (default) branch Paul Burnell's repo does not support MQTT. I can see some mention of MQTT in the experimental branch. 

I can't see how this pull request would adversely effect a non emon-pi emonhub user since they will not be using MQTT from emonHub. Is it just a documentation issue we need to resolve? How about I add clarification to he documentation I added to notify the user of the differences in the emonhub versions? 

If a user want's to use the new topic structure they will need to switch to using emon-pi emonhub. If they don't want to then al long as they don't change their emonhub.conf file everything will continue to work as before. 

@TrystanLea and I we will open conversation with @pb66  regarding merging to have a single version of emonhub going forward.

Paul-Reed:
The docs is the main issue as it assume emonpi. I've just committed https://github.com/emoncms/emoncms/commit/a0da22fe445ad5dcfcabaf9574a9877a75758854 which splits up emonpi/emonbase in the guide. It's a bit rushed, so please edit or revert as you wish.

glynhudson:
Hi Paul, 

Thanks for the update. Just had a read through, couple of things:

Unless the non-emonpi user uses the 'old' mqtt input script then they will also need to install the new PHP-mqtt library. The old script with the old library can be used with:   

`cp /var/www/emoncms/scripts/phpmqtt_input.php cp /var/www/emoncms/scripts/new.phpmqtt_input.php`

`cp /var/www/emoncms/scripts/old.phpmqtt_input.php cp /var/www/emoncms/scripts/new.phpmqtt_input.php`

before `cd /etc/init.d && sudo ln -s /var/www/emoncms/scripts/mqtt_input`

However I would suggest that non-emonpi users use the new script with the new better php-mqtt library. Once installed the new script will function in exactly the same way as the old script with the added flexibility of being able to set the basetopic and account user ID from setting.php. There is nothing specific to the emonPi about the new script. 

I think the confusion is due to me including emonPi specific emonhub documentation in the document. How about I move all mentions of emonhub to the OEM emonhub repo documentation and then just provide an optional link?

Paul-Reed:
Aaah, yes I'll add the new script commands this evening if you haven't done
so already.
As for moving the references to emonhub, I'll leave that for you to decide,
but once a decision is made about a standardising emonhub, they will need
to be added back in again as I presume that they will then also apply to
emonbase too.

glynhudson:
I've just made some changes. Feel free to modify as you see fit https://github.com/emoncms/emoncms/commit/e2d80c92c74aaf3d6c794e4bf1a5ca320650defa

glynhudson:
Actually we want to deleate the contents of /var/lib/redis not the folder itself. I'll update docs to `rm -rf /var/lib/redis/*'

Paul-Reed:
But...
`sudo rm /var/lib/redis/*`
Does just delete the contents, not the folder....
`sudo rm -rf /var/lib/redis/*`
deletes folders and files within redis, but there are no folders, just files?

glynhudson:
The desired outcome is that /var/lib/redis/ is totally empty `sudo rm -rf /var/lib/redis/*` will ensure all files and folder in /var/lib/redis/ are deleted. It's just a failsafe in case there is a rouge folder in there. Either will work for what we want though. Just trying to be on the safe side

Paul-Reed:
It all looks good to me, and works well here.

glynhudson:
Been merged, thanks for your help and updating the forum: http://openenergymonitor.org/emon/node/12104

chaveiro:
On https://github.com/emoncms/emoncms/pull/471 i reverted the user id key location from settings.php to the phpmqtt_input.php script file because it's not related to emoncms instalation but to the instance of that script running for a particular user.
Best is to make the phpmqtt_input.php script generic for multiuser support, this will require mqtt autentication. Use of apikey or device key?
What kind of auth is used usualy for mqtt across the internet?

PR #484, opened 2016-02-25
Master contribution (5 Fixes)

RDCH106:
- Fixed error "Cannot read property 'scrollHeight' of null" in Admin Panel
- Fixed gethostbyaddr(): Address is not a valid IPv4 or IPv6 address --> Warning massage
- Fixed localization codes
- ~~Added 'Reset to ONE' process as opposite of 'Reset to ZERO'~~
- Other minor fixes

### DIFF
--- a/Modules/process/process_processlist.php
+++ b/Modules/process/process_processlist.php
@@ -156,7 +156,6 @@ class Process_ProcessList
         $list[58] = array(_(" / source feed"),ProcessArg::FEEDID,"divide_by_source_feed",0,DataType::UNDEFINED,"Virtual", 'desc'=>"");
         $list[59] = array(_("1/ source feed"),ProcessArg::FEEDID,"reciprocal_by_source_feed",0,DataType::UNDEFINED,"Virtual", 'desc'=>"");
         
-        $list[60] = array(_("Reset to ONE"),ProcessArg::NONE,"reset2one",0,DataType::UNDEFINED,"Misc", 'desc'=>"The value '1' is passed back for further processing by the next processor in the processing list.");
         return $list;
     }
 
@@ -196,12 +195,6 @@ class Process_ProcessList
          return $value;
     }
     
-    public function reset2one($arg, $time, $value)
-    {
-         $value = 1;
-         return $value;
-    }
-
     public function reset2original($arg, $time, $value)
     {
          return $this->proc_initialvalue;
--- a/Modules/process/process_processlist.php
+++ b/Modules/process/process_processlist.php
@@ -155,6 +155,8 @@ class Process_ProcessList
         $list[57] = array(_(" * source feed"),ProcessArg::FEEDID,"multiply_by_source_feed",0,DataType::UNDEFINED,"Virtual", 'desc'=>"");
         $list[58] = array(_(" / source feed"),ProcessArg::FEEDID,"divide_by_source_feed",0,DataType::UNDEFINED,"Virtual", 'desc'=>"");
         $list[59] = array(_("1/ source feed"),ProcessArg::FEEDID,"reciprocal_by_source_feed",0,DataType::UNDEFINED,"Virtual", 'desc'=>"");
+        
+        $list[60] = array(_("Reset to ONE"),ProcessArg::NONE,"reset2one",0,DataType::UNDEFINED,"Misc", 'desc'=>"The value '1' is passed back for further processing by the next processor in the processing list.");
         return $list;
     }
 
@@ -191,6 +193,12 @@ class Process_ProcessList
     public function reset2zero($arg, $time, $value)
     {
          $value = 0;
+         return $value;
+    }
+    
+    public function reset2one($arg, $time, $value)
+    {
+         $value = 1;
          return $value;
     }
 
--- a/Modules/vis/visualisations/timestoredaily.php
+++ b/Modules/vis/visualisations/timestoredaily.php
@@ -30,7 +30,7 @@
 </style>
 
 <div id="vis-title"></div>
-<div id="#stats"></div>
+<div id="\#stats"></div>
 
 
 <div id="placeholder_bound" style="width:100%; height:400px; position:relative; ">
--- a/Modules/vis/visualisations/timestoredaily.php
+++ b/Modules/vis/visualisations/timestoredaily.php
@@ -30,7 +30,7 @@
 </style>
 
 <div id="vis-title"></div>
-<div id="\#stats"></div>
+<div id="#stats"></div>
 
 
 <div id="placeholder_bound" style="width:100%; height:400px; position:relative; ">
--- a/locale.php
+++ b/locale.php
@@ -77,13 +77,13 @@ function set_lang($language)
         $lang=$language[0];
 
         switch($lang) {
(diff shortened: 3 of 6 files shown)


Modules/process/process_processlist.php
```diff
@@ -155,6 +155,8 @@ public function core_process_list()
         $list[57] = array(_(" * source feed"),ProcessArg::FEEDID,"multiply_by_source_feed",0,DataType::UNDEFINED,"Virtual", 'desc'=>"");
         $list[58] = array(_(" / source feed"),ProcessArg::FEEDID,"divide_by_source_feed",0,DataType::UNDEFINED,"Virtual", 'desc'=>"");
         $list[59] = array(_("1/ source feed"),ProcessArg::FEEDID,"reciprocal_by_source_feed",0,DataType::UNDEFINED,"Virtual", 'desc'=>"");
+        
+        $list[60] = array(_("Reset to ONE"),ProcessArg::NONE,"reset2one",0,DataType::UNDEFINED,"Misc", 'desc'=>"The value '1' is passed back for further processing by the next processor in the processing list.");
```
chaveiro: Dont agree on this, 'one' is not a special number different from any other than zero. So you can get the same behaviour for all numbers by setting to 'zero' and adding a fixed number with process '+' after.

Modules/vis/visualisations/timestoredaily.php
```diff
@@ -30,7 +30,7 @@
 </style>
 
 <div id="vis-title"></div>
-<div id="#stats"></div>
+<div id="\#stats"></div>
```
chaveiro: What is this supposed to do?

locale.php
```diff
@@ -77,13 +77,13 @@ function set_lang($language)
...
 
         switch($lang) {
-            case 'cy': $lang='cy_GB'; break;
-            case 'da': $lang='da_DK'; break;
-            case 'en': $lang='en_EN'; break;
-            case 'es': $lang='es_ES'; break;
-            case 'fr': $lang='fr_FR'; break;
-            case 'it': $lang='it_IT'; break;
-            case 'nl': $lang='nl_NL'; break;
+            case 'cy_GB': $lang='cy_GB'; break;
```
chaveiro: Tested on firefox and IE ?

chaveiro:
Thanks for the pull request, i've a made a few questions on the code where i have doubts.

Modules/process/process_processlist.php
RDCH106: I explain my use case. I have a feed value and I need to check if the feed value is more/less than x value to set other virtual feed to true/false value. The value true/false is for alert activation. Setting to ONE and ZERO the behaviour is as a boolean logic. With boolean virtual feed value, I can show a led alert in a Dashboard.

I attach you the following logic with the process logic:

reset_to_one

The logic is possible to use in a input logic and virtual feeds.

reset_to_one_leds

**It is very usefull for binary alerts generation.**

locale.php
RDCH106: Yes. I tested the code in Firefox 44 and in IE 10.

Modules/vis/visualisations/timestoredaily.php
RDCH106: It could be a mistake. Let me to check the reason for this change.

RDCH106: Yes it is a mistake. I have reverted the change. Sorry!

glynhudson:
Hi guys, what's the status of the pull request? Do you have any further comments @chaveiro? Can it be merged?

chaveiro:
Hi glin, personaly dont agree with new processlist 'reset to one' because it's redundant, 'reset to zero' and 'reset to null' already exist, setting to any other value can be done with 'reset to zero' and '+ value'.  The rest i have no problem as long as it works fine.

glynhudson:
@RDCH106 would it be possible to remove the  processlist 'reset to one' to changes from this request? 

@TrystanLea do you have any thoughts on this?

jeremypoulter:
For what it is worth I also have a requirement similar to @RDCH106. I think the problem is the way the way the conditionals are implemented (if something skip next) there will be an increased desirer to do a lot more functionality in one step rather than multiple steps to make things a little less confusing. So I think there is value in the Reset to 1, especially for 'driving' the LED widgets, would a compromise be a 'reset to value' to just set to an arbitrary value?

That being said, IMHO, I don't think the problem of driving the LED widgets is best solved in the input processing. I think it would be a lot more use to extend the LED widget to allow specification of the condition which it should be on.

RDCH106:
I can remove from the processlist the `reset to one` item, but internally I will continue using this operation because **I need to operate with leds (zero and one values)**.

I think that the `Reset to desired value` is not bad idea and it offers more flexibility and it does not overload the reset function.

What do you think?

The other good alternative has been proposed by @jeremypoulter in the last paragraph.

Paul-Reed:
I'm with Chaveiro on this, because with a bit of careful thought this can be achieved through existing processes, and is therefore not really needed.
For example, the below is a quick example of returning a 0 or a 1 depending upon the criteria set.
criteria
It returns a '0' if the input value is less than 2000, or a '1' if the value is 2000 or above.

RDCH106:
Of course that it is possible to do that, but for a newcomer is not intuitive first Reset to Zero and then use "+" operator.

@Paul-Reed in your example you use "Reset to NULL" to set to zero. It is unsafe operation despite of it works because null is a special value that PHP is assuming as 0.

Paul-Reed:
I agree that it's not intuitive for new users, but that could be said for many other process functions, and bloating the function list with every conceivable scenario is probably not the best option when users can achieve the same result using the tools that we already have.  
@RDCH106 I don't see a problem in using "reset to NULL" in this context. The PHP Manual describes NULL to 'represents a variable with no value' and not '0' (zero), which is exactly what I want when I start a new calculation within an existing process list.
If "Reset to NULL" was an unsafe operation, I'm pretty sure the senior developers would not have added it!

RDCH106:
@Paul-Reed you has said! Null is a meta value that represents no value. What is the result of null + 1?? You have done that but adding an input value. If you want a result value to operate with it, use  Reset to Zero. But it does not matter because it is offtopic.

I defend the Reset to One or more generic operatoe Reset to desired value because this operation is 1 unit cost and in the other case are 2 unit cost. One unit to set the value (Reset to Zero) and another one to operate and set value because the operation is calculated against the first operator. When it is a scenario with a heavy computation of process lists, this topics are relevant.

Paul-Reed:
Rubén, can we bring this to some conclusion.  
**Chaveiro** (the lead developer of v9) doesn't support the Reset to one.
**Glyn** (OEM project lead) has asked if you could remove the Reset to one process from this request.  
If you remove this process, your pull request will be considered.  
Paul

RDCH106:
OK 
I have removed the `Reset to One`.

For the persons who wants to use the function I provide the following patch:
[Reset_to_One.zip](https://github.com/emoncms/emoncms/files/181240/Reset_to_One.zip)

Paul-Reed:
Thanks Rubén

Paul

PR #488, opened 2016-03-10
add port to mysqli function

Hotfirenet:


chaveiro:
It wont upset anything if supported, maybe some personal requirement of the contributor.

Hotfirenet:
Because I use mysql in a container Docker I map ports otherwise.

PR #510, opened 2016-04-21
Fixed timezone bug copied over from emoncms.org

TrystanLea:


tgmaxx:
Hi Trystan,
This problem is still existing in #539. My kwhd bars disappear. I am using MySQL for all data storage and Eastern USA timezone. I made a note on the commit to Glyn.

Thanks,
Tom

PR #516, opened 2016-05-06
Update profile.php

AndyTaylorTweet:
Re-named the QR code to "App Integration QR Code"
Added some text beneath the QR code to give the user some pointers about what to do with it.
Added link to the app on Google Play


Paul-Reed:
That's great Andy, thanks for doing this work. The QR code setup should help a lot of users.

PR #517, opened 2016-05-06
Update feed_model.php

AndyTaylorTweet:
Added line 456 to mitigate divide by 0 errors;
This is *NOT* the correct way to fix this issue, this is a reference only to help some of you understand where the issue is.
I have not worked out yet where the variable $outinterval is supposed to be populated from, but at this point in the code its either null or 0 - and this un turn causes the divide by 0 errors.

My code simply detects when $outinterval is less than 1 and sets this variable to 1 before its used. (on line 457 in this version).

Applying my code appears to fix the connection issues with the android app, fixing the divide by zero cleaned that right up, but my fix may have other consequences (although quite unlikely).

### DIFF
--- a/Modules/feed/feed_model.php
+++ b/Modules/feed/feed_model.php
@@ -453,6 +453,7 @@ class Feed
                 
                 // Merge buffered data into base data timeslots (over-writing null values where they exist)
                 if ($engine==Engine::PHPFINA || $engine==Engine::PHPTIMESERIES) {
+                    if ( $outinterval < 1 ) { $outinterval = (int) 1; } //Fix div by zero errors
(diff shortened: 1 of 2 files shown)


glynhudson:
This issue that caused the connection issues with the Android app has been fixed with an update to the Android app 1.2.2: https://community.openenergymonitor.org/t/android-app-connection-error/250/8

However I will keep this open for @TrystanLea to take a look

AndyTaylorTweet:
This latest commit is nothing to do with the first one, I'm going to close the PR and make a new one for the latest commit - the changes to feed_model.php were easy enough to find that I could do so again anyway.

PR #521, opened 2016-05-14
Update profile.php

AndyTaylorTweet:
Add CSS Button style (copy to clipboard buttons)
Change Button label (CSS made the buttons longer - causing unwanted wrap)


Paul-Reed:
Thanks Andy. It's good to see the buttons changed to match existing design code.

PR #522, opened 2016-05-14
Update admin_main_view.php

AndyTaylorTweet:
Modified the tables around the log output to fill the screen width - this really helps on small screen devices (like phones).

Also added so Pi specific code, so the page detects if it is running on Raspberry Pi (Based on the Ethernet address of the on board interface - this is by far the most reliable detection method) and displays some information about the CPU temperature and the two filesystems' free space.

These will not display on non Pi hardware.

### DIFF
--- a/Modules/admin/admin_main_view.php
+++ b/Modules/admin/admin_main_view.php
@@ -109,8 +109,6 @@ if(is_writable($log_filename)) {
 }
 ?>
             </p>
-            
-            <pre id="logreply-bound"><div id="logreply"></div></pre>
         </td>
         <td class="buttons">
 <?php if(is_writable($log_filename)) { ?>
@@ -125,6 +123,7 @@ if(is_writable($log_filename)) {
           
         </td>
     </tr>
+    <tr><td colspan="2"><pre id="logreply-bound"><div id="logreply"></div></pre></td></tr>
 <?php
 }
 if ($allow_emonpi_admin) {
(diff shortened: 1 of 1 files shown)


Paul-Reed:
Andy, like the fuller width log window, and especially the addition of the IP address can be useful for users wanting to use the command line, however, can you have another look at the disk space indicator.
I've no doubt that this would work fine on a emonpi. as it uses the  /home/pi/data directory for writeable data, however not so on a emonbase which by default stores data in /var/lib and doesn't have a /home/pi/data directory causing the admin_main_view.php to throw an error.

Perhaps a check to see if /home/pi/data exists?

Paul

AndyTaylorTweet:
This version (I don't have the words to tell you how awkward this was) removes the reliance on the EmonPi image; so it should work on home brewed raspberry pi builds too.

In short it now lists the total size and free space for all mounted block devices - see what you think.

Paul-Reed:
Works well here, I've attached a print below.
Just a couple of things to consider;

1) Do you think the heading should be 'Pi' instead of 'EmonPi' (because the data relates to the pi, and nothing to do with emonpi/emonbase)??

2) For users with a external drive, a lot of Mb's are reported, would it be better if scaled to Gb's?

andy

3) If I compare the disk space using df, I get conflicting figures;

pi@raspberrypi:/var/www/emoncms $ df -h
Filesystem      Size  Used Avail Use% Mounted on
/dev/root       9.5G  4.7G  4.4G  52% /
devtmpfs        459M     0  459M   0% /dev
tmpfs           463M     0  463M   0% /dev/shm
tmpfs           463M  6.3M  457M   2% /run
tmpfs           5.0M  4.0K  5.0M   1% /run/lock
tmpfs           463M     0  463M   0% /sys/fs/cgroup
/dev/mmcblk0p1   60M   20M   41M  34% /boot
/dev/sda2       456G   58G  399G  13% /mnt/media
tmpfs            93M     0   93M   0% /run/user/1000

For your consideration Andy.

Paul

AndyTaylorTweet:
Paul, comments taken on board - many changes rolled into this version;
New look including disk space graphics (coloured to match the base style)
Double checked the math in use, you are right there are small discrepancies between "df -h" and my script, I suspect that df calculates disk usage slightly differently than the in-built PHP function. I've seen "df" report a file system at more than 100% full before now (that was a bad day I can tell you) so its probably using a different calculation (maybe kilo = 1000 rather than 1024?) - so I will investigate that a little more.

Pi hardware (CPU monitoring at the moment) separated from the filesystem info, also renamed to just "Pi".

File system information will now show on all systems since it is in no way reliant on the Pi hardware.

Paul-Reed:
This is a really useful addition Andy, and the graphics show at a glance the remaining disk space. 
Thanks for the additional work.

andy

Paul

glynhudson:
Fantastic, this is great. Super useful thanks

PR #524, opened 2016-05-16
Small tidy of user_model and locale

inverse:
- Small tidy of user_model

https://github.com/emoncms/emoncms/blob/master/Modules/user/user_model.php#L554 - Not sure what should be done here? When this method is being called it sets `$result` to the void value. Should this be passing back the saved data instead?

- Made consistent return on `directoryLocaleScan` in locale

### DIFF
--- a/Modules/user/user_model.php
+++ b/Modules/user/user_model.php
@@ -14,13 +14,32 @@ defined('EMONCMS_EXEC') or die('Restricted access');
 
 class User
 {
+    /**
+     * @var mysqli
+     */
     private $mysqli;
+
+    /**
+     * @var Rememberme
+     */
     private $rememberme;
+
+    /**
+     * @var bool
+     */
     private $enable_rememberme = false;
+
+    /**
+     * @var Redis
+     */
     private $redis;
+
+    /**
+     * @var EmonLogger
+     */
     private $log;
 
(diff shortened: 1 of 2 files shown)


glynhudson:
Thanks. Are there any functional changes here for just code style tidy? @chaveiro would you like to comment on this?

inverse:
Only changes effecting error conditions so shouldn't happen anyway. It does use PHP 5.5  short array syntax but as previously discussed you're targetting 5.6 so shouldn't be an issue. I think It'll be worth documenting this on the readme.

chaveiro:
Change in  https://github.com/emoncms/emoncms/pull/524/files#diff-fa279668f942a3cddad4091c1bf2bc99R25 makes result empty (not null) if there are no Locale directory, but in https://github.com/emoncms/emoncms/blob/master/Modules/user/profile/profile.php#L31  we check if NULL was returned.

It seems to change the behaviour with this commit. Did you test it?

About https://github.com/emoncms/emoncms/blob/master/Modules/user/user_model.php#L554   function seems incomplete, $result var is not used and can be removed.

inverse:
Okay - please disregard this PR. There is a few errors in there which I've seen now I've setup a vagrant box to test things out.

PR #525, opened 2016-05-17
Update admin_main_view.php

AndyTaylorTweet:
1. Add some extra information for the "Pi" section (System Load and RAM usage)
2. Modify the disk usage setup so that it will ONLY produce output on *nix based systems that have the "lsblk" command. (should help avoid errors if the command does not exist).


glynhudson:
Nice, works great. Thanks.

PR #526, opened 2016-05-17
Bug in vis/Stacked[solar].php with day vs. daily

fab33:
I didn't check if the bug still exist in other visualisations.


fab33:
Just find this commit https://github.com/emoncms/emoncms/commit/b2304bd83f4c009072bea45721a4b9b572e9c8b0 with all the file concern with this bug.

TrystanLea:
thanks much appreciated

PR #527, opened 2016-05-18
Added vagrant development box

inverse:
Added vagrant development box
- Accessible on localhost:8080 on the host machine
- Included xdebug for debugging
- Based on Ubuntu 14.04 with php 5.6

Could integrate Redis etc but as a starting point this worked for me to get going. Opinions?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ settings.php
 old.settings.php
 emoncms.log
 .idea
+.vagrant/
 
 # Ignore all modules except core ones
 Modules/*
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,0 +1,48 @@
+# -*- mode: ruby -*-
+# vi: set ft=ruby :
+
+Vagrant.configure(2) do |config|
+
+    config.vm.box = "bento/ubuntu-14.04"
+    config.vm.network "forwarded_port", guest: 80, host: 8080
+    config.vm.synced_folder ".", "/var/www/html", :owner => "www-data", :group => "www-data"
+
+    config.vm.provider "virtualbox" do |vb|
+        vb.name = "emoncms"
+        vb.cpus = 1
+        vb.memory = "2048"
+    end
+
+    config.vm.provision "shell", inline: $script, keep_color: true
+end
+
+$script = <<SCRIPT
+    export DEBIAN_FRONTEND=noninteractive
+    sudo debconf-set-selections <<< 'mysql-server mysql-server/root_password password vagrant'
+    sudo debconf-set-selections <<< 'mysql-server mysql-server/root_password_again password vagrant'
+
+    sudo apt-get update
+
+    sudo apt-get install -y php5 php5-xdebug php5-mysql apache2 mysql-server
+
+    sudo printf "zend_extension=xdebug.so\ndebug.max_nesting_level = 400\nxdebug.remote_enable=on\nxdebug.remote_connect_back=on\nhtml_errors=1\nxdebug.extended_info=1" > /etc/php5/mods-available/xdebug.ini
+
+    sudo a2enmod rewrite
+    sudo a2enmod mime
+    sudo a2enmod deflate
+    sudo a2enmod filter
+
+    cd /etc/apache2
+
+    sed -i 's/AllowOverride None/AllowOverride All/g' apache2.conf
+
+    sudo service apache2 restart
+
+    cd /var/www/html
+    sudo rm index.html
+    cp default.settings.php settings.php
+    sed -i 's/_DB_USER_/root/g' settings.php
+    sed -i 's/_DB_PASSWORD_/vagrant/g' settings.php
+
+    mysql -uroot -pvagrant -e "CREATE DATABASE IF NOT EXISTS emoncms"
+SCRIPT
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -34,7 +34,7 @@ $script = <<SCRIPT
 
     cd /etc/apache2
 
-    sed -i 's/AllowOverride None/AllowOverride All/g' apache2.conf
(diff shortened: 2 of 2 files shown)


glynhudson:
Nice! I've never used Vagrant, how is it different from Docker? There are lots of Emoncms docker images floating the web, I would really like to get an 'official' docker Emoncms image made. Totally sold on the advantages of docker, however I don't feel that I know enough about to make an Emoncms image.

inverse:
I've never used Docker but from my little understanding it's just containerized and not a full virtual isolated environment. If you wish to go down that route feel free to close this PR down. It helped me get up an environment to work with emon for development nicely though :)

glynhudson:
There is no harm in offering both options, this is just a file in the repo.
It's doing no harm. I will merge this PR.

- sent from my mobile device

elyobelyob:
I like this lots. I'm more a vmware person, I guess that would be easy
enough to add in.

Test data? It's a real time product, I'm not sure what live test data to
use. Especially if developing off-line.

Happy to help with ideas.

Nick

elyobelyob:
But basically, the swap of provider is like this ... (taken from another
project) ...

Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
  config.vm.provider "virtualbox" do |v|
    config.vm.box = 'centos6.4-min.vbox'
    config.vm.box_url =
'https://github.com/2creatives/vagrant-centos/releases/download/v0.1.0/centos64-x86_64-20131030.box'
    auto_correct = true
  end

  config.vm.provider "vmware_fusion" do |v|
    config.vm.box = 'base'
    config.vm.box_url =
'https://oss-binaries.phusionpassenger.com/vagrant/boxes/latest/ubuntu-14.04-amd64-vmwarefusion.boxx'
  end

  config.vm.network "private_network", ip: "192.168.111.111"


....

inverse:
You don't even have to go that far - you could just do the provider specific things. Adding something like this to the VagrantFile should do the job. I haven't tested this though.

```
config.vm.provider "vmware_fusion" do |v|
	v.vmx["memsize"] = "2048"
	v.vmx["numvcpus"] = "1"
end
```

elyobelyob:
The boxes are specific for the provider. Certainly were with the ones I was
using at work.

Any suggestions on data? I've been wondering about off-line development,
and data for me is the key to this, generating some example feeds that show
the range of data, but also auto-populate - perhaps using a cron job.

Nick

inverse:
We could configure the VM like you said via a cron to call a CLI script to 'fake' data points. :+1:

Paul-Reed:
Just asking... but shouldn't this be separated from the emoncms/emoncms core files, and occupy its own directory within the emoncms repo, in a similar way to useful scripts, backup, app, device, etc so that it can be added 'as an optional module' if users wish to use it?

Paul

elyobelyob:
You're right. There should be an emoncms repo for all specific areas. I
could see this growing quickly. Best to separate early.

I'll let you guys decide on a decent name. I've already struggled.

emoncms/*

Anyway, yes, it could contain bespoke stuff for developers, definitely
should stay out of the core.

Keep the emoncms core entirely separate.

p.s. can I actually delete the emoncms/event or at least I will delete from
my own repo. That could be moved into an archived area of this new thing
we're talking. But I'm under the impression report module is now ready to
retire. Not used it myself for a few iterations.

Spring clean!

inverse:
Alternatively we could have it and all relevant configuration in it's own repo? That you just need to have emoncms accessible via some config/mapping. Opinions?

Paul-Reed:
Why not emoncms/vagrant?
That would enable the addition of a brief README to assist in installation, and also allow users to clone the vagrant file(s) directly into their emoncms installation directory, so the path would be;
/var/www/emoncms/vagrant/VagrantFile

Paul

elyobelyob:
I think it should be more generic emoncms/developmentarea or something.
Scripts put into this area can be used by those not using Vagrant as well,
but who are involved in development. The test data population scripts I'm
thinking of would be an example.

Not having a development setup has made it difficult for me to dev in the
past. So, this is potentially a big step forward.

Nick

Paul-Reed:
Sounds a good idea Nick. Similar to the emoncms/usefulscripts directory except more focussed upon development work?

Paul

glynhudson:
Agree, best not in the core. I like emoncms/vagrant, this keeps it self explanatory. I can see a emoncms/docker coming soon! @davehun

elyobelyob:
Only problem is that it'll be a whole repo for pretty much one file. My
suggestion would cover all development files including shared stuff that we
think and dream up in the future ;)

Paul-Reed:
Nick, the advantage in it having it's own repo is that if users want to clone it, they will only get the vagrant files, and not a whole lot of other development ideas. Take the usefulscripts repo for example, users are unable to clone the one script that they want, and have to clone the whole repo with 9 other scripts too, which they may not want. Another benefit is as I said above, it could include it's own README to assist users.

There is already a [emoncms/development repo](https://github.com/emoncms/development), which has been pretty much unused over the past 12 months, and you'll note that 4 development ideas were moved from there, to their own repos last April https://github.com/emoncms/development/commit/66904dedb36552587aacceb356a6589e56f5751a

If you have some really early dev ideas, perhaps the existing emoncms/development repo could be used, and as development progresses, they could be moved from there to their own repo's, like what has been done before?

Paul

inverse:
I agree with @Paul-Reed, having the vagrant configuration in it's own repo may be a better solution as later we may wish to break out the configuration into files or provision with puppet/chef which could add a lot more files.

Paul-Reed:
@glynhudson in view of the comments above, are you moving this to its own repo to facilitate future development.

glynhudson:
Done: https://github.com/emoncms/vagrant

@inverse has been given write access to this repo.

PR #528, opened 2016-05-21
Graph clean up, add reboot button.

AndyTaylorTweet:
Cleaned up the code around the graphs for RAM / disk usage.
Show RAM graph for Pi hardware only.
Show /var/log specifically when its mounted on tempfs

The reboot button does require a supporting change in `/etc/sudoers`
`,/sbin/shutdown` should be appended to the www-data line to allow Apache to be able to execute the shutdown command.

The reboot code has been setup such that it can ONLY be executed on Pi hardware, this is a safety measure to help reduce any risks on production class (read non raspberry pi) hardware.

### DIFF
--- a/Modules/admin/admin_main_view.php
+++ b/Modules/admin/admin_main_view.php
@@ -1,6 +1,6 @@
 
 
-<?php global $path, $emoncms_version, $allow_emonpi_admin, $log_enabled, $log_filename, $mysqli, $redis_enabled, $redis, $mqtt_enabled, $feed_settings, $rebootPi;
+<?php global $path, $emoncms_version, $allow_emonpi_admin, $log_enabled, $log_filename, $mysqli, $redis_enabled, $redis, $mqtt_enabled, $feed_settings, $shutdownPi;
 
   // Retrieve server information
   $system = system_information();
@@ -182,7 +182,7 @@ if ($mqtt_enabled) {
 <?php
 } // Raspberry Pi Detection and additions.
 if ( exec('ifconfig | grep b8:27:eb:') ) {
-              if (isset($rebootPi)) { shell_exec('sudo shutdown -r now 2>&1'); }
+              if (isset($shutdownPi)) { if ($shutdownPi == 'reboot') { shell_exec('sudo shutdown -r now 2>&1'); } elseif ($shutdownPi == 'halt') { shell_exec('sudo shutdown -h now 2>&1'); } }
               echo "              <tr><td><b>Pi</b></td><td>CPU Temp</td><td>".number_format((int)exec('cat /sys/class/thermal/thermal_zone0/temp')/1000, '2', '.', '')."&degC".chkRebootBtn()."</td></tr>\n";
                 define("ramTotal", "ramTotal");
                 define("ramUsed", "ramUsed");
@@ -308,10 +308,20 @@ function backup_log_update() {
     }
   });
 }
+$("#haltPi").click(function() {
+  if(confirm('Please confirm you wish to shutdown your Pi, please wait 30 secs before disconnecting the power...')) {
+    $.post( location.href, { shutdownPi: "halt" } );
+  }
+});
+
 $("#rebootPi").click(function() {
   if(confirm('Please confirm you wish to reboot your Pi, this will take approximately 30 secs to complete...')) {
-    $.post( location.href, { rebootPi: "1" } );
+    $.post( location.href, { shutdownPi: "reboot" } );
   }
+});
+
+$("#noshut").click(function() {
+  alert('Please modify /etc/sudoers to allow your webserver to run the shutdown command.')
 });
 </script>
 <?php //Disk Size function
@@ -334,7 +344,11 @@ function chkRebootBtn(){
   $chkReboot = shell_exec('sudo shutdown -k --no-wall 2>&1'); //Try and run a fake shutdown
   if (stripos($chkReboot, "scheduled ") > 0) {
     shell_exec('sudo shutdown -c --no-wall'); //Cancel the fake shutdown
-    return "<button id=\"rebootPi\" class=\"btn btn-info btn-small pull-right\">"._('Reboot')."</button>";
+    return "<button id=\"haltPi\" class=\"btn btn-info btn-small pull-right\">"._('Shutdown')."</button><button id=\"rebootPi\" class=\"btn btn-info btn-small pull-right\">"._('Reboot')."</button>";
+  }
+  else {
+    return "<button id=\"noshut\" class=\"btn btn-info btn-small pull-right\">"._('Shutdown Unsupported')."</button>";
   }
 }
+
 ?>
--- a/Modules/admin/admin_main_view.php
+++ b/Modules/admin/admin_main_view.php
@@ -182,9 +182,8 @@ if ($mqtt_enabled) {
 <?php
 } // Raspberry Pi Detection and additions.
 if ( exec('ifconfig | grep b8:27:eb:') ) {
-              if (isset($rebootPi)) { shell_exec('sudo shutdown -r now'); }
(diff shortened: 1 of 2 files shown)


Paul-Reed:
Hi Andy, is there a different approach to trigger a reboot?
The change to /etc/sudoers could be implemented in the emoncms image files as a permanent change , but what about the emonbase users?
Anyone pulling this update would expect the 'Reboot' button to just work, without having to make OS  changes.

Paul

AndyTaylorTweet:
I understand exactly, the short version is that in order to reboot the Pi, you need root privileges;
I could prompt the user for those, but again that is going to cause more trouble than its worth.

I was trying to write some check code so that if you cannot execute the command successfully you get some browser feedback, so far I can't get that idea to work, but this strikes me at about the best compromise; I may even be able to hide the button if the command is not executable.

I'll take whatever feedback you all have on those...

AndyTaylorTweet:
Slight modification with a new function that works out if the Apache user can actually run the shutdown command, the button only shows on systems that are configured to allow apache to run /sbin/shutdown.

Hows that?

AndyTaylorTweet:
Ahh - dont commit this - found a small error - correcting that now....

AndyTaylorTweet:
Ok that last commit should be clean, working etc.

Paul-Reed:
Can't you get root privileges for example by calling a bash script to effect the reboot.
IMO rather than hiding the reboot button for some users, it would be preferable to have the facility available for all.

**For debate** - did you consider having a 'shutdown' button instead of 'reboot'? A shutdown button would allow users to safely shutdown their system, for example to move it elsewhere, and also effect a reboot simply by removing and reconnecting power after a shutdown.

Paul

AndyTaylorTweet:
Expanding this to add shutdown is simple enough, yes I considered it but I was focussed on the reboot only functionality first.

Could I call a bash script? - Yes, I could, but then I would also need the bash script held in the web directory, thats not such a good move, and over all would have the same effect, the bash script would be run as the user that apache runs as, with no root privileges, so the net effect would be 0. The base issue is that the www-data user (that apache runs as) does not have root privileges by design, this I should point out is a good thing.

Paul-Reed:
Andy, if you look at the emoncms/scripts directory you'll see that emoncms already uses lots of bash scripts, which are housed within html.

I thought that unless otherwise specified, bash scripts always ran with root privileges by default?

Paul

glynhudson:
Hi Andy,

Nice work, just tested and it works as advertised 👍 

After checking out your PR I did not initially have the change made in sudoers, I could not see the reboot button. After adding in the line to sudoers the button appeared. Maybe the button could be grayed out with a link to a readme for instructions of how to make the changes to sudoers. 

As far as the emonPi / emonBase users are concerned it would be easy to make the change to sudoers as part of an emonPi update. The few lines of bash required to do this could be appended to `emoncmsupdate` script which is called as part of emonPi update process:

https://github.com/openenergymonitor/emonpi/blob/master/emoncmsupdate

Might you be able to add a couple of lines of bash to check for the change in suders and append if not to the emoncmsupdate file?

Yes, I think also expanding to include a Shutdown button would be useful. But reboot is the big one. This will be very useful.

I'm happy for this PR to be merged when you're ready. I'm away for the next couple of days so might be a bit slow responding. I will leave merged into the `stable` branch until the corresponding change has been made in emoncmsupdate to make changes to sudoers.

AndyTaylorTweet:
Modified to add shutdown option;
Added better handling for situations where shutdown isn't (yet) supported.
So now if the web server cannot run shutdown, the user gets a button option, but on clicking that they are told why it doesn't work.

Obviously text can be changed on the buttons and in the messages.

Paul-Reed:
> The reboot button does require a supporting change in /etc/sudoers
,/sbin/shutdown should be appended to the www-data line to allow Apache to be able to execute the shutdown command.

Just trying your latest but there is no www-data line in /etc/sudoers in Raspbian Jessie.

Paul

AndyTaylorTweet:
It's already there for the emonpi image users, because there has clearly already been some work done to allow the apache user to run some tools / apps as root.

Add this:
www-data ALL=(ALL) NOPASSWD:/sbin/shutdown

As has been suggested by Glyn, I could work up some code for the emoncms update script to check for and add that where required.

Paul-Reed:
Emonbase users don't use the update script, so it would not benefit them.
The update script is for emonpi users only.

Paul

pb66:
Please make this feature optional in settings.php so that it is can be default on in emonPi/Base and off/optional for the rest.

Not all users are using an emonPi or emonBase, not all are even using a Pi for the server or for that fact linux. and some of us just don't want a reboot option available to "emoncms only" admins when there are other things running on the server, grayed out or not. This is a system admin funtion.

I see why this is useful to "One man and his emonPi" users but please don't make it a mandatory feature that may not work or even appear broken or incomplete rather than opted-out of.  

Paul

glynhudson:
>Emonbase users don't use the update script, so it would not benefit them.
The update script is for emonpi users only.

This is incorrect, emonBase users DO run the emoncms update script. The
update script works exactly the same as on the emonpi. The only difference
is the RFM69Pi firmware does not get updated. A firmware update is
attempted but them fails due to baud rate difference.

- sent from my mobile device

Paul-Reed:
> This is incorrect, emonBase users DO run the emoncms update script

Not in my experience @glynhudson 
The update script is written from the perspective of emonpi, and includes things which are not installed by default within a standard emonbase installation (via [emoncms installation guides](https://github.com/emoncms/emoncms/tree/master/docs/RaspberryPi])), but are present by default in a emonpi image. Such as - app, wifi, dashboard, config (isn't the emonpi variant of emonhub a requirement of 'config', whilst the standard emonbase uses @pb66 emonhub version?).
The script stops if any of those are not present.

It also force installs Modules, which may not be wanted, such as Postprocess and backup. And in the case of 'Postprocess' for some reason it's not installed within emoncms/Modules, and the update script does not install a symlink for it to run, nor create it's postprocess.settings.php file (which also contains the wrong location for 'emonbase' data files).

Line 150 copies default.emonpi.settings.php to settings.php - instead of the emonbase default.settings.php

Line 157 tries to run a script which is not present on a emonbase, only emonpi.

Paul

glynhudson:
Hi Paul,

Ah, we're both right! It's just an issue of what is an 'emonBase'.. . We
supply the same pre built image with the emonPi as we do with' emonBase'
purchased from the shop: That is emonBase = Raspberry Pi + RFM69Pi.

What you are referring to as your emonBase is a custom raspberry pi
install. Whatever we call it, yes your correct the update script will not
work for your custom install. Maybe we can work on converging so we're all
running the same software albeit with customisations.. But this is another
topic. The update script should more intelligently deal with customisation
to settings.php.

For most users using the pre built image is the recommended starting point,
we would not recommend a user run emoncms on a RaspberryPi without all the
low write modifications on the pre built image due to sd card life span
issues. Here is the full image build guide;

https://github.com/openenergymonitor/emonpi/blob/master/docs/SD-card-build.md

It's a significant amount to build the image from scratch!

- sent from my mobile device

elyobelyob:
Hi, just wondering what the need for the reboot button is? I'm not running
on a Pi and am interested to know the purpose.

Thanks

AndyTaylorTweet:
It does what it says on the tin, reboots the host.
In a conversation with Glyn some time back, I asked him if there was anything specific he wanted implemented, this was on the list, so I bashed out the code.

I should point out that I made the assumption (and we know what is said about those) that anyone running EmonCMS on something other than a Pi, probably wouldn't appreciate the button showing up in the interface, so the code is deliberately crippled on non Raspberry Pi hardware (it won't even show up in the interface, and even *iF* you post the correct information to the page, still nothing will happen).

glynhudson:
In a perfect world there would be no need for a reboot button I.e system
would be 100% reliable and infinity stable.  But as things stand and
especially when developing on the Pi it's useful to easily be able to
trigger a reboot. In the same way as it's possible to reboot a Internet
router via the Web interface.

- sent from my mobile device

PR #529, opened 2016-05-24
Multigraph, new options - stack, refresh

blaal02:
Add new options on Multigraph:
Lines/Bars can be defined as 'stacked'. E.g. can be used to show usage on individual branch circuits combining to a total usage.
The graph can be set to 'auto refresh' periodically. So that a dashboard can be configured to automatically update with current data. (The automatic update will be disabled if the graph has been scrolled so that the current time is no longer the end of the graph)
Show/Hide Legend. Either for simple graphs where the title is all that's needed. Or on a dashboard where multiple graphs have the same legend & only needs shown once.

### DIFF
--- a/Modules/vis/Views/multigraph_edit.js
+++ b/Modules/vis/Views/multigraph_edit.js
@@ -11,6 +11,8 @@ var multigraphs=[];
 var multigraphs_name=[];
 var movingtime = false;
 var showtag = true;
+var autorefresh = 0;
+var showlegend = true;
 
 var baseElement = "#box-options";
 
@@ -55,6 +57,16 @@ function draw_multigraph_feedlist_editor(){
   else
     showtag = true;
 
+  if (typeof multigraph_feedlist[0] !== 'undefined' && typeof multigraph_feedlist[0]['autorefresh'] !== 'undefined')
+    autorefresh = multigraph_feedlist[0]['autorefresh'];
+  else
+    autorefresh = 0;
+
+  if (typeof multigraph_feedlist[0] !== 'undefined' && typeof multigraph_feedlist[0]['showlegend'] !== 'undefined')
+    showlegend = multigraph_feedlist[0]['showlegend'];
+  else
+    showlegend = true;
+
   var out = "";
   out += '<div id="myModal" class="modal hide" tabindex="-1" role="dialog" aria-labelledby="myModalLabel" aria-hidden="true" data-backdrop="static">';
   out += '<div class="modal-header"><button type="button" class="close" data-dismiss="modal" aria-hidden="true">×</button><h3 id="myModalLabel">Delete Multigraph</h3></div>';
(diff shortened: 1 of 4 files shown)


glynhudson:
Does this not duplicate the stacked visualization already present in vis? 

I've just tested you PR and it seems to have broken multigraph, non on my multi graphs load and I can't create new ones. I can see the 'stacked' option but without the multigraph data loading I an unable to test. I think we would prefer to keep multi graph separate from stacked vis. What do you think @TrystanLea

fab33:
Same here, this PR doesn't work for me, no data loading.
But I think this option is useful for many reasons : 
- stacked vis can only show 2 feeds. For my need I would like to have 4 or 5 stacked feeds
- stacked vis have some bugs (show kWh/d for kW feed, initial view start years before first data, can't see last 2 days of feed data ...)
- missing options in stacked vis like scale

Perhaphs a good option is to had stacked option in new graph visualisation. Actually you can have bar viex but feed are overlaped.

blaal02:
Apologies; new to github, I was making some changes manually - and one "slipped" a couple of lines! Should be fixed now.
I have a dozen+ CT & 3-4 Temp/Humidity sensors, so Multigraph is the only visualization that's much use to me (and it sounds like a few others too, probably anyone monitoring branch circuits).
We could make this into a new graph type - but that would mean duplicating 95% of the current multigraph code and having to maintain both copies.

glynhudson:
>Lines/Bars can be defined as 'stacked'. E.g. can be used to show usage on individual branch circuits combining to a total usage.

Seems to be working, however I'm not quite sure how this is expected to work. Could you post a screen cast / screen grab illustrating operation?

blaal02:
Certainly!
Here's the definition of my main usage page.
The individual branch usages are stacked (basically, yes - added together), then the total (black) line is not (in theory all the branches should stack to the total).
And this graph automatically refreshes every 5 mins. (300 secs.) with new data
usage_def1
usage_def2
And here's my main dashboard that combines this graph with a couple of others. The daily usage graph uses the same legend as the first graph, so there's one I've got the legend turned off for.
dashboard

TrystanLea:
Thanks @blaal02 looks great! and tidy commit changes

glynhudson:
Do you think you could post your screen grabs and a quick explanation about this new feature on our emoncms community forums to notify users https://community.openenergymonitor.org/c/emoncms

fab33:
Working now But I don't find how to make stacked bar like your "Total Usage" graph.

blaal02:
The 'stack' code doesn't change anything there. The type of graph you get in multigraph (line vs. bar) is controlled by the type of feed.
In my case, I am collecting an explicit feed (DAILY PHPTIMESERIES) for my kWh/d - multigraphing those gives you a bar chart (that can now be stacked). There's some new code been added to many of the other visualizations (from Trystan I believe) to dynamically calculate daily usage (kWh/d) from deltas of a kWh feed - but this isn't available in multigraph.
(I've been wondering about implementing this kind of thing as a mod to Virtual Feeds, that could then be used in all visualizations - but not done anything yet).

PR #532, opened 2016-05-26
Inject dependencies into router and clean up a little

inverse:
- Inject all dependencies into router instead of relying on global objects
- Clean up code a little bit for readability

Do you guys have an opinion on using something like composer for autoloading/dependency management? 

Also what is your stance on adding tests?

### DIFF
--- a/route.php
+++ b/route.php
@@ -117,8 +117,8 @@ class Route
     /**
      * @return bool
      */
-    public function isRouteDefined()
+    public function isRouteNotDefined()
     {
-        return !empty($this->controller) && !empty($this->action);
+        return !$this->controller && !$this->action;
     }
 }
--- a/route.php
+++ b/route.php
@@ -17,17 +17,47 @@ defined('EMONCMS_EXEC') or die('Restricted access');
 
 class Route
 {
+    /**
+     * @var string
+     */
     public $controller = '';
-    public $action = '';
-    public $subaction = '';
-    public $format = "html";
 
-    public function __construct($q)
+    /**
+     * @var string
+     */
+    public $action = '';
+
+    /**
+     * @var string
+     */
+    public $subaction = '';
+
+    /**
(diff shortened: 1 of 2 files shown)


route.php
```diff
@@ -62,18 +92,33 @@ public function decode($q)
...
+        }
+        if (count($args) > 1) {
+            $this->action = $args[1];
+        }
+        if (count($args) > 2) {
+            $this->subaction = $args[2];
+        }
+
+        if (in_array($requestMethod, ['POST', 'DELETE', 'PUT'])) {
+            $this->method = $requestMethod;
```
inverse: I can't see where method is ever used in code? Should we just strip it out?

chaveiro: You have to take inconsideration all modules that might exist that are not part of core repo.

inverse: That makes sense, I won't change this behavior then :+1:

glynhudson:
Thanks, is this PR now ready to be merged?

>Do you guys have an opinion on using something like composer for autoloading/dependency management?

No experience of using composer, sounds like a good idea. Does it work in a similar way to Ruby gems and bundler? What dependencies are you suggesting could be managed with composer?

inverse:
Seems good from the testing I did. I haven't tried any of the other plugin though. Functionally it's the same though.

Totally the same - Just pull in 3rd party libraries and also handles autoloading - https://getcomposer.org/doc/01-basic-usage.md

glynhudson:
Just tested this and it seems after applying your changes I could load load the [Wifi module](https://github.com/emoncms/wifi). Attempting  to navigate to `http://emonpi/emoncms/wifi`, just redirected me back to `http://emonpi/emoncms/feed/list`

inverse:
Fixed :+1:

glynhudson:
Working now 👍

PR #534, opened 2016-06-01
Router further tidy

inverse:
- Rework router to pass in base path
- Added composer for class loading and dependencies
- Added unit test for router

Updated vagrant script to pull in composer and install dependencies: https://github.com/emoncms/vagrant/commit/e706cfc92f2ca04bdbebd75ba4ab19dab035dc37

Is this something you guys would be interested in going further with?


src/Core/Route.php
```diff
@@ -45,35 +46,33 @@ class Route
...
     }
 
     /**
      * @param  string $q
      * @param string $documentRoot
+     * @param string $basePath
      * @param string $requestMethod
      */
-    public function decode($q, $documentRoot, $requestMethod)
+    public function decode($q, $documentRoot, $basePath, $requestMethod)
```
inverse: Perhaps the naming isn't the best?

inverse:
Failing on PHP 5.4 and 5.5 because of PHPUnit 5.3 requires 5.6 minimum. We could run an older version PHPUnit 4.8. Opinions?

chaveiro:
You moved  route.php → src/Core/Route.php  is this really necessary ? Why not Lib/core path?
Watch out for unnecessary case changes, follow existing patterns.
If route.php is moved, then all other 'core' files on root should be moved along. Dont you agree?
This is starting a major refactor of essential core files that should be much tested because every change is critical.
Recommend to do all tests on a side branch and commit when stable.

inverse:
Lib seems like a strange place for such files, but if it's the pattern that you guys believe should be then I can change it. To me it feels like src is a good place to start to try and move things towards.

As for moving other files, I didn't know if this approach would be something that would be taken and as such I didn't want to invest lots of time for nothing. 

I also think committing to master frequently is a good practice as having long running branches can often leave things stagnating and cause conflicts as others develop. Also it helps identify issues early.

Paul-Reed:
Even though the 'Master' branch is for development, there is a significant
user group that are subscribed to that branch, and breaking it would cause
problems for a lot of users.
Committing to a temporary branch would allow for staged testing.

Paul

glynhudson:
>I vote ditching 5.4 as it's EOL and 5.5 is only security fixes till August.

I have removed 5.4 and 5.5 tests from Travis test https://github.com/emoncms/emoncms/commit/0616bba8aa65e0289b180767dafa8f87fd0bb789

inverse:
@glynhudson I added back that before_script check

TrystanLea:
What are the implications of adding composer here for existing installs? If this is merged will any further steps be needed beyond a git pull to update changes?

inverse:
The implication of bringing composer in right now will be that all installations will require composer and to execute `composer install`. How would be the best way to phase in such a requirement into this project?

TrystanLea:
will it affect existing users if composer is not installed? what kind of additional disk space are we talking about?

Paul-Reed:
Why not merge it into an "experimental" branch and then the full
implications can be fully explored without exposing existing users.

Paul

inverse:
@TrystanLea It will not work if the user doesn't have composer installed - unless we commit the whole vendor directory, which isn't a good thing to do. 

@Paul-Reed There could be an experimental branch where this could be explored. I just worry about things stagnating with long lived branches. 

@glynhudson Are PHP5.4, 5.5 now supported versions?

chaveiro:
This is getting interesting, i can give advice on any doubt you might have on actual emoncms code.
A few questions. 
1 - Will composer support bring managed modules to emoncms ?
(I mean to deal with optional emoncms modules, not external libs/dependencies)

2 - Can an Admin page be created to list / install  / uninstall /  update emoncms modules ?

3 - If user has no root access will he be able to install composer?

inverse:
1. If emoncms was to use composer across other modules a user could update their dependencies to include the said module and we had the relevant loading mechanisms in place than yes. But we would need to rethink the way in which currently the process only looks within Modules. Symfony does this nicely with how bundles are registered. 

Or did I misunderstand your question?

2. Like how WordPress works? This has nothing to do with composer, but I guess you could produce some kind of require/update trigger but I've not seen this done.

3. Sure - just download the phar file and execute that, it doesn't require root access

chaveiro:
1 - I'm was refering to emoncms modules (see 2)

2 - Yes like WordPress, it's is a long wanted feature.
We have some modular support already on emoncms, all folders inside ./Modules are modules that if exist will bring new functionality. Modules/dashboard for example is a module.
A special kind of module also supported are widgets, they are discovered on this paths (Modules/module_name/widget/widgetname/) and will bring specific modules widgets (when that module_name is installed) to the dashboard module. Additional info https://openenergymonitor.org/emon/node/11209

What is missing is a way of managing this modules: an Admin page to list / install / uninstall / update emoncms modules.

inverse:
For this to work you would need some form of module discovery - I guess via an API or leveraging packagist if we ensure that all modules have [keywords](https://getcomposer.org/doc/04-schema.md#keywords) defined.

Composer could help with some aspects. For example if emoncms and modules follow [semver(http://semver.org/) we could safely upgrade to the latest compatible version of modules. But safely is very much dependent on the module vendor respecting semver ;)

What do you think of this?

TrystanLea:
Im happy to merge in the router changes they look fine. Id prefer to keep the Lib/ path. 

Personally I'm happy with the current way installation works and Im weary of the work involved and all the related documentation changes, image rebuilds etc that would be required with a large change of this nature. 

At the same time I'm developing emoncms less at the moment as Im generally happy with where it has got to, Im working more on other parts of the project. But I dont want that to hold emoncms development back if others such as yourself @inverse are keen to progress things. 

I think @Paul-Reed suggestion of this being in an experimental branch is a good idea, especially if there is a group of advanced users/developers who would be keen to explore it. I think the current branch of emoncms would need to stay as is because of the number of emonpi/raspberrypi/self hosted systems that all update from it.

inverse:
Okays - that makes sense - Is anyone else interested in working on an experimental branch that leverages composer and other things?

glynhudson:
Yes, I'm interested. I don't have the same knowledge of emoncms and php as @TrystanLea but I have been doing my best to setup a docker container for emoncms. The main pain point has been installing the php  dependances. I would be interested to learn know this could be a achieved using composer and move away from pear which seems abandoned. Would you be interested in helping refining the docker container? The idea is to give us a common multi platform dev environment which can be setup from scratch in minutes and can be expanded to include test data and simulated inputs. 

https://github.com/emoncms/emoncms-docker

I'm away from home for this coming week, therfore apologies for slow reply. The forum is probably the best place for this discussion 

https://community.openenergymonitor.org/c/emoncms

PR #535, opened 2016-06-02
EmonCMS 4.71 2016-06-02

chaveiro:
Bugfixing, code cleaning and patterns assert.


Modules/vis/visualisations/common/vis.helper.js
```diff
@@ -41,38 +41,7 @@ var view =
   {
     this.start = ((new Date()).getTime())-(3600000*24*time);	//Get start time
     this.end = (new Date()).getTime();	//Get end time
-  },
-  
-  'round_interval':function(interval)
```
TrystanLea: This is being used by several visualisations, was this mistakenly removed?

Modules/feed/feed_controller.php
```diff
@@ -55,7 +55,7 @@ function feed_controller()
                 if ($feed->exist($feedid)) {  // if the feed exists
                    $f = $feed->get($feedid);
                    if ($f['public'] || ($session['userid']>0 && $f['userid']==$session['userid'] && $session['read'])) {
-                       $result[$i] = 1*$feed->get_value($feedid); // null is a valid response
+                       $result[$i] = $feed->get_value($feedid); // null is a valid response
```
TrystanLea: will this return the value with speech marks? I think the intention of the *1 was to return as a number rather than string

Modules/feed/engine/PHPFina.php
```diff
@@ -336,10 +336,8 @@ public function get_data($name,$start,$end,$interval,$skipmissing,$limitinterval
             }
             
             if ($value!==null || $skipmissing===0) {
-                // Re-enable see https://openenergymonitor.org/emon/node/11260
-                // if ($time>=$start && $time<$end) {
```
TrystanLea: It might be useful to leave this comment in for now, I nee to look at this again to ensure the behaviour is correct

Modules/feed/feed_controller.php
chaveiro: The answer should return numeric values and "null".
Multiply by 0 will make null return as 0 that is wrong.

Modules/vis/visualisations/common/vis.helper.js
chaveiro: I only saw it being used on multigraph.js, it disrupts touch usage causing small jumps.
And is bad code to use ifs that way, should use if-else.

Modules/feed/engine/PHPFina.php
chaveiro: No problem

Modules/vis/visualisations/common/vis.helper.js
TrystanLea: Here's the forum thread where I added this to fix an issue with the multigraph: https://openenergymonitor.org/emon/node/12490, dont mind too much about the use of ifs in this case - although perhaps it could rewritten by iterating over an array of allowed intervals..

Modules/feed/feed_controller.php
TrystanLea: fine yes, need to double check if the value is being returned with speech marks, this could be fixed further upstream if better.

Modules/vis/visualisations/common/vis.helper.js
chaveiro: Ok i saw the post.
So this is only showing in multigraph because it displays the previous value just before loading the new data asynchronously.
But the real reason is the underlay data that coming wrong. Mysql for example does not seems to have this problem.
I'd like the fix to be applied to the engine and not the graphic because it will bite us later on other usecases.
 Why not always return the previous available timestamp at the engine timestamp slot instead of averaging to to the requested start time ?

Modules/feed/feed_controller.php
inverse: If you're wanting a numeric value, why don't you cast the result? e.g. (float)$feed->get_value($feedid)

chaveiro: Just casting will not work because NULL values are valid data.
A null means engine had no data, so in graphics you know there is no data (a gap appears) instead of it joining the last data points.
Also virtual feeds make use of null and there are specific processors dealing with null.

inverse: Isn't that similar to 1*null = 0 vs (float)null = 0.0?

chaveiro: Yes, both are wrong :)

Paul-Reed:
Why is this failing Travis, is Travis still not configured correctly, or is there a inherent problem with this merge?

inverse:
@Paul-Reed It could be that @chaveiro hasn't pulled upstream changes from this branch and is missing test.php which is causing Travis to bail on the phpunit test.php line

Modules/vis/visualisations/common/vis.helper.js
TrystanLea: Sorry havent had another chance to look at this, but I think we need to keep track of the effect of removing this. I dont see it as an issue with the engine as its more the way the request is designed to return data closest to request parameters defined by start time and interval. 
Rounding the interval is also nice in that the interval of the data in the graph adhere to values that are more likely to be an integer number of the base data interval.

PR #536, opened 2016-06-06
Encapsulate redis configuration

inverse:
I think we should start to try and wrap some of the configuration settings into classes to provide more structure and make it easier to slowly move away from `global $blah` declarations. I went with the Redis settings as they are fairly straight forward to wrap.

I maintained `$redis = false` to keep BC with all modules so it shouldn't impact anything. Also Redis has been setup on the vagrant scripts so you can test this out with that :+1: 
## Todo
- Add tests ;D
## To Discuss
- Do you think this is a good idea?
- Since we should be using namespaces, where should such files live? I stuck these in `/Libs` for now, but is this where we want to keep things?
- Autoloading (I vote following PSR-0 or PSR-4, but we should really use composer to handle this)
- Testing,  we need to organize the code file structure e.g. we could have `/Libs/Config/RedisConfig.php` and `/Libs/Test/Config/RedisConfig.php`. Where the test structure mirrors the Libs folder.

### DIFF
--- a/Lib/Config/RedisConfig.php
+++ b/Lib/Config/RedisConfig.php
@@ -1,0 +1,103 @@
+<?php
+
+namespace Emoncms\Config;
+
+class RedisConfig
+{
+    /**
+     * @var string
+     */
+    private $host;
+
+    /**
+     * @var int
+     */
+    private $port;
+
+    /**
+     * @var string
+     */
+    private $auth;
+
+    /**
+     * @var string
+     */
+    private $prefix;
+
+    /**
+     * @var bool
+     */
+    private $isRedisEnabled;
+
+    /**
+     * @param bool $isRedisEnabled
+     * @param string $host
+     * @param int $port
+     * @param string $auth
+     * @param string $prefix
+     */
+    public function __construct($isRedisEnabled, $host, $port, $auth, $prefix)
+    {
+        $this->isRedisEnabled = $isRedisEnabled;
+        $this->host = $host;
+        $this->port = $port;
+        $this->auth = $auth;
+        $this->prefix = $prefix;
+    }
+
+    /**
+     * @return string
+     */
+    public function getHost()
+    {
+        return $this->host;
+    }
+
+    /**
+     * @return int
+     */
+    public function getPort()
+    {
+        return $this->port;
+    }
+
+    /**
+     * @return string
+     */
+    public function getAuth()
+    {
+        return $this->auth;
+    }
+
+    /**
+     * @return bool
+     */
+    public function hasAuth()
+    {
+        return !empty($this->auth);
+    }
+
+    /**
+     * @return string
+     */
+    public function getPrefix()
+    {
+        return $this->prefix;
+    }
+
+    /**
+     * @return bool
+     */
+    public function hasPrefix()
+    {
+        return !empty($this->prefix);
+    }
+
+    /**
+     * @return bool
+     */
+    public function hasValidConfig()
+    {
+        return $this->isRedisEnabled && !empty($this->host) && !empty($this->port);
+    }
+}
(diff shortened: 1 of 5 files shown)


Lib/Config/RedisConfig.php
```diff
@@ -0,0 +1,103 @@
...
+    /**
+     * @param bool $isRedisEnabled
+     * @param string $host
+     * @param int $port
+     * @param string $auth
+     * @param string $prefix
+     */
+    public function __construct($isRedisEnabled, $host, $port, $auth, $prefix)
+    {
+        $this->isRedisEnabled = $isRedisEnabled;
```
inverse: Ideally we should be able to drop this if the other settings are empty, but baby steps :dancer:

Lib/Config/RedisConfig.php
```diff
@@ -0,0 +1,103 @@
+<?php
+
+namespace Emoncms\Config;
```
inverse: Or should this live under `Emoncms\Redis\Config`?

Lib/Redis/Redis.php
```diff
@@ -0,0 +1,94 @@
...
+
+    /**
+     * Get Redis connection. Return false if connection not setup.
+     *
+     * @return \Redis|false
+     */
+    public function getRedis()
+    {
+        if (!$this->isConnected()) {
+            return false;
```
inverse: Feels a bit hacky returning mixed types, but keeps BC with existing implementations :+1:

TrystanLea:
This looks nice, I'm not that familiar with namespaces apart from an awareness that it is good practice and helpful in larger projects. The changes look relatively minimal and as long as the general use of redis is not encapsulated through another layer Im happy. i.e as long as you can call $redis->get, set, etc directly. If your happy to test and ensure the transition over to it is smooth Im happy to support this.

inverse:
Namespaces are certainly a good thing to have and help with managing and organizing code. 

The change is pretty minimal and moves the bulk of the redis initialization into the Redis class which will simplify index.php

inverse:
@TrystanLea do you have any comments about the discussion points? E.g. autoloading etc?

TrystanLea:
@inverse sorry for no update on this. Im not really familiar enough with composer, autoloading etc to comment. I tried testing your changes and got the following error which I guess is related.   

PHP Warning:  require(/var/www/master/vendor/autoload.php): failed to open stream: No such file or directory in /var/www/master/index.php on line 19

Where is the best place to start?

inverse:
@TrystanLea You shouldn't need composer to get this working as I included the require statements directly.

Your error looks like it's trying to require the composer autoloader. I don't think I included that reference in this PR.

TrystanLea:
@inverse I tried your branch 'route-test-tidy' but got the above errors. Should I try replacing:

    require __DIR__ . '/vendor/autoload.php';

with something else?

inverse:
@TrystanLea Ahh that branch required composer to be installed.

If you want to use it you'll need to setup composer. See https://getcomposer.org/doc/00-intro.md

then run `composer install`

TrystanLea:
Thanks @inverse, I need to learn more about namespaces and composer and use them on smaller test projects before I'm really in a position to merge this with an good understanding of what I'm merging. I think its one I will have to come back to once I've had the chance to do this or let someone else take this forward.

inverse:
@TrystanLea No problem :+1: Feel free to close this if you think it's adding noise

TrystanLea:
Thanks @inverse I will leave this here for now, as inspiration for further development once the right time for it arises, I appreciate the input on this!

inverse:
Feel free to reach out if you have any questions :)

inverse:
@TrystanLea makes sense ;)

inverse:
See you added dev composer dependencies now though

PR #539, opened 2016-06-18
hot fix 

glynhudson:
API helper url change is not needed with latest routing


Modules/vis/visualisations/common/vis.helper.js
```diff
@@ -41,38 +41,7 @@ var view =
   {
     this.start = ((new Date()).getTime())-(3600000*24*time);	//Get start time
     this.end = (new Date()).getTime();	//Get end time
-  },
```
TrystanLea: Is the removal of this going to bring up multigraph issues again?

PR #542, opened 2016-06-23
is_nan throws an error if used on empty string

BenK22:
This error occurs when creating a virtual feed comprised of newly created feeds resulting in the list of feeds not displaying.


glynhudson:
Thanks 👍

PR #545, opened 2016-06-28
Issue with que_input_controller

unreg:
Issue: blank list of inputs with use que_input_controller.

Feature:

- added _input_view_ in _que_input_controller_;
- removed limits for nodeid;
- added _set_processlist_ for subaction "set";
- changed readme.


TrystanLea:
Thanks for this, good to have it updated, great!

PR #547, opened 2016-07-01
Enable environment-variable settings in default.settings.php

spikeheap:
This change allows settings to be defined as environment variables, but takes the existing defaults if they are not present. This enables Docker (and other services) to provision emoncms without needing to alter the default settings file at all.


chaveiro:
Hi. This kind of chances should be made in process_settings.php not settings.php files.

PR #555, opened 2016-09-09
New endpoint for feed list with start of day, month, year values added

mattjgalloway:
# RFC

This started as a [question over on the forums](https://community.openenergymonitor.org/t/incrementing-feeds-daily-monthly-yearly-values/1491).

I'm posting this as an RFC right now. I'm not certain that this is the right way to do it. We could add this information to `list`, but my concern is speed. If you have redis, it's cached at least, but not everyone has that.

This adds a new endpoint called `listdmy` which is the same as `list`, but with values for the feed at the start of the day, week and year added in. So this returns something that looks like this:

```
{
    "id": "3",
    "userid": "1",
    "name": "use_kwh",
    "datatype": "1",
    "tag": "emonpi",
    "public": "0",
    "size": "99999",
    "engine": "5",
    "processList": "",
    "time": "1473448379",
    "value": "100",
    "dvalue": {
      "time": "1473375600",
      "value": "95"
    },
    "mvalue": {
      "time": "1472684400",
      "value": "46"
    },
    "yvalue": {
      "time": "1472156140",
      "value": "0"
    }
  }
```

Current limitation is that it only supports PHPFINA. But that might be OK.

### DIFF
--- a/Modules/feed/engine/PHPFina.php
+++ b/Modules/feed/engine/PHPFina.php
@@ -277,6 +277,46 @@ class PHPFina
         }
         return false;
     }
+    
+    /**
+     * Get array with last day, month or year time and value from a feed
+     *
+     * @param integer $feedid The id of the feed
+     * @param integer $time The time to get
+     * @param bool $clip If $time is before the start or after the end, clip to the start/end
+    */
+    public function valueattime($feedid, $time, $clip)
+    {
+        if (!$meta = $this->get_meta($feedid)) return false;
+        $meta->npoints = $this->get_npoints($feedid);
+        
+        $pos = round(($time - $meta->start_time) / $meta->interval);
+        if ($clip) {
+            if ($pos < 0) {
+                $pos = 0;
+                $time = $meta->start_time;
+            } else if ($pos >= $meta->npoints) {
+                $pos = $meta->npoints - 1;
+                $time = $meta->start_time + ($pos * $meta->interval);
+            }
+        }
+        
+        $value = null;
+        
+        if ($pos >= 0 && $pos < $meta->npoints) {
+            // read from the file
+            $fh = fopen($this->dir.$feedid.".dat", 'rb');
+            fseek($fh,$pos*4);
+            $val = unpack("f",fread($fh,4));
+
+            // add to the data array if its not a nan value
+            if (!is_nan($val[1])) {
+                $value = $val[1];
+            }
+        }
+        
+        return array('time'=>$time, 'value'=>$value);
+    }
 
     /**
      * Return the data for the given timerange
--- a/Modules/feed/feed_controller.php
+++ b/Modules/feed/feed_controller.php
@@ -63,6 +63,12 @@ function feed_controller()
             // Export multiple feeds on the same csv
             // http://emoncms.org/feed/csvexport.json?ids=1,3,4,5,6,7,8,157,156,169&start=1450137600&end=1450224000&interval=10&timeformat=1
             $result = $feed->csv_export_multi(get('ids'),get('start'),get('end'),get('interval'),get('timeformat'),get('name'));
+        } else if ($route->action == "listdmy") {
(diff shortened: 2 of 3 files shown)


mattjgalloway:
For an example using this, see my [commit on my emoncms dashboard fork](https://github.com/mattjgalloway/emoncms-dashboard/commit/42274ae74ab0e9d2b8ec6f8abfc8722ffa2ec9fb).

Paul-Reed:
I'm not sure if this is the right approach either Matt, let's defer for @TrystanLea to decide...

Paul

mattjgalloway:
@TrystanLea - sorry for pinging if you're not back from holidays yet. I just wondered if you had any thoughts on this? It's very useful to have IMO. I don't think this is necessarily the right way to add it though.

Other options are

  1. A parameter on `list` to enable adding this.
      * Simpler. Just as backwards compatible.
      * Clutters the `list` endpoint.
  2. Just add it anyway, always, on `list`.
      * The simplest.
      * Makes every `list` a bit slower. Granted not much most likely (untested - do we have performance tests?).
  3. A separate endpoint that just gets the DMY values for a given feed ID and not all the usual feed fields as well.
      * Less duplication (not the code, but rather endpoint semantics).
      * Harder to deal with. Consumers would then have to request twice.

TrystanLea:
Hello Matt, sorry for lack of engagement on this. My preference would be for a separate endpoint, as the request is feed type specific to accumulating kwh/wh. 

For the apps such as myelectric and mysolar the last day is calculated via a last value feed list request and a specific request to the start of the day or month - or just the result of a daily data request over a number of days.

It the aim of this to include this in bargraphs? or/and a feed value widget? I guess that gets a bit more complicated implementation wise. We need to add an extra bit of code to make that last day calculation.

mattjgalloway:
> Hello Matt, sorry for lack of engagement on this.

No worries at all! I'm sure you're busy!

> My preference would be for a separate endpoint, as the request is feed type specific to accumulating kwh/wh.

OK, that makes sense. So it currently is a separate endpoint in this PR. But it also returns all of what a feed is. It sort of smells a bit wrong. It just made life a lot easier when I was making the dashboard - as all I had to do was change that from fetching `list.json` to `listdmy.json` (the endpoint I add here).

> For the apps such as myelectric and mysolar the last day is calculated via a last value feed list request and a specific request to the start of the day or month - or just the result of a daily data request over a number of days.

I could have done a similar thing. But I don't think there's currently a way to ask for data at a specific time is there? You can do a `data.json` with a `start` and `end`, and then I guess set the `interval` as `end` - `start` and that'd give you the value at the `start` time I think?

> It the aim of this to include this in bargraphs? or/and a feed value widget?

It was driven from me wanting a feed value widget on my dashboard which shows the consumption/import/solar/etc /today/.

mattjgalloway:
Some more discussion on this is over at https://community.openenergymonitor.org/t/incrementing-feeds-daily-monthly-yearly-values/1491/7 where I put a new proposal up.

TrystanLea:
Its possible to use data.json to return a value at a time with: 
https://emoncms.org/feed/average.json?id=0&start=UNIXTIME_MILLISECONDS&end=(UNIXTIME_MILLISECONDS+1000)&interval=1

Then pull out the first datapoint from the array. It would be nice to wrap this in its own endpoint if youd be interested to do that. To save having to know to set the interval to 1 and end at 1000 milliseconds ahead..

mattjgalloway:
Yeh I just discovered that little trick last night. That's neat and would certainly help being wrapped.

Then all that needs adding is the automatic "start of today" time stamp calculation. Although maybe that is worth keeping on the client. Hmm.

TrystanLea:
Closing this as chosen solution is to a have a wrapped separate endpoint. If anyone would like to help by adding in such an endpoint I would happily merge the pull request. Thanks Trystan

PR #556, opened 2016-09-10
Add iOS full-screen app capability to embed=1 pages

mattjgalloway:
This ensures that if you have `embed=1`, to have a page with no chrome, that you can use this successfully with "Add to home screen" on iOS.

Without this, when you open the "app" that gets added to the home screen, it just opens Safari rather than showing the page in full screen without Safari's chrome.

I believe the intended behaviour is to show it full screen. That way you can use `embed=1` and "Add to home screen" to make a dashboard that shows literally just the dashboard.

### DIFF
--- a/Theme/basic/embed.php
+++ b/Theme/basic/embed.php
@@ -9,13 +9,17 @@
   Part of the OpenEnergyMonitor project:
   http://openenergymonitor.org
   */
-  global $ltime,$path,$fullwidth,$emoncms_version;
+  global $ltime,$path,$fullwidth,$emoncms_version,$theme;
 ?>
 <html>
     <head>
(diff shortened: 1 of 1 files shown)


Paul-Reed:
Matt, I see that most of these additions are already included in [Theme/basic/theme.php](https://github.com/emoncms/emoncms/blob/master/Theme/basic/theme.php#L20-L23). 
Is it necessary to duplicate them in embed.php?
Paul

mattjgalloway:
Yeh. `theme.php` gets used for non-embed, and `embed.php` gets used for embed. So for the iOS home screen stuff for embed pages to work, you need to have this in `embed.php` as well.

Paul-Reed:
Thanks for the explanation Matt.

Pau

PR #567, opened 2016-10-03
Patch to enable 'multiple' MQTT base topics

beaylott:
This is a small patch to enable the base topic for MQTT to have sub-topics just in case that is considered useful. This is of no use in standard configuration, only where data is being received from external MQTT broker with more complex topics. 
This has NOT been tested with standard emonpi + emonhub setup, but is working for achieving above in custom setup. I could do testing in near future. 
Some additional warning/error messages have been added to log as well to help debug. 
Other extensions could be to allow multiple bast topics.


glynhudson:
Tested, works fine on emonPi using single `emon` base topic. Nice work, thanks :+1:

PR #571, opened 2016-10-09
Make dropdowns a maximum height and allow them to scroll

mattjgalloway:
This aids usability when you have a long list of items. Moving your mouse all the way down can be tedious. Scrolling the list is easier.

In particular this is useful when you have a long list of dashboards which ends up going off the screen. Now it'll scroll so you can get to all of the items. This was picked up here:
https://community.openenergymonitor.org/t/dashboard-menu-does-not-scroll-not-all-dashboards-accessible-from-it

### DIFF
--- a/Theme/basic/emon.css
+++ b/Theme/basic/emon.css
@@ -93,3 +93,9 @@ input[type=text][class=variable-name-edit] {
 .menu-dashboard,.menu-left,.menu-extra,.menu-setup,.menu-right {
   display: inherit !important;
 }
+
+.scrollable-menu {
+  height: auto;
+  max-height: 250px;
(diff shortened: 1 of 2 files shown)


Paul-Reed:
Matt looking through your code, I'm assuming that these change will equally apply to the other dropdown menus, such as 'Apps' or 'Extras' (should their number of items increase!!), so there will be some future consistency on the menu bar.

Paul

mattjgalloway:
That's right.

Paul-Reed:
Thanks Matt, makes life easier especially when trying to access a long list on a small screen. I've also spoken with Willems (joyrider3774) who also finds the fix of benefit.

Paul

PR #575, opened 2016-10-16
Update input api with http post section

joyrider3774:
As i found out in the sources that the input api accepted http post parameters using the http post "data" variable i updated the input api documentation and added examples for http post using curl.

I also updated the json http get & http post example to include multiple values as it is not clear with only one data value that you can add multiple values at once. This is however clear if you provide a sample with multiple inputs as you can then deduct from multiple values that one values is possible also. However deducting from one value that multiple are possible is not so clear. Hence the change.

I only tested the curl examples in windows, they should work in linux also but i'm not certain of the double quote syntax in linux, perhaps one of you using linux can double check them ?

edit: don't worry about the screenshots showing an api key as this is just a seperate dev install i'm using so it's okay this is public and it's only showing my local url

screenshots:
Json example with multiple values

http post section

### DIFF
--- a/Modules/input/Views/input_api.php
+++ b/Modules/input/Views/input_api.php
@@ -23,9 +23,9 @@
 <h3><?php echo _('Available JSON commands'); ?></h3>
 <p><?php echo _('To use the json api the request url needs to include <b>.json</b>'); ?></p>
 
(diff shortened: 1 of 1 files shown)


glynhudson:
thanks! 👍

PR #576, opened 2016-10-19
Added "saveAsImage" functionality to Visualisations

RDCH106:
Added saveAsImage functionality to:

- timestoredaily
- editrealtime
- editdaily
- orderbars
- histgraph
- timecompare
- graph
- rawdata
- zoom
- stackedsolar
- stacked
- simplezoom
- realtime
- threshold
- orderthreshold
- dailyhistogram
- compare
- bargraph
- multigraph

_Note: The functionality is available clicking with the secondary mouse button_

Flot Plugins
http://www.flotcharts.org/plugins/

Plugin saveAsImage origin:
https://github.com/Jeff-Tian/jquery.flot.saveAsImage.js

### DIFF
--- a/Lib/flot/plugin/saveAsImage/README.md
+++ b/Lib/flot/plugin/saveAsImage/README.md
@@ -1,0 +1,51 @@
+canvasAsImage.js
+================
+
+Reference this file in your header of which page you are using canvas, then when you right click on your canvas, a context menu would appeared and gave you a option to save your drawings on canvas as image to your local disk.
+
+##Online examples:##
+http://zizhujy.com/GraphWorld
+
+jquery.flot.saveAsImage.js
+==========================
+
+Flot plugin that adds a function to allow user save the current graph as an image by right clicking on the graph and then choose "Save image as ..." to local disk.
+
+Copyright (c) 2013 http://zizhujy.com.
+Licensed under the MIT license.
+
+##Screen shot:##
+http://zizhujy.com/blog/post/2013/07/02/A-Flot-plugin-for-saving-canvas-image-to-local-disk.aspx
+
+##Installation and usage:##
+```html
+bower install flot.saveasimage
+
+<script type="text/javascript" src="bower_components/flot.saveasimage/lib/base64.js">
+<script type="text/javascript" src="bower_components/flot.saveasimage/lib/canvas2image.js">
+<script type="text/javascript" src="bower_components/flot.saveasimage/jquery.flot.saveAsImage.js">
+```
+    
+Now you are all set. Right click on your flot canvas, you will see the "Save image as ..." option.
+
+##Online examples:##
+http://zizhujy.com/FunctionGrapher is using it, you can try right clicking on the function graphs and
+you will see you can save the image to local disk.
+
+##Dependencies:##
+
+This plugin references the base64.js and canvas2image.js.
+
+##Customizations:##
+
+The default behavior of this plugin is dynamically creating an image from the flot canvas, and then puts the 
+image above the flot canvas. If you want to add some css effects on to the dynamically created image, you can
+apply whatever css styles on to it, only remember to make sure the css class name is set correspondingly by 
+the options object of this plugin. You can also customize the image format through this options object:
+
+```javascript
+options: {
+    imageClassName: "canvas-image",
+    imageFormat: "png"
+}
+```
--- a/Lib/flot/plugin/saveAsImage/jquery.flot.saveAsImage.js
+++ b/Lib/flot/plugin/saveAsImage/jquery.flot.saveAsImage.js
@@ -1,0 +1,169 @@
+/* Flot plugin that adds a function to allow user save the current graph as an image
+    by right clicking on the graph and then choose "Save image as ..." to local disk.
+
+Copyright (c) 2013 http://zizhujy.com.
+Licensed under the MIT license.
+
+Usage:
+    Inside the <head></head> area of your html page, add the following lines:
+    
+    <script type="text/javascript" src="http://zizhujy.com/Scripts/base64.js"></script>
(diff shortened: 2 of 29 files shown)


glynhudson:
Hi Ruben, thanks for taking the effort the submit this PR. 

I like the feature and think it's useful. I often take emoncms screenshots to illustrate a particular graph etc. 

However, the functionaility does need to be balnced with potential 'bloat' caused by this flot plugin. Measuring the effect of the changes introduced with the PR it seems it's not overly significant: an extra 600ms and 1KB of data (this was measured using Chrome 53 and an emonPi RasPi3 connected via Wifi on a local network). 

screenshot 2016-10-20 at 00 26 57

One issue I have notived is that the save feature does not work on the graph module. I guess this is because the relevant changes will need to be mirroed in: https://github.com/emoncms/graph

Accepting this PR or not will come down to: *does this feature add extra functionality that cannot be achevied via a screen grab tool*? 

IMO I beleive it does. What do you think @emoncms/developers ?

Paul-Reed:
It only adds value if users find it useful, and I personally would not use it as I have other browser tools that enable me to capture partial/full screenshots as well as annotate, resize, even save as an animated gif.

Although probably difficult to implement, I would rather see a feature like this being an optional plug-in, which users could add if they considered it of value to them personally.

Paul

RDCH106:
First of all, all the graphs have the function "save as" by default, If you use it, the downloaded graph will be black and It does not show the graph. Because of that, I decide to add functionalitiy because the option already exists before my "saveAsImage" functinality addition.

The extra downloaded data generated by the functionality is not relevant beacasue all the browsers make cache of the data and in the next request only the new data is downloaded. I think that it is not problem. It is true that it would be better if the functionality be optional as says @Paul-Reed. I can study the option and implement it in the config file of EmonCMS (for example).

@glynhudson, I  have applied the functionality generally and I don´t know the existence of the [graph repository](https://github.com/emoncms/graph). I can mirroring the changes to the repository.

@glynhudson, I agree with you in the idea of a button / icon for the graph download, but for each graph, not only for a panel as in the image shared. Can you give me the reference of the topic in the Discourse forum?

Other thing that I would say is that the best way to add other libraries or projects to EmonCMS, is adding them as submodules referenced to a concrete commit. Usually all the libraries are copied to the EmonCMS project, if the libraries are updated, the libraries must copied again to update them. Check submodule command for Git:

https://git-scm.com/docs/git-submodule

(This is a suggestion as developer.)

To finish I will say that **it is true that it is extra functionality that it can be achieved via a screen grab tool, but I think that it is worst to have an option that it does not work and it is proposed to the user.**

pb66:
If the existing function is present and broken I think this could be worth including even if it's not a wishlist priority.

It could be very handy for those users that want to post something to the forum and do not have a 3rd party utility, to be able to tell someone how to do an easy screenshot (not dependent on whet toools they might have or browser or OS they are using) might prove a useful support tool, but we have got by ok with out it so far, but I see no good reason not to include the functionality.

The "open in another tab" is quite useful as it gives you a "freeze/pause" or "copy" without a screen capture being saved or opening a second "live" browser tab. 

I did wonder if the "Copy image address" could be used to post directly to the forum without saving a local copy, using the "Add image or file / From the web" option, I would find that useful, faster and it would prevent cluttering by hdd, but I think it needs some work still. An example url is 

```
data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAABHQAAANICAYAAACmE7juAAAgAElEQVR4XuydB5wUZfL3n11gyQgooigKekbMiTMi5pz1Vc9wwb933umhAkpY2 (EDITED SEE BELOW)	
		
```
It's the same url you see in the browser bar after using "open in another tab" but it doesn't work if you "copy image url" and then try and paste that into the forum dialog box, in fact it freezes and doesn't seem to recover (well not soon anyway, I had to shut the browser down).

Slightly related, it would be handy if we could include print/copy "process list", "feed list" and/or "input list" at the click of a button, preferably in a friendly text form rather than a screen dump.		
		
(EDIT) I had to remove most of the "copy image address" url due to this message



It's also now pretty self explanatory why the forum page become unresponsive when I tried to use the copied url

glynhudson:
I will merge this PR, as @pb66  mentioned since the save-as functionality is already present in emoncms it's best that it works. I personally think this is a useful feature, I look forward to using it. However, I do think that most users won't noticed it, e.g I use Emoncms daily and had not noticed until @RDCH106 mentioned that the right-click-save-as option existed! This is why I think a little save icon on the bottom of each graph would be useful. Every image on our Discourse forum has a download button when the image is clicked e.g. https://community.openenergymonitor.org/t/floating-palette-for-dashboard/1755/41?u=glyn.hudson

@RDCH106 for consistency, please could make a PR to the graph module to enable save-as functionality. Yes, I am aware of git-submodules. [I have used them in the past](https://blog.openenergymonitor.org/2016/04/Git-Sub-Modules/). However, I wasn't particularly impressed. They seemed like a bit of a git afterthought and not particularly well implemented. I know I am not alone with my reservations about sub-modules. Do you know of any other tools what other that do the same job? Have you any experience with PHP Composer? Does it work in a similar with to Ruby bundler?  Could Composer be used to specify a 3rd party library version dependency?

RDCH106:
@glynhudson As I have understood, Composer works with PHP libraries dependencies. It is necessary use a JS dependencies manager as [Bower](https://bower.io/) and connect it with Composer. Git submodules it is more generic, because Git helps to manage all the project, the implementation language it does not matter too manage the dependencies.

When a submodule is added to a project, it is possible manage the submodule without git submodule. Submodule is a high level bulk manager and manage more than one submodule with one command. You can also go to the submodule root and work with it as a standalone project. This is the real value of a Git submodule concept, very easy and powerfull.

About PR to the graph module to enable save-as functionality, I will do it as soon as possible.

Thanks for the merge!!!

PR #577, opened 2016-10-20
display "real" memory instead of Linux memor

glynhudson:
Change Administration page to display "real" memory instead of Linux memory.

Source: http://www.linuxatemyram.com


chaveiro:
This may be misleading if you're used to linux terminology.
Why not show two more lines for buffers and cached ram?

MemTotal — Total amount of physical RAM, in kilobytes.
MemFree — The amount of physical RAM, in kilobytes, left unused by the system.
Buffers — The amount of physical RAM, in kilobytes, used for file buffers.
Cached — The amount of physical RAM, in kilobytes, used as cache memory.

JonMurphy:
I had started out with this but thought it was too confusing:
screen shot 2016-10-20 at 8 14 05 pm

PR #581, opened 2016-10-24
Remove reliance on linux / unix tail command.

AndyTaylorTweet:
Replaced the reliance on 'tail' in the admin module.
Source PHP from here: http://tekkie.flashbit.net/php/tail-functionality-in-php

This should provide a like for like drop in for tail so that Windows server users can still get the log file displayed in the admin module.

### DIFF
--- a/Modules/admin/admin_controller.php
+++ b/Modules/admin/admin_controller.php
@@ -59,7 +59,45 @@ function admin_controller()
                 $route->format = "text";
                 if ($log_enabled) {
                     ob_start();
-                    passthru("tail -25 " . $log_filename);
+                      // PHP replacement for tail starts here
+                      // full path to text file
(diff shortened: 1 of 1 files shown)


AndyTaylorTweet:
This PR relates to this;
https://community.openenergymonitor.org/t/for-windows-users/1958

Paul-Reed:
Thanks Andy - a nice update which Windows users will appreciate.
Works fine on Linux, and once merged hopefully will be tested by a Windows user.
Hate to ask Willems @joyrider3774, but...

Paul

joyrider3774:
@Paul-Reed works perfect on windows
@AndyTaylorTweet thanks for the modifcation modifcation it removes the need for gnu utils forks well at least there, don't know if there are other places that uses these utils (i know cat is used but only on the emonpi section so that's okay).

PR #582, opened 2016-10-24
Add emonPi Release

AndyTaylorTweet:
Requested by Glyn here: https://community.openenergymonitor.org/t/admin-display-currently-installed-emoncms-modules/1954/11
This should only display on Raspberry Pi hardware, and ONLY when using the emonPi SD Card image.


glynhudson:
Wow, thanks, Andy. That was quick 👍 

Works perfect for me. I see the code for this is in the RasPi section so will not affect any other non-Pi installations. I tested removing the emonSD file and noticed the 'release' section disappeared. Fantastic, thanks a lot 

screenshot-emonpi 2016-10-25 00-23-25

PR #584, opened 2016-10-25
Option to save log files with a click on the button in the admin page

joyrider3774:
As per request of @glynhudson  i added a quick option to download the log file. I intentionally did not put text as content type so you always get a download box (depending on the browser) to choose where to save the file and don't view the file inside the browser.

I tested it on my windows and it seems to work fine for me should be tested on linux

### DIFF
--- a/Modules/admin/admin_controller.php
+++ b/Modules/admin/admin_controller.php
@@ -159,7 +159,9 @@ function admin_controller()
                     flush();
                     if (file_exists("/home/pi/data/emonpiupdate.log"))
                     {
-                      trim(readfile("/home/pi/data/emonpiupdate.log"));
+                      ob_start();
+                      readfile("/home/pi/data/emonpiupdate.log");
+                      echo(trim(ob_get_clean()));
                     }
                     else
                     {
@@ -193,7 +195,9 @@ function admin_controller()
                     header("Expires: 0");
                     flush();
                     if (file_exists("/home/pi/data/emonpibackup.log")) {
-                      trim(readfile("/home/pi/data/emonpibackup.log"));      
+                      ob_start();
+                      readfile("/home/pi/data/emonpibackup.log");
+                      echo(trim(ob_get_clean()));
                     }
                     else
                     {
--- a/Modules/admin/admin_controller.php
+++ b/Modules/admin/admin_controller.php
@@ -63,7 +63,13 @@ function admin_controller()
                 header("Pragma: no-cache"); 
                 header("Expires: 0");
                 flush();
-                readfile($log_filename);
+                if (file_exists($log_filename)) {
+                  readfile($log_filename);
+                }
+                else
+                {
+                  echo($log_filename . " does not exist!");
+                }
                 exit;
               }
             }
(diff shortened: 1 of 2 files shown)


joyrider3774:
I added 2 extra options to download the emonpi update log and backup log for the emonpi update log i added a button but for backupload i did not (but did add the route) as i could not find where it is being used.

Please test this on an actual emonpi as i don't have an emonpi i blindly programmed this !

joyrider3774:
for the emonpi related logfiles i added $session['admin'] verifications as i saw the normal getlog routes did this as well. I also added a fileexists check when going the download route and when the file does not exist it still downloads a file with a message in it saying the logfile did not exist. This is to prevent the browser from navigating away when no file would be downloaded or displaying blank pages. (It makes sure a file will always be downloaded even if the log does not exist) The normal logfile does not have this problem as a file is always created if we have write access and if we don't have write access no button is being shown but the emonpi buttons should always be shown as you trigger the update itselve without reloading the page (ajax call)

glynhudson:
Yah! The logs download great :+1: 

However, when I open the emonPi log files there seems to be some invalid characters at the start

selection_103

I don't see these characters when I view the log directly on the Pi:

selection_104

JonMurphy:
glyn - those have always been there in the emonpiupdate.log file

screen shot 2016-10-25 at 12 52 00 pm

glynhudson:
Ah, interesting. So it's no fault of the download. 

I wonder where those characters are coming from? I can't see them in the emoncms logview or when I view the log directly 

```
head ~/data/emonpiupdate.log 
Starting emonPi Update >
EUID: 1000

Mon 24 Oct 14:55:32 UTC 2016

git pull /home/pi/emonpi
Updating 10e2b76..c3bee09
Fast-forward
 .gitignore                                      |  27 +++++
 .travis.yml                                     |  42 ++++++++
```

Mmmm..I'm not sure where to look. Could it be the log file format?

joyrider3774:
i could try adding ob_clean back in it should clear the buffer before downloading the file... but that would only work if something already had written data to the buffer somehow (like extra whitespaces after ?> etc) it even seems they are 0 characters so not sure if they would come from extra whitespaces in the php files. I'll try the ob_clean thing. I'll change it immedialty and do the pr this way you can test if that's the problem

joyrider3774:
the zero's might also come from something that "service-runner-update.sh" does but i don't have that file and can't find in the repositories as all the update function does is run that script and pipe the output to a file

joyrider3774:
the reason the log on the admin page does not have those zero characters was because trim got called on the result. I'm explicitly doing the same thing now. But i don't know what kind of side effect this might have if you got a logfile of say 500 mb or 1 gigabyte etc as your passing that amount of data to the trim function then.

Calling trim should fix the zero characters when downloading the file as well though

joyrider3774:
don't merge / try yet i made a mistake with the trim statements

joyrider3774:
it should work now. I only trim emonpi update & backup log but not emoncms log. The emonpi logs probably stay very small while the emoncms log can be become large and did not have the problem

joyrider3774:
i tested with a raspberry pi and emonSD-03May16 image and as far as i can tell it seems to work. Perhaps someone else can do a final test before actually merging

Modules/admin/admin_controller.php
```diff
@@ -122,13 +142,34 @@ function read_file($file, $lines) {
...
                 
+                if ($route->subaction == 'downloadupdatelog' && $session['admin'])
+                {
+                    header("Content-Type: application/octet-stream");
+                    header("Content-Transfer-Encoding: Binary"); 
+                    header("Content-disposition: attachment; filename=emonpiupdate.log");
+                    header("Pragma: no-cache"); 
+                    header("Expires: 0");
+                    flush();
+                    if (file_exists("/home/pi/data/emonpiupdate.log"))
```
glynhudson: the path for `emonpiupdate.log` is defined on line 133

https://github.com/joyrider3774/emoncms/blob/ebe7dc607da65ba2fcb5233bc12d8877f2176656/Modules/admin/admin_controller.php#L133

It would be best to reference this variable `$update_logfile` here rather than hardcode the abs path in another location

Modules/admin/admin_controller.php
```diff
@@ -122,13 +142,34 @@ function read_file($file, $lines) {
...
+                    {
+                      ob_start();
+                      readfile("/home/pi/data/emonpiupdate.log");
```
glynhudson: same here

Modules/admin/admin_controller.php
```diff
@@ -122,13 +142,34 @@ function read_file($file, $lines) {
...
+                      echo(trim(ob_get_clean()));
+                    }
+                    else
+                    {
+                      echo("/home/pi/data/emonpiupdate.log does not exist!");
```
glynhudson: and here

glynhudson:
emoncms.log can only be a max of 1Mb (on an emonPi) due to an aggressive log rotate since we are keeping the log on a tmpfs ramdisk. I don't think the log will be too large for trim to handle. The emonpi update log also won't be very large at all few Kb max probably. 

https://github.com/openenergymonitor/emonpi/blob/master/logrotate.conf#L12

Working great, just made a couple of minor comments with regard to variable referancing.

Modules/admin/admin_controller.php
joyrider3774: that would not work as far as i can tell the variables are only known on that route (in that if statement). I can move them outside the if statement and reference adapt my code as well as the other code that was there before to use the variable. The only not so clean down side is that the variables would be set each time you use the emonpi route action while perhaps never having need for them although i don't really consider that a bad thing as it would not have an impact and then you can reference the variables

the code actually used harcoded absolute paths before i made my changes ..

glynhudson:
Perfect, thanks. Merged 👍

PR #587, opened 2016-10-27
Re-enable auto scrolling of update log view...

AndyTaylorTweet:
Re-worked my original commit for this from a couple of days ago.
It now scrolls the log to the end, once the end "emonPi update done" is reached, control of scrolling is relinquished allowing the user to scroll back up the log.

The one caveat to this - is that if the update process bails before completion, the user cannot scroll up - however that is somewhat of a mute point given the new option to download the log for diagnosis / debugging / support.


AndyTaylorTweet:
Relates to this thread:

https://community.openenergymonitor.org/t/wouldnt-it-be-nice-if-the-log-windows-on-the-admin-page-would-scroll-before-your-eyes/1986/10

glynhudson:
thanks! 👍

PR #590, opened 2016-10-30
Modify the "emoncms Modules" line...

AndyTaylorTweet:
Modify the behaviour to list ALL module folders (regardless of presence of a .git file);
Modify the behaviour such that "Stock" modules present in the vanilla version of emonCMS are plain text, and any other modules are BOLD tagged.

Continuation of the work here: https://community.openenergymonitor.org/t/admin-display-currently-installed-emoncms-modules/1954/18

### DIFF
--- a/Modules/admin/admin_main_view.php
+++ b/Modules/admin/admin_main_view.php
@@ -124,16 +124,16 @@
       return $partitions;
   }
 
-$emoncms_modules = NULL;
-$emoncmsModulesPath = substr($_SERVER['SCRIPT_FILENAME'], 0, strrpos($_SERVER['SCRIPT_FILENAME'], '/')).'/Modules';
-$emoncmsModuleFolders = glob("$emoncmsModulesPath/*/.git", GLOB_BRACE);                                                 // find all the subfolders containing a .git file
+$emoncms_modules = NULL;                                                                                                // Set the variable to null
+$emoncmsModulesPath = substr($_SERVER['SCRIPT_FILENAME'], 0, strrpos($_SERVER['SCRIPT_FILENAME'], '/')).'/Modules';     // Set the Modules path
+$emoncmsModuleFolders = glob("$emoncmsModulesPath/*", GLOB_ONLYDIR);                                                    // Use glob to get all the folder names only
         foreach($emoncmsModuleFolders as $emoncmsModuleFolder) {                                                        // loop through the folders
         $emoncmsModuleFolder = str_replace($emoncmsModulesPath."/", '', $emoncmsModuleFolder);                          // clean up the formatting, removing the path from the $emoncmsModulesPath variable
-        $emoncmsModuleFolder = str_replace("/.git", '', $emoncmsModuleFolder);                                          // more clean up to remove /.git from the end of the output
-        if ($emoncmsModuleFolder == 'app' || $emoncmsModuleFolder == 'config' || $emoncmsModuleFolder == 'dashboard' || $emoncmsModuleFolder == 'graph' || $emoncmsModuleFolder == 'wifi') {$emoncmsModuleFolder = "<b>".$emoncmsModuleFolder."</b>";}
-        if (!is_null($emoncms_modules)) { $emoncms_modules = $emoncms_modules.", ".$emoncmsModuleFolder; } else {$emoncms_modules = $emoncmsModuleFolder;} // add the commas as appropriate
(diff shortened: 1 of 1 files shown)


chaveiro:
There are logic to detect modules already on emoncms. 
You are hard coding known modules names here.

AndyTaylorTweet:
There is ?

chaveiro:
There are 4 supported modules type files that emoncms looks for:
All directories bellow /Modules like /Modules/mod_name are assumed to be a module and a controler file is accepted to  be accesed via an url like http://emonhost/mod_name.

1. Controler modules:
"Modules/".$controller_name."/".$controller.".php"
See core.php

2. Menu entries:
"Modules/".$mod_name."/".$mod_name."_menu.php"
See core.php

3. Process list modules:
"/Modules/".$module_name."/".$module_name."_processlist.php"
See: Modules/process/process_model.php

4. Widgets from the dashboard module.
/Modules/[module_name]/widget/[widget_name]/[widget_name].js or php files

See Modules/Dashboards/Views/loadwidgets.php


This information should be available at a forum post. I will do that.

pb66:
Thank you @AndyTaylorTweet.

chaveiro:
This PR is in discussion. Pb66 merge was precipitated. 
Hope you can review this code later to remove the hard coded modules name.

AndyTaylorTweet:
I hope you will have read the code and can see that the hard coding ONLY differentiates between modules considered as *standard* and those concidered an addition.

should a new standard module be included and this code not get updated, the only side effect is that the standard module would be bolded in the list.

chaveiro:
A module is a module, core or not. So why differentiate this here?

AndyTaylorTweet:
because it was asked for to aid support.

Paul-Reed:
I don't see why they should be differentiated either. Looking at the forum,
Glyn shared the same view...

Paul

chaveiro:
My recommendation: keep it simple. 
Avoid this kind of exceptions in code that need to be maintained.

AndyTaylorTweet:
I don't disagree, I wrote the code in answer to the original request - if the cencencus is that it should not differentiate between stock modules and those that are user added - thats simple enough to pull out.

pb66:
This was a simple amendment, the original PR was swiftly merged without discussion and only displayed some modules and those in bold related to the emonPi so were a bit misleading, this merge is an improvement (minor correction?) regardless of your position on whether the now correctly identified "additional" modules should be bold or not.

Glyn did comment about not identifying emonSD, emonPi or what were or weren't "considered" core or essential, my reply to his comment confirmed the only differentiation made in this PR was "in-built" or not, which he "liked" that post and I merged the PR as the existing was confusing.

The reason behind making the modules that are not "in-built" bold is to highlight the differences or customization of the emoncms instance at a glance to aid support and also to help users understand the difference between "in-built" and "additional", it is clear from the forum thread that this can be a confusing point even for the more experienced users and devs alike. 

The "inclusion" of "additional" modules on the emonSD card further confuses things and for users who are using the emonSD to be able to see they have some "additional" modules already is very useful.

Since all in-built modules must also be listed in the [.gitignore](https://github.com/emoncms/emoncms/blob/master/.gitignore#L10) documentation or a comment added to the .gitignore, or even an emoncms/Modules/dev-notes.md could remind devs to make the entries in .gitignore and Modules/admin/admin_main_view.php. I had also already considered using the .gitignore file for this function but in the rare case a user uses the download rather than a clone it would break.

I agree the code could be made more compact if a existing "in-built" list was available but to my knowledge there isn't, I assume from the fact Andy has done it this way it is more efficient or practical than using some form of "if module is in an array" rather than the string of "if", "ands". I'm not up on PHP enough to comment.

If the bold offends you that much remove it, but in it's current form it is very informative for both user and supporter, the bold gives clear insight to what is not mandatory ie "in-built" or optional ie "additional" even if "included" on a pre-built image. 

But one (granted rather long) line of code is hardly going to impact the size or complexity of emoncms and as Andy has already stated, the worse that can happen is that bold is incorrectly applied (or not) if not kept updated alongside the .gitignore.

glynhudson:
I vote for keeping it simple and just listing the installed modules. The definition of 'core' / 'essential' modules may change between installations and will almost certainly change in the future. Sorry for any confusion, very much appreciate your work on this @AndyTaylorTweet

AndyTaylorTweet:
I'll get a new PR done.

PR #598, opened 2016-11-11
Additional checking for memory information

AndyTaylorTweet:
Some Linux hosts do not allow access to '/proc/meminfo' in the case of some VPS services.
Actually check to make sure /proc/meminfo is available (remove error messages from those cases where its not available.)


Paul-Reed:
Thanks Andy

Paul

PR #601, opened 2016-11-13
Add iOS App Store link

mattjgalloway:
This adds a link on the profile page to the iOS App Store for the iOS app.


glynhudson:
Thanks!

PR #602, opened 2016-11-13
Enable environment-variable settings in default.settings.php

spikeheap:
A belated follow-up to https://github.com/emoncms/emoncms/pull/547. 

This change adds the ability to define some core settings as environment variables. When set, environment variables take precedence over `settings.php` and defaults. This enables Docker (and other services) to provision emoncms without needing to alter the default settings file at all, and (for example) allows the same codebase to be run against different environments for testing.

This doesn't affect the behaviour on systems where the environment variables are not set.

### DIFF
--- a/process_settings.php
+++ b/process_settings.php
@@ -122,7 +122,7 @@ if (isset($_ENV["MYSQL_PORT"]))     $port = $_ENV["MYSQL_PORT"];
 // create the array if it's not already been done
 if (!isset($redis_server)) $redis_server = array();
 
-if (isset($_ENV["REDIS_ENABLED"]))  $redis_enabled = $_ENV["REDIS_ENABLED"] === 'true' : false;
+if (isset($_ENV["REDIS_ENABLED"]))  $redis_enabled = $_ENV["REDIS_ENABLED"] === 'true';
 if (isset($_ENV["REDIS_HOST"]))     $redis_server['host'] = $_ENV["REDIS_HOST"];
 if (isset($_ENV["REDIS_PORT"]))     $redis_server['port'] = $_ENV["REDIS_PORT"];
 if (isset($_ENV["REDIS_AUTH"]))     $redis_server['auth'] = $_ENV["REDIS_AUTH"];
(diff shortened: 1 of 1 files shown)


glynhudson:
Thanks! Just tested and can confirm is does not affect existing setups when env variables are not set

chaveiro:
Hum this seems a bit hacked.. Why not use instead one env variable to select the config file name/path. That has all the settings for that env?

spikeheap:
Hi @chaviero, thanks for taking a look at my PR. Actually, using the environment variable to point to particular config file only solves half the problem this PR addresses. The problem with static config files is that it's all-or-nothing and ends up quite brittle. For example, suppose I want to test the app with an identical setup _except_ for the database – I would need to clone `settings.php` in addition to using an environment variable, and I'd need to manually keep them in sync. With the environment-variable approach you can just override the things you'd like to and it's nice and explicit.

Manually setting environment variables in Apache configs can be a bit tedious, but fortunately the community has developed [phpdotenv](https://github.com/vlucas/phpdotenv) which allow you to store your environment variables in a `.env` file, which is a similar approach to how you're currently doing it with `settings.php`, albeit with a stronger separation of config and code.

Using environment variables for configuration is a pretty well established best-practice in software development. I think (though I'm probably misattributing here) that it was first formalised by Heroku as part of the [12-factor app](https://12factor.net/config) movement. Without environment variable support it would be quite onerous to deploy emoncms to a PaaS like Heroku or Elastic Beanstalk, or to use tools like Docker in tandem with more standard approaches.

Does that addresses your primary concern about it being a hacky approach? 

In terms of the layout, I am a little uncomfortable with the effect it's had on `process_settings.php` (it's longer and harder to read), and I'm considering an optimisation along the lines of:

```php
function validateSettings() {
  //validation code
  return $error_out;
}

function loadSettingsFromEnvironment(){
  // the new environment variable loading code here
}

if(file_exists(dirname(__FILE__)."/settings.php"))
{
    require_once('settings.php');
}
else
{
    echo "<div style='width:600px; background-color:#eee; padding:20px; font-family:arial;'>";
    echo "<h3>settings.php file error</h3>";
    echo 'Copy and modify default.settings.php to settings.php<br>';
    echo 'For more information about configure settings.php file go to <a href="http://emoncms.org">http://emoncms.org</a>';
    echo "</div>";
    die;
}

loadSettingsFromEnvironment();
validateSettings();
```
which would then let us separate out some of the concerns in the code block. I'm not entirely sure how to split this out into files sympathetically though. Any suggestions?

chaveiro:
My recommendation to implement this feature : clone settings.php on a folder outside version control for each dev environment like settings.dev.php or settings.test.php and set ONE env variable to point emoncms to use that file accordingly to your environment.

Pros: 
- Code is cleaner.
- Does not slow down emoncms with IF's, only one is used.
- Supports every config usecase.
- It's future proof.

PR #603, opened 2016-11-13
Validate variables after environment loaded

spikeheap:
This PR improves the recently merged environment variable support (https://github.com/emoncms/emoncms/pull/602) by validating the setup _after_ the environment variables have been pulled in. This has a couple of benefits:

1. In a setup where environment variables are used, the config no longer needs to be duplicated in `settings.php`, but omitting the config from _both_ yields the correct error.
2. Any validation added in future will be applied to values loaded from environment variables, making the behaviour consistent across config approaches.

### DIFF
--- a/process_settings.php
+++ b/process_settings.php
@@ -17,9 +17,50 @@ require_once('Lib/enum.php');
 // Check if settings.php file exists
 if(file_exists(dirname(__FILE__)."/settings.php"))
 {
-    // Load settings.php
+    /*
+        Load settings.php
+    */
     require_once('settings.php');
 
+    /*
+        Load settings from environment variables
+
+        Environment settings override settings.php and defaults, 
+        and allow you to run multiple variants of the same
+        installation (e.g. for testing).
+    */
+
+    //1 #### Mysql database settings
+    if (isset($_ENV["MYSQL_HOST"]))     $server = $_ENV["MYSQL_HOST"];
+    if (isset($_ENV["MYSQL_DATABASE"])) $database = $_ENV["MYSQL_DATABASE"];
+    if (isset($_ENV["MYSQL_USER"]))     $username = $_ENV["MYSQL_USER"];
+    if (isset($_ENV["MYSQL_PASSWORD"])) $password = $_ENV["MYSQL_PASSWORD"];
+    if (isset($_ENV["MYSQL_PORT"]))     $port = $_ENV["MYSQL_PORT"];
+
+    //2 #### redis
+    // create the array if it's not already been done
+    if (!isset($redis_server)) $redis_server = array();
+
+    if (isset($_ENV["REDIS_ENABLED"]))  $redis_enabled = $_ENV["REDIS_ENABLED"] === 'true';
+    if (isset($_ENV["REDIS_HOST"]))     $redis_server['host'] = $_ENV["REDIS_HOST"];
+    if (isset($_ENV["REDIS_PORT"]))     $redis_server['port'] = $_ENV["REDIS_PORT"];
+    if (isset($_ENV["REDIS_AUTH"]))     $redis_server['auth'] = $_ENV["REDIS_AUTH"];
+    if (isset($_ENV["REDIS_PREFIX"]))   $redis_server['prefix'] = $_ENV["REDIS_PREFIX"];
+
+    //3 #### MQTT
+    // create the array if it's not already been done
+    if (!isset($mqtt_server)) $mqtt_server = array();
+    if (isset($_ENV["MQTT_ENABLED"]))  $mqtt_enabled = $_ENV["MQTT_ENABLED"] === 'true';
+
+    if (isset($_ENV["MQTT_HOST"]))     $redis_server['host'] = $_ENV["MQTT_HOST"];
+    if (isset($_ENV["MQTT_PORT"]))     $redis_server['port'] = $_ENV["MQTT_PORT"];
+    if (isset($_ENV["MQTT_USER"]))     $redis_server['user'] = $_ENV["MQTT_USER"];
+    if (isset($_ENV["MQTT_PASSWORD"]))     $redis_server['password'] = $_ENV["MQTT_PASSWORD"];
+    if (isset($_ENV["MQTT_BASETOPIC"]))     $redis_server['basetopic'] = $_ENV["MQTT_BASETOPIC"];
+
+    /*
+        Validate settings are complete
+    */
     $error_out = "";
 
     if (!isset($config_file_version) || $config_file_version < 8) $error_out .= '<p>settings.php config file has new settings for this version. Copy default.settings.php to settings.php and modify the later.</p>';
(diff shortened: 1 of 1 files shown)


glynhudson:
What do think @chaveiro? Does @spikeheap 's explanation on https://github.com/emoncms/emoncms/pull/602 make sense to you? 

I have tested this PR and can confirm that it does not effect existing setups. We would like to make it easier to run Emoncms in a docker container. @spikeheap 's PR to enable .env variables will avoide the need for a second docker specific `settings.php`

chaveiro:
Hi glyn, my recommendation to implement this feature : clone settings.php on a folder outside version control for each dev environment like settings.dev.php or settings.test.php and set ONE env variable to point emoncms to use that file accordingly to your environment.

Pros:
    Code is cleaner.
    Does not slow down emoncms with IF's, only one is used.
    Supports every config usecase.
    It's future proof.

Cons:
    None.

spikeheap:
Hi @chaveiro,

I don't really understand why you're so against the current implementation, but I'm happy to discuss the objective merits of using environment variables for configuration more if you think there are specific flaws, either in the implementation or the approach.

Just to try to address the approach of using a config file specified by environment variable:

> Code is cleaner.

Mixing configuration and code is well established as a bad practice in software development. PHP is code. The code actually ends up much __less__ clean because configuration is duplicated between multiple `settings.php`s unless a developer adds logic to abstract the common elements into a further file. Using code for configuration promotes this behaviour.

It also requires the developer to add `.gitignore` entries for each additional file they create.

> Does not slow down emoncms with IF's, only one is used.

Can you demonstrate that the handful of `if` statements added slow down emoncms by any real-world measure? 

> Supports every config usecase.

Your proposed solution actually doesn't support the precise usecase I mentioned in my previous comment. In any environment where immutable architecture is used, or you want to version your deployments, or where Docker is used, using multiple settings files is quite onerous.

> It's future proof.

As it doesn't address the current needs I have now, it's difficult to agree that it solves all future needs too. 

Just to be clear, I actually agree that it's a good idea to set the location of `settings.php` using an optional `EMONCMS_SETTINGS_FILE` (as you suggest) to allow users to change the location of their config file. However, that's a completely different use-case, and could easily be added _in addition_ to this change.

chaveiro:
Im not against the feature. That is a nice feature. Im against the implementation.
Just keep it simple.

spikeheap:
Hi @chaveiro,  I'm keen to hear how it could be refactored to be made simpler. For clarity, the feature is:

> The ability to define some core settings as environment variables. When set, environment variables take precedence over settings.php and defaults.

Sorry I couldn't see a simpler way to implement it than what has already been merged. I would certainly like to refactor it to DRY up some of the duplicated logic, but I kept it consistent with the rest of `process_settings.php` for an initial implementation.

Trying to move this forward, this PR just moves the already merged environment variable logic to make sure the validations are applied consistently (as they are for values set in`settings.php`). Hopefully this change isn't contentious, and can be merged anyway @glynhudson?

chaveiro:
In process_settings.php add support to one env variable named  EMONCMS_CONFIG_FILE where you set the config file path if different than the default.

Now create a new file named settings-env-override.php with something like this:
<php require "settings.php";
 if (isset($_ENV["MYSQL_HOST"]))     $server = $_ENV["MYSQL_HOST"];
if (isset($_ENV["MYSQL_DATABASE"])) $database = $_ENV["MYSQL_DATABASE"];
// any other override you want
?>

Now when you want to use the env vars  set EMONCMS_CONFIG_FILE to point to settings-env-override.php. 
It includes the settings.php and allows you to override what you want without touching emoncms core files.

spikeheap:
Hi @chaveiro,

I'm sorry, but that's a terrible solution. That would require any user that wants to use environment variables to write a custom settings file, and means adding way more logic into `settings.php`.

At this point I really don't know what to say. No meaningful comments have been raised against my changes other than that you don't like the approach, despite it being a well established practice and generally accepted by the software engineering community to superior to embedding config in code, which is what you're currently suggesting. Also, as I keep on repeating, __your proposed solution does not solve my problem__. 

PR #602 was created precisely because you told me that was the best location for this code, so it seems to be the right place. It would have been nice to find out that you objected to the principle of the changes back then.

TBH this is becoming way more effort than its worth for me. I'm quite disheartened by how dismissive your comments are, how you don't seem to have read through the comments I've made, and how you don't seem to be making an effort to understand my point of view. If it's this hard to make even minor changes (that don't even directly affect you), I really don't see me making many more contributions to this project.

This PR should be merged because it doesn't add the functionality that's under dispute, and if it's really that contentious, a new PR should be created to roll back the functionality completely.

glynhudson:
@spikeheap makes a good argument for using the environment variable methodology as proposed. 

I agree with @chaveiro that often 'keeping it 'simple' is best. However, as the setup @chaveiro described sounds to me to be *more* complicated and requires an additional file: 

> In process_settings.php add support to one env variable named EMONCMS_CONFIG_FILE where you set the config file path if different than the default.
> 
> Now create a new file named settings-env-override.php with something like this:
> if (isset($_ENV["MYSQL_HOST"])) $server = $_ENV["MYSQL_HOST"];
> if (isset($_ENV["MYSQL_DATABASE"])) $database = $_ENV["MYSQL_DATABASE"];
> // any other override you want
> ?>

I don't believe the extra 'IF' will create any significant performance degradation. If we can prove this then I will happily roll back the merge. 

Since this PR is an improvement to https://github.com/emoncms/emoncms/pull/602 , I will merge.

chaveiro:
No the user wont write any custom settings file, he can use the settings-env-override.php.
If you dont want to follow the rules and produce nice and clean code, i will fix it for you this time.

chaveiro:
See https://github.com/emoncms/emoncms/pull/606/files
set EMONCMS_SETTINGS_FILE to 'scripts/settings.env.php' to activate env settings support

spikeheap:
@chaveiro your behaviour here is completely unacceptable. The PR you just merged without anyone else's input goes directly against the argument you made here – it's way more complex, both in terms of code in this codebase __and__ effort a user has to put in to get the desired effect.

If your intention was to put me off contributing here, congratulations. I've not contributed to a more hostile project in the 11 years I've been a professional software developer.

glynhudson:
I have reverted your refactor @chaveiro . See my comment on your commit. https://github.com/emoncms/emoncms/commit/0e4bd3746aea153a4c086d7687d08fbd7a847fd4

TrystanLea:
Hello All, I'd like to suggest a 3rd option that I think makes a good combination of both approaches, we could avoid needing an additional file and avoid making all the if statements present for setups where env variables are not used.

I have created a suggestion pull request here: https://github.com/emoncms/emoncms/pull/608

What do you all think? Apologies if I have misunderstood the way its meant to work.

spikeheap:
Hi @TrystanLea, thanks for having a look at this and proposing an alternative approach.

That PR looks like a good compromise, but I'm not sure it's necessary. The reluctance to have it part of the core flow is based on the assumption that the `if` statements are expensive. I created a [script to test the performance](https://gist.github.com/spikeheap/79b0d0ffa3f49de20cf224b516c487d5), and on my laptop the full set of `if` statements add an average 0.0000006 seconds to the execution time (over 10 runs of 10,000,000 iterations). Put another way, it increases the execution time by 1 second for every 1,428,571 times the environment variables are evaluated. Am I right in thinking that's for every request? 

Is that perf improvement isn't enough to warrant the additional complexity for developers? I'd suggest that unless there's a compelling reason it's worth keeping it simple for devs – I'm imagining support requests like "I've set MYSQL_HOST, but it's not working".

My vision here is to retain environment variables as a first-class method for configuring emoncms. I've got a whole discussion issue in draft, but what I'm leaning towards longer-term is being able to __optionally__ use environment variables and _not_ have a `settings.php` – I think this can be possible while retaining the current error messaging on startup, including warning users that the `settings.php` hasn't been put in place _if_ any core settings are missing. This is definitely something to discuss elsewhere, but it's worth mentioning that the benefits of this approach are a little wider than just enabling development/testing within Docker.

Does that sound reasonable? If it's a blocker I'd definitely accept #608 over a rollback, as it does allow the behaviour I'm after.

TrystanLea:
Thankyou @spikeheap and great to see those numbers. Im happy to keep things in their present form without the ENV_ENABLE to reduce the additional complexity for developers. For interest on my laptop i get 0.4-0.6us for the env code only, 0.3-0.5us with the ENV_ENABLE check and 37us per iteration with one redis->get in the loop. But Im also happy to find another solution if this is a blocker too.

chaveiro:
Hi all, https://github.com/emoncms/emoncms/pull/608 is fine for me since this is a dev only requirement and it's nice to be able to disable it globally as it's not used on production.

spikeheap:
Hi @chaveiro, actually I intend to use this in production. It's not a dev-only requirement, though I can see how it could come across that way. Sorry if I wasn't clear before. 

As far as I can see the currently merged setup works fine, solves the required use case, and doesn't introduce any negative side-effects, so I think we can probably close this (and #608), and get on with our lives. It would be trivial to introduce the changes of #608 if we find an issue in the future, or if there are any valid criticisms of the current implementation, so there really is very little risk to continuing with the current implementation.

chaveiro:
Just a side note, since environment variables are global, it could be nice if they hint what they belong to. Eg: prefix them with EMONCMS_

spikeheap:
@chaveiro that sounds quite reasonable, that's a simple change to make too. I'll raise a PR separately (though I'm a bit short on time this weekend, so it'll likely be early next week).

Environment variables have different scopes, so you can have command-level environment variables (e.g. `RACK_ENV=review rails server` only applies to that one command), set variables for a single shell or user, or set them globally. Generally when they're used this way, you'd set them at the finest granularity you can, to prevent them bleeding out to other things. In time, I'll add some docs on this approach so I can extoll its virtues more :).

spikeheap:
I had some post-lunch downtime: https://github.com/emoncms/emoncms/pull/609.

glynhudson:
Thanks a lot.  #609 has been merged to prefix environment variables with EMONCMS_ 👍

PR #614, opened 2016-11-25
Add tests for process_settings

spikeheap:
This PR adds unit tests for `process_settings.php`. It also fixes a bug in the environment variable loading code which was caught by the tests. Yay for tests :).

To run the test suite with docker:

```bash
docker run -v $(pwd):/app --rm phpunit/phpunit -c phpunit.xml
```

Alternatively, if you've got phpunit installed:

```bash
phpunit -c phpunit.xml
```

To allow `process_settings.php` to be tested, I needed to make a couple of minor changes which don't affect the standard behaviour of the code:

- `process_settings.php` no longer calls `die`, instead setting a variable which is evaluated by the caller of `process_settings.php` (which then chooses to `die` if that's set). This was necessary because `die` can't be caught by the test suite, so invalid settings could not be tested without killing the test where it stood (silently).
- The path/filename of `settings.php` can now be overridden by an environment variable. This was necessary to enable testing, because we can't assume that `settings.php` will or won't be present, and also can't assume that it won't have been modified. By allowing the path to be customised, we can test on arbitrary files (of variants of files). This can be useful for (for example) testing legacy configuration files for the appropriate error/warning messages.
- `process_settings.php` can now `require` `settings.php` more than once. This enables testing, as without it the first load of `settings.php` is cached, so subsequent calls don't actually set the appropriate variables.

Questions, comments, etc. welcome.


spikeheap:
__TL;DR:__ The Codacy warnings aren't valid for this test file, and the Travis error can be fixed with a configuration change. 

Looking at the failing checks:

- We need to update the Travis configuration to remove the `phpunit test.php` call. @glynhudson/@TrystanLea, any chance we could update Travis? 
- When/if this PR is merged, PHPUnit should be invoked in Travis using just `phpunit` to enable these tests in CI.
- The Codacy warnings fall into two categories:
  - ` accesses the super-global variable $_ENV`. It's necessary for the test to alter environment variables because that's the behaviour it's exercising. If it's possible to disable this warning for all tests that would be prudent.
  - `Avoid unused local variables`. Because of the current structure of `process_settings.php` we need to test the local variables rather than attributes on a class. Longer-term, it should be possible to kill this error by refactoring the configuration into a class, but that requires modification of `process_settings`, which is beyond the scope of this PR.

glynhudson:
Hi @spikeheap 

Thanks a lot for the PR. This is a really good start to adding a comprehensive testing to Emoncms. 

I've just tested by running the phpunit test via docker as you suggested and the tests passed 👍 

What exactly is the change required to `.travis.yml` to run the tests, is it just a case of removing [script: phpunit test.php](https://github.com/emoncms/emoncms/blob/master/.travis.yml#L13) ? If the change to `.travis.yaml` was included in this PR then the PR should build and test successfully on Travis. Would you be able to add this in? 

Am I correct that after this change Travis won't mow check all Emoncms code for syntax errors? While [test.php](https://github.com/emoncms/emoncms/blob/master/test.php) is a simple test is has caught a number of syntax/typo like errors. It would be nice to keep this, would be possible for both tests to run together?

***

The change to how `die` is called is slightly awkward. I understand the reasoning, however, we need to be careful that *all* files that require `process_settings.php` evaluate `$failed_settings_validation`. There are a few scripts and possible some modules outside of the main emoncms repo that require `process_settings.php` e.g. there are 14 scripts in the `usefulscripts` repo that include `process_settings.php`  https://github.com/emoncms/usefulscripts/search?utf8=%E2%9C%93&q=process_settings.php. I will need to check 

We will need to be careful to ensure all the changes are made to external files before this PR is merged. Is there any other way the same result could be achieved without requiring a change to all files that call `process_settings.php`.

spikeheap:
Hey @glynhudson, thanks for taking a look.

Sorry, I should have checked for a `.travis.yml` before I created this PR. I've updated it to run just `phpunit`, and (interestingly) it fails to start the test for earlier versions of PHP. I'll dig into that.

I'm not sure I understand the intended behaviour of https://github.com/emoncms/emoncms/blob/master/test.php, as I'd thought it was just a proof-of-concept to show PHPUnit working. That test doesn't load in any additional files, so I don't think it is actually checking syntax for the rest of the codebase. It looks like https://github.com/spikeheap/emoncms/blob/feature/test_process_settings/.travis.yml#L11 does the syntax checking, so the same level of protection should be present with these changes.

I did a quick check:

```bash
echo "jhkvds;shdhh33 ? ___ {/}"  >>  Lib/enum.php
phpunit test.php
```

and the test passed fine even with the erroneous file present. 

---

On the `die`/`$failed_settings_validation` front, that's an interesting challenge. Although I've been trying to minimise the number of changes in this PR, it looks like a simple abstraction there would solve the problem once and for all. Is there a canonical list (or way to generate a list) of all the things outside`emoncms/emoncms` which load `process_settings`?

Three options spring to mind:

1. Propagate the `if($failed_settings_validation)` blocks into each of those utility scripts. This is the easiest, but results in the most duplication, and margin for error in new scripts.
2. Create a `bootstrap.php` (or similar) which does the validation and dies on error (preferably as a class). This is a little more effort, but would allow us to rationalise the common setup and preferences loading into a single place for all scripts and site. This PR wouldn't need to have all of that, so `bootstrap` would just be a thin facade over process_settings.
3. Reinstate `die` within `process_settings` but have it conditional on some global/environmental variable. This feels fairly hacky, as it introduces more global state and increases the responsibilities of the settings processor, however this option is the only one which doesn't require any changes to the files which load the process_settings file.

I'd plump for [2] because it helps push things in the 'right' direction, though it does mean a breaking change on the scripts repo to switch to requiring `bootstrap` instead of `process_settings`. Although that may feel fairly painful, it highlights the need for such an abstraction, and once we've made this change, subsequent changes should be trivial. Over time it should also have the effect of simplfying the setup in external scripts and modules.

What does everyone think? Is there an option I've not thought of?

Related to all of this, the `usefulscripts` repo doesn't appear to have versions, but the scripts could be reliant on the emoncms codebase. In particular, these changes wouldn't be backwards-compatible with older versions of emoncms. Would it be worth moving those scripts into the `emoncms` repo, or is there another way we could manage those dependencies?

glynhudson:
Hi @spikeheap 

>Is there a canonical list (or way to generate a list) of all the things outsideemoncms/emoncms which load process_settings?

No, I don't think so. I know some users have written custom modules that work with Emoncms. My worry if we merge these changes is that is will break lots external scripts and users custom modules. 

> 1. Propagate the if($failed_settings_validation) blocks into each of those utility scripts. This is the easiest, but results in the most duplication, and margin for error in new scripts.

As I mentioned above this change would be difficult to manage and potentially break scripts / modules for users. 

> 2. Create a bootstrap.php (or similar) which does the validation and dies on error (preferably as a class). This is a little more effort, but would allow us to rationalise the common setup and preferences loading into a single place for all scripts and site. This PR wouldn't need to have all of that, so bootstrap would just be a thin facade over process_settings.

This potentially sounds like a good solution. I would need to understand a bit more about how this would work. Will this require changes that I'm trying to avoid to all scripts / modules that call process settings? 

> 3. Reinstate die within process_settings but have it conditional on some global/environmental variable. This feels fairly hacky, as it introduces more global state and increases the responsibilities of the settings processor, however this option is the only one which doesn't require any changes to the files which load the process_settings file.

What do you mean by `have it conditional on some global/environmental variable`? 

> Related to all of this, the usefulscripts repo doesn't appear to have versions, but the scripts could be reliant on the emoncms codebase. In particular, these changes wouldn't be backwards-compatible with older versions of emoncms. Would it be worth moving those scripts into the emoncms repo, or is there another way we could manage those dependencies?

Correct, the scripts don't have versions. This is a good point, the same applies to all the Emoncms modules. @TrystanLea and @chaveiro had this discussion a while back. It was decided to modularity that it was best to only include 'core' functions in the core repo and then have separate repos for all extra modules e.g. dashboard, config, wifi etc. Versioning modules and scripts would be a good idea.

chaveiro:
There is an alternative to modify all files that use process_settings.php, that is to modify just this file so it detects that it'ss being tested and return something else for the test instead of dieing there.
Also a good time to move all that checks to inside a new class and functions. 
Them from the same file do something like :
Settings->Load():
if (Settings->gotErrors())  {
    if (defined("EMOCMS_TEST_RUNNING")) throw new MySettingsException('settings are wrong!'); 
    else Settings->Process();  // this will echo errors to ui and die
}

spikeheap:
Hi @glynhudson,

From what you've said, it sounds like module management and backwards compatibility would make the more elegant solutions a lot more taxing to implement, and I'm keen to get some test coverage merged so we can iterate on it. I think [1] and [2] from the list above cannot be achieved without potentially breaking some unsuspecting scripts, so let's leave those for now.

We want to get to a point where we can completely remove the `die` from `process_settings` (and probably everywhere else) and consolidate it into a single place. To that end we'd like the following behaviour:

1. If a script is 'old', or just doesn't know how old it is, have `process_settings.php` `die` if the settings don't validate.
2. Have the emoncms core and any 'new' modules or scripts use the non-`die`ing approach. 

The solution that jumps to mind is to leverage the version portion of `$emoncms_version` to decide whether we should use the new behaviour or not, and to default to `die`ing if we don't get a positive result. That way we're not introducing any new variables, but scripts which don't define the version will continue to behave as they always have.

There is a handy utility for [comparing semver versions](https://github.com/composer/semver) which we could use to do something like:

```php
use Composer\Semver\Comparator;

function handleValidationError(message){
  preg_match('/(?:low-write )?([0-9.]+) \|.*/', get_env('emoncms_version'), $matches)
  $emoncms_version_semver = $matches[1]

  if(Comparator::greaterThan('9.7.7', $emoncms_version_semver)) {
    throw new MySettingsException(message)
  } else {
    die;
  }
}
```

We could easily roll the version comparison ourselves too. This approach can then be used to build compatible scripts in the `usefulscripts` repo by setting the `emoncms_version`. It can also be used to effect any larger architectural changes you might choose to do in future.

@chaveiro is spot-on suggesting we take a class-based approach to this. The intention of this PR is to make as few changes as possible to integrate the tests, as we can then use those tests to refactor the existing code into a class for a new PR. I intend to chase this PR quite closely with one that moves it to a `Settings` class :+1:.

It's always a good idea to have the tests execute the production logic, so I'd be loathe to introduce an `EMONCMS_TEST_RUNNING` variable into the mix. The non-test code should never need to be aware of when it's under test, and I think the issues we're coming up against here are worth solving in the general case rather than just for testing.

TrystanLea:
What's the status of this @spikeheap @chaveiro . I dont really know enough about php testing to comment and its not my area of expertise, I'm happy to defer to your expert opinion, if there is common agreement Im happy to merge this.

Issue #337, opened 2015-06-05
Post multiple values in JSON format


vosmont:
Hello,
is it possible to send multiples values in JSON format, with a timestamp ?

Perhaps something like that :
http://emoncms.org/input/post.json?json={{power:200,offset:-10},{temperature:25,offset: -5}}&time=1433504598

vosmont:
It seems that
http://emoncms.org/input/post.json?json={power:200,temperature:25}&time=1433506452
at least should work, but when I set a time different from current time, the values are not stored.

vosmont:
In fact, the value is stored only if it is after the last recorded value.
You cannot thus insert values post factum.

So my need is to be able to set an offset on the sent values.

arthurlutz:
I think I've bumped into the same bug as you @vosmont see #368 and #371, have you found a fix ?

PR #615, opened 2016-11-29
Adds support for named bulkInserts

fredrike:
eg. supports `input/bulk.json?data=[[-10,16,1137],[-8,17,{"1": 1437},{"2":3164}],[-6,19,{"data": 1412},{"anotherData":3077}]]` 

Solves #337.

### DIFF
--- a/Modules/input/input_controller.php
+++ b/Modules/input/input_controller.php
@@ -137,6 +137,12 @@ function input_controller()
                             $name = 1;
                             for ($i=2; $i<count($item); $i++)
                             {
+                                if (is_object($item[$i]))
+                                {
+                                    $value = (float) current($item[$i]);
+                                    $inputs[key($item[$i])] = $value;
+                                    continue;
+                                }
                                 if (strlen($item[$i]))
                                 {
                                     $value = (float) $item[$i];


jeremypoulter:
This could add quite a large overhead for large datasets, it would be nice to only have to send the name once. That being said it i better than nothing so probably worth merging, especially on emoncms.org

TrystanLea:
Sorry for the slow response on this. does it produce the following error for you:
PHP Warning:  strlen() expects parameter 1 to be string, object given in...

Perhaps amending to the following would be best:

if (is_object($item[$i])) {
    
} else if (strlen($item[$i])) {

}

fredrike:
Sorry for being even slower on this..

the `continue;` on line [144](https://github.com/emoncms/emoncms/pull/615/files#diff-6f4cc2f54a9c2c476d95333d0d52da0dR144) actually skips the rest of the content in the for loop and it is included to not increment the `$name` variable so it should never reach the second `if`.

TrystanLea:
Thanks @fredrike my turn to apologise again, great, is there any chance you could add this option to the input api documentation?

https://github.com/emoncms/emoncms/blob/master/Modules/input/Views/input_api.php

PR #617, opened 2016-12-01
Fixes to feed get data with mode=daily|weekly|monthly

hagai-shatz:
Feed API data.json with mode=daily|weekly|monthly is designed for cumulative feeds and is only supported by PHPFina and PHPTimeSeries engines. However, both engines do not fully support it. 
For PHPFina: this fix enable to return a value even if the value at midnight is null, by looking back to the latest valid value.
For PHPTimeSeries: this fix enable to return "weekly" and "monthly" modes.

### DIFF
--- a/Modules/feed/engine/PHPFina.php
+++ b/Modules/feed/engine/PHPFina.php
@@ -357,7 +357,7 @@ class PHPFina
         $end = intval($end/1000);
                
         // If meta data file does not exist exit
-        if (!$meta = $this->get_meta($id)) return array('success'=>false, 'message'=>"Error reading meta data feedid=$name");
+        if (!$meta = $this->get_meta($id)) return array('success'=>false, 'message'=>"Error reading meta data feedid=$id");
         $meta->npoints = $this->get_npoints($id);
         
         $data = array();
@@ -373,7 +373,9 @@ class PHPFina
         if ($mode=="monthly") $date->modify("first day of this month");
         
         $n = 0;
-        while($n<10000) // max itterations
+        $lastpos = 0;
+        $lastvalue = null;
+        while($n<10000) // max iterations
         {
             $time = $date->getTimestamp();
             if ($time>$end) break;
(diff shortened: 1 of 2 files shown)


Paul-Reed:
This looks a nice improvement, thanks Hagai.
@chaveiro I would value your opinion.

Paul

TrystanLea:
Sorry for the slow reply on this. The PHPTimeSeries changes look great happy to merge these. For PHPFina the midnight null value issue is designed to be avoided by using the power_to_kwh or wh_accumuator input processor which join across missing values as these are cumulative type feeds.

That said I would be happy to accept this change if there is a limit to how far back or forward from midnight it tries to find a valid value, it look like at the moment it could try and scan through a large number of datapoints, perhaps it could be limited to scanning back 30 minutes from midnight?

hagai-shatz:
@TrystanLea Are you sure the power_to_kwh or wh_accumuator will solve the problem? If there is no power transmitted to the input during midnight, the processor will not run and will not generate a value in the feed.
As for the lookback limit - it is only go back until the last interval (day/week/month) and no more than the absolute total reads set in the service - 10000.
I will not expect it to read back more than a few values. However, in my feeds, I'm importing data from a Solaredge inverter that does not provide any values when there is no sun so the lookback might be for a few hours in a feed with 5 minutes interval, that is 12 reads per hour to look back at the last value of the day.

Paul-Reed:
Any update pls @TrystanLea 

Thanks
Paul

TrystanLea:
Hello @hagai-shatz apologies for my slow reply, a few too many things on in parallel as usual, power_to_kwh or wh_accumuator will solve this issue for PHPFina as the missing data is filled on post, This is handled here: https://github.com/emoncms/emoncms/blob/master/Modules/feed/engine/PHPFina.php#L217 if join mode is disabled it enters null otherwise it joins with valid datapoints.

TrystanLea:
Very happy to merge the changes to PHPTimeSeries if you could send me this separately.
Thanks a lot and apologies again for the delay
Trystan

TrystanLea:
I've merged the PHPTimeSeries changes in manually, see: https://github.com/emoncms/emoncms/commit/c116b96b95be17f6fadb05c3beff41599bba1afc

PR #620, opened 2016-12-05
Encode URL GET params for registration, password change, etc.

### LINKED ISSUE #616: Email addresses with + symbols are rejected
register with a new address with a plus in is rejected as an invalid email address

spikeheap:
This PR resolves #616. 

The cause of #616 was illegal URI characters (in this case a `+`) being replaced with ` `, which in turn results in an illegal email address on the server.

Encoding all URI components is sensible because it also prevents spurious input from having unintended consequences (e.g. if a user entered the username `bob&email=somethingelse@gmail.com`). 

I've tested this change locally and it fixed it for the `+` character issue at least.

### DIFF
--- a/Modules/user/profile/profile.php
+++ b/Modules/user/profile/profile.php
@@ -249,7 +249,7 @@ function languagecode_to_name($langs) {
         {
             $.ajax({
                 url: path+"user/changeemail.json",
-                data: "&email="+email,
+                data: "&email="+encodeURIComponent(email),
                 dataType: 'json',
                 success: function(result)
                 {
--- a/Modules/user/user.js
+++ b/Modules/user/user.js
@@ -24,7 +24,7 @@ var user = {
     $.ajax({
       type: "POST",
       url: path+"user/register.json",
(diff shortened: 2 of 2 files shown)


Paul-Reed:
Not fixed here Ryan.
Using emoncms 'self build' low-write 9.7.8 | 2016.11.10, and your diff applied, cache cleared, I get;

error

Paul

Paul-Reed:
What took you so long?

spikeheap:
Hi @Paul-Reed, 

Sorry to waste your time earlier. That was a stupid typo on my behalf after I'd tested it. 

Update so it does the right thing.

Paul-Reed:
I think the same encoding needs to be applied to the password reset function, otherwise lost passwords would not be recoverable.

password

Paul

spikeheap:
👍 , good call.

Paul-Reed:
Thanks for your patience Ryan 👍 
Note - I've considered the negative 'Codacy' report, but the 'path' issue reported is not directly attributable to this PR. 

Paul

davehun:
will have to wait for it to hit emoncms.org 👍

PR #625, opened 2017-01-08
MQTT script support for non systemd systems

Paul-Reed:
As raised by pb66, removing the readme section about init.d systems, stops non systemd users from running the mqtt_input script.


glynhudson:
Perfect, that's great. Apologies for failing to remember that older raspbian OS's (wheezy) don't have systemd support. Thanks for helping to updated to docs, much appreciated.

Issue #589, opened 2016-10-29
Unicode characters


kraujinis:
Hi,
I install fresh cms 9.7.5 copy to host. I have problem with displaying Lithuanian letters.
In Visualizations feed name looks correctly.
![visualisations](https://cloud.githubusercontent.com/assets/10329476/19831368/dc5a35ca-9e10-11e6-8dc8-d98c7e43493f.png)
When I try to look in Full screen I'm missing one letter.
![multigraph](https://cloud.githubusercontent.com/assets/10329476/19831406/a78666b0-9e11-11e6-980f-9a1366af2c0d.png)
The same problem in Dashboard view.

pm-cz:
Hello, possibly my pull request #629 addresses this issue

glynhudson:
Hi @pm-cz I am trying to test your PR #629 . Could you give me an example unicode to test with? I have beeing trying to use the Lithuanian character ė as used by @kraujinis above. However unlike @kraujinis I don't seem to be able to even get Emoncms to save a feed name with this character. The ė character gets replaced with a ? when the feed name is set. See example:

![screenshot 2017-01-15 at 22 30 43](https://cloud.githubusercontent.com/assets/758844/21966386/7422bae6-db72-11e6-898a-c05cdf1ab656.png)
![screenshot 2017-01-15 at 22 31 17](https://cloud.githubusercontent.com/assets/758844/21966387/743e05da-db72-11e6-8c1b-d367483d166a.png)

Therefore am I correct that the issue is not just affecting multigraph but also the main feed list? 

I have tried #629 but it does not seem to have made a difference. 

Sorry, I am quite unfamiliar with unicode characters, we don't use them often in English, however I can see the importance of Emoncms being able to handle unicode as feed name.

pm-cz:
Hello, 
the unicode sequence I have successfully used is "směr větru" (wind direction). 

But if you are getting question mark in the feed name instead, I wonder, do you have the MySQL database configured to use utf-8 characters? Because by default the MySQL database I installed emoncms in had a problem with character set being configured to latin1 and mangling all unicode characters. But emoncms.org works with these feed names (but replaces unicode in the multigraph labels).

pm-cz:
The problem is probably in the [Installation guide](https://github.com/emoncms/emoncms/blob/master/docs/LinuxInstall.md) where creating the database has following command:

`mysql> CREATE DATABASE emoncms;`

But the correct command to use UTF-8 would be

[...]

PR #629, opened 2017-01-15
Support for unicode characters in multigraph labels

pm-cz:
Hello,
the multigraph labels mangle unicode sequences dropping them from labels in multigraph, even when they are permitted in feed names. I have changed the line to match the pattern on $name line and permit unicode sequences as well.


glynhudson:
I've just responded on https://github.com/emoncms/emoncms/issues/589

PR #631, opened 2017-01-16
Create MYSQL db with utf8 decoding

Paul-Reed:
Update docs to ensure that the MYSQL database is created with utf8 encoding.


glynhudson:
thanks

PR #642, opened 2017-02-26
Trim spaces when receiving json input in post.json.

### LINKED ISSUE #641: post.json incompatible with JSON produced by Python
In particular, Python's `json` module typically includes spaces between components of its output.  A typical return value from `json.dumps` might be

    {"vrms": 245.12, "pulse": 1, "temp3": 300, "temp2": 300, "temp1": 300, "temp6": 300, "temp5": 300, "temp4": 300, "rssi": "-27", "power4": 0, "power3": 0, "power2": 0, "power1": 384}

In this case, emoncms splits the key/value on the ':', but this means that the space between the comma and the key for all but the first value is included in the key itself and new inputs are created with a space at the beginning of their name (which you can't actually see in the rendered output).

This doesn't affect use of emoncms with the built in utilities because for pushing groups of values normally they're either posted sequentially (using the nodevars format) or in the bulk format.  However, `post.json` claims to be able to take input of this form, and by adding it we can improve handling of input from other sources.

adpeace:
This makes the JSON input compatible with that generated by Python's json
module

Fixes #641

### DIFF
--- a/Modules/input/input_controller.php
+++ b/Modules/input/input_controller.php
@@ -231,7 +231,7 @@ function input_controller()
                         if (isset($keyvalue[1])) {
                             if ($keyvalue[0]=='') {$valid = false; $error = "Format error, json key missing or invalid character"; }
                             if (!is_numeric($keyvalue[1])) {$valid = false; $error = "Format error, json value is not numeric"; }
-                            $data[$keyvalue[0]] = (float) $keyvalue[1];
+                            $data[trim($keyvalue[0])] = (float) $keyvalue[1];
                         } else {
                             if (!is_numeric($keyvalue[0])) {$valid = false; $error = "Format error: csv value is not numeric"; }
                             $data[$csvi+1] = (float) $keyvalue[0];


TrystanLea:
Hello adpeace, thanks for this pull request, This looks great and results in preferable behaviour, there is however a risk that existing users with leading spaces in their inputs will be affected. Id be happy to merge this but I foresee potential issues @pb66 @chaveiro @Paul-Reed What do you think? Should I merge and place a notification on the forums?

adpeace:
Is there a place in the code where you could do database version migrations?  If so could you strip the spaces from values in the database there?

TrystanLea:
Unfortunatly not really anything setup to do that at the moment and Id be worried about issues resulting when trying to update all the emoncms instances out there.

Emonhub is written in python and works fine with emoncms, perhaps the best solution is to prepare your request to remove any spaces?

chaveiro:
How about using json_decode to process the json received?

TrystanLea:
The input/post API has always accepted non strict JSON strings so making it strict would affect running systems. For example: "http://localhost/master/input/post.json?node=1&json=power1:100,power2:200,power3:300" is accepted.

I vote for fixing this one outside of emoncms if possible, accepting leading spaces is not technically an emoncms bug as this is the input given.

adpeace:
> Emonhub is written in python and works fine with emoncms, perhaps the best solution is to prepare your request to remove any spaces?

The reason it works fine is because normally either the bulk import is used or only one reading is submitted at a time and Python's JSON library doesn't include a space after the opening brace.  In my client to emoncms, I'm posting multiple readings to it in one call.  The JSON I'm providing is definitely valid and does not contain spaces that should not be there or that should be parsed as part of the key name.  E.g. `{ "foo": "bar" }` would be erroneously parsed by emoncms currently as a key `" foo"` and value `" bar "` (though because the values are numeric ordinarily, the space that is captured in the string gets ignored by the conversion and so masks the problem.

In the general case, removing spaces in the client is problematic - if that is the required solution then I would just carry this patch in my version of emoncms instead.  The main problems are:
* Clients are free to insert those spaces as part of valid JSON, and if a user updates or changes the version of a JSON library they are using then presence or absence of the spaces might change.
* Generically taking a JSON string with spaces and removing them would require parsing it, which is a bunch of work that is error-prone just to implement compatibility with a buggy server.

> How about using json_decode to process the json received?

That would be better.  I am not too familiar with PHP so wasn't aware of that function, but definitely using a proper json decoder would be better.

> I vote for fixing this one outside of emoncms if possible, accepting leading spaces is not technically an emoncms bug as this is the input given.

This is definitely an emoncms bug: JSON of the form given is incorrectly parsed to include spaces when it should not.  I do understand your backwards compatibility concerns though, of course.

I think very few users would be affected for the reason I mentioned above, however to avoid that problem altogether, how about adding a new mode (via a 'strict' flag to post.json that defaults to false) that accepts proper JSON, using the decoding function suggested.  The existing behaviour would be preserved if that flag is not present.

If that solution is acceptable, I'd be happy to take a look at implementing it if it would be helpful.

TrystanLea:
I see your issue, I appreciate { "foo": "bar" } becoming " bar" is a bug. I thought you where sending the equivalent of " bar".

How about adding something like this which would remove the trailing space in the case of its occurrence proceeding double quotes:

$datain = str_replace(' "',"",$datain);
$json = preg_replace('/[^\p{N}\p{L}_\s-.:,]/u','',$datain);

Or to send from python without whitespaces:

http://stackoverflow.com/questions/16311562/python-json-without-whitespaces

adpeace:
> $datain = str_replace(' "',"",$datain);

Could do; that would still be a change of behaviour for clients that had been hitting this issue before and therefore had the space in the database.  The reason I noticed this in my case is I had a lot of inputs with logging configured before switching to pushing multiple values at once, at which point I got two sets of inputs that looked the same (one with the invisible space) in the UI and my original ones weren't updating.  

I suppose my preferred solution is probably the `strict` flag to make post.json parse the input using `json_decode`.

borpin:
I'll add that I have been bitten by this bug. I'd vote for some means of offering a strict JSON so there was an option to use either and possibly a script to trim the spaces in the database of an existing setup? Perhaps then a global flag could be set so all inputs are processed correctly?

It is tricky to solve without breaking existing installations as the space is introduced in the parsing.

TrystanLea:
I think a strict flag could work, I think the easiest way would be to add the option to set &strict=true in the request. Otherwise we would need a way to record on a per user basis their input settings which would get more complex.

adpeace:
Closing and will create a new pull request with the approach discussed here.

PR #650, opened 2017-03-04
Add mode for strict JSON input parsing.

### LINKED ISSUE #641: post.json incompatible with JSON produced by Python

adpeace:
Fixes #641 

This pull request replaces the previous one that added a `trim` around the keys to avoid a problem where parsing could introduce additional spaces in the input names.

The approach here is to add a new input field `fulljson` ('full' meaning 'parsed with full json parser') that is parsed differently (using `json_decode`) and leaving the other input fields as-is (parsed using the old string-splitting method).

The change also adds a check for the input having been valid before attempting to process it that was previously missing - this avoids a situation where PHP raises errors because the contents of `$data` were not of the expected type to iterate over.

### DIFF
--- a/Modules/input/input_controller.php
+++ b/Modules/input/input_controller.php
@@ -223,49 +223,67 @@ function input_controller()
                 if (isset($_GET['time'])) $time = (int) $_GET['time']; else $time = time();
 
                 $datain = false;
-                // code below processes input regardless of json or csv type
+                /* The code below processes the data regardless of its type,
+                 * unless fulljson is used in which case the data is decoded
+                 * from JSON.  The previous 'json' type is retained for
+                 * backwards compatibility, since some strings would be parsed
+                 * differently in the two cases. */
                 if (isset($_GET['json'])) $datain = get('json');
+                else if (isset($_GET['fulljson'])) $datain = get('fulljson');
                 else if (isset($_GET['csv'])) $datain = get('csv');
                 else if (isset($_GET['data'])) $datain = get('data');
                 else if (isset($_POST['data'])) $datain = post('data');
 
                 if ($datain!="")
                 {
-                    $json = preg_replace('/[^\p{N}\p{L}_\s-.:,]/u','',$datain);
-                    $datapairs = explode(',', $json);
-                    $data = array();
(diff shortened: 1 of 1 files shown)


TrystanLea:
Great, works well! Could you remove the printing of $datain on failure to decode? From what I understand doing so like this can create a security issue.

http://localhost/master/input/post.json?node=1&fulljson="test"
this input also passed the $valid test giving the following error:

<br />
<b>Warning</b>:  Invalid argument supplied for foreach() in <b>/var/www/master/Modules/input/input_controller.php</b> on line <b>269</b><br />
ok

adpeace:
Updated the branch.

glynhudson:
Thanks a lot, very sorry so long to merge this. 

Would it be possible to add the new fulljson API with an example to the input API helper page: 

https://github.com/emoncms/emoncms/blob/master/Modules/input/Views/input_api.php

adpeace:
Hi; yes, I'd be happy to add an entry on that page, I can submit a PR in the next couple of days.  The example you gave isn't quite right: the `fulljson` is the name of the parameter, not the object (i.e. `post.json?node=1&fulljson=...`).

glynhudson:
Thanks! Ah yes, the following fulljson post works great :100: 

http://emonpi/emoncms/input/post.json?node=1&fulljson={"power1":1100,"power2":2100,"power3":3100}

PR #651, opened 2017-03-08
Added softlink support for module schemes

adminde:
Make load_db_schema() aware of softlinks, to enable database table creation of
external modules

PR #654, opened 2017-03-22
Add softlink support for module processlists

adminde:
Modules with only a softlink in the Modules directory do not get their *_processlist.php processed correctly. 
Make Process->load_modules() aware of softlinks, to enable custon processes of external modules.

Almost identical to my last pull request regarding [module schemas](https://github.com/emoncms/emoncms/pull/651) and again just a very (very) small commit :)


glynhudson:
nice, thanks 👍

PR #658, opened 2017-04-13
Update processlist help text to use auto-append 

pgalbavy:
Further changes to processlist help text;

1. Abstracted out generic text to process_ui.js and moved static definitions in there
2. Setting a key 'nochange' to true says that the value will not change
3. If the key 'requireredis' is set and true it will add test t that effect
4. Also added unused 'help_page' and 'help_url' key tests to add a "See Also: ABC" test to help. Both must be set; 'help_page' is the link text and 'help_url' is the link

There is minimal checking of strings here but all sources of data should be inside the code.

Also updates eventp and schedule to match the main process Module.


glynhudson:
thanks :+1:

PR #663, opened 2017-04-21
Display phpmodules versions on admin page

pb66:
See #660

Issue #660, opened 2017-04-19
Emoncms calculation errors using redis 3.1.0 and 3.1.2


pb66:
I have found emoncms will stop accumulating when a limit defined by the update rate and size of increase reaches a certain level.

In my case recording PV using a power to kWh process to record total kWh's has maxed out at 10000kWh's and has not recorded any increase since reaching that threshold.

See the ["10000 kWh ceiling in emoncms?"](https://community.openenergymonitor.org/t/10000-kwh-ceiling-in-emoncms/3750?u=pb66) thread on the OEM forum, I can provide more data and screenshots if required.

I'm not sure where to go with this as I'm not good with PHP, this is the ["power to kWh" process in question](https://github.com/emoncms/emoncms/blob/master/Modules/process/process_processlist.php#L272-L300)

I know there was a change, way back to make the "value" field in the feeds table (mySQL) a double but I have no idea if the sums are being done using floats or doubles, also is it possible that Redis could cause a "bottle-neck" somewhere? 

Any assistance appreciated

chaveiro:
I remember the cause was adding a very small value to a large float that has lower precision ended up with a close number that when rounded is the same as it started before the add. Are you using redis?

pb66:
Yes that is exactly the problem I'm seeing and yes I am using Redis but I do not know enough about how it works to debug. Do you have any suggestions?

joyrider3774:
@pb66 do you have a local installation of emoncms ?

if so make backups first (just to be safe) and try this change then update the database:

It's a change i did to my local installation to prevent the problem i had with smaller values and getting the same value after rounding. As soon as i added that a long time ago i never had that problem again (with rounding errors).  Also i don't know what it currently is set at, since i'm not using latest emoncms

![image](https://cloud.githubusercontent.com/assets/10714776/25204640/cbe2e482-255e-11e7-89fb-c1ea00de2109.png)

joyrider3774:
sorry it probably is not related to this as i read through the old forum thread describing this problem i had initially it seems the problem did not happen with people using redis. https://openenergymonitor.org/forum-archive/node/11774.html . I did not read thourougly enough

TrystanLea:
@pb66 I imagine that changing the precision in mysql does not solve your problem as your using redis?

I've been trying to replicate the issue this morning, to no avail so far.
[...]

PR #665, opened 2017-04-24
Display phpmodules versions on admin page - Part 2

JonMurphy:
See issue #660 and issue #663.

The php modules are in alphabetical order, round brackets removed, added "v" for version.


JonMurphy:
@joyrider3774 - can you look at this change (L331) and determine if it works for Windows users?

joyrider3774:
@JonMurphy i'll check when i find some time. I haven't updated my emoncms in a while and i have customn changes to the code so will have to reapply them afterwards. I'll let you know when i had found the time

PR #668, opened 2017-04-28
Device autocreate

TrystanLea:


TrystanLea:
Thanks @JonMurphy good point, I will make that change

PR #688, opened 2017-07-04
Theme adjustments and sun theme

adminde:
With the (photovoltaic research) institute I work with using emoncms for a while now - and my love for emoncms and its modularity and themes - I made a yellow "sun" theme.

With the blue theme emerging I noticed, that the login screen background was changed to being blue in #680, which clashes a little with the yellow navbar :)
I added a body-login class to the theme styles and then add this class to the body in the login-form, instead of a hardcoded color. Dunno if this is best-practice to be honest.
Further, I renamed the emoncms_logo.png to logo_login.png to make it a little more generic for potential customization and at last, added the sun theme. Which to be honest is a yellow theme... but yellow does not sounds very romantic and... well... we are a solar institute.

I hope this is not too arbitrary and maybe someone else takes a like in a yellow emoncms.


adminde:
The failing codacy checks are caused by the background-color **!important;** annotations, which originated actually just from the emon-blue theme.

I don't know if this can be ignored for the base theme, as the codacy warning is just to avoid inflationary use of the annotation?

TrystanLea:
Looks great!! thanks 
suntheme

PR #690, opened 2017-07-13
Fixed issue with split on 30 minute interval

stuartpittaway:
Hi, I've discovered a bug when requesting data from "get_data_DMY_time_of_day" on 30 minute split points, which causes duplicate data to be returned.   Looks like a simple Python floating point number rounding issue.

I've also increased maximum split points from 24 to 48 to allow a days worth of 30 minute data to be returned for a new energy cost comparison (time of use based) app for emoncms.


chaveiro:
Python ?

stuartpittaway:
Sorry, was working on several different projects at the same time, meant
PHP !!

TrystanLea:
Great thanks @stuartpittaway

rcojoe:
@TrystanLea On the cloud version,  this hasn't been implemented yet.  Is there an ETA?
https://github.com/emoncms/app/issues/46

PR #692, opened 2017-07-18
EmonCMS 9.8.8 | 2017.07.18

chaveiro:
Minor bug fixes.
Code refactor


TrystanLea:
Thanks @chaveiro

.travis.yml
```diff
@@ -3,6 +3,7 @@ php:
   - '5.4'
   - '5.5'
   - '5.6'
+  - '7.0'
```
glynhudson: Hi @chaveiro, is there are reason you have added PHP7 checks? It causes Travis CI tests to fail https://travis-ci.org/emoncms/emoncms/jobs/254672803

chaveiro: I use PHP7, was expecting travis to be able to test it, but now i see it doesn't.

PR #693, opened 2017-07-19
Deal with last value nulls

chaveiro:


TrystanLea:
Hello @chaveiro 

I think your modification to the feed model might work but the input model fails under the case where the value is NAN, see this test:

    <?php

    $lastvalue = array();
    $lastvalue["time"] = NAN;

    if ($lastvalue['time']) {
        $lastvalue['time'] = (int) $lastvalue['time'];
    } else {
        $lastvalue['time'] = null;
    }

    print json_encode($lastvalue)."\n";

TrystanLea:
Sorry I meant this test (the above works):

    <?php

    $lastvalue = array();
    $lastvalue["time"] = NAN;

    $lastvalue['time'] = ($lastvalue['time'] ? $lastvalue['time'] : null);

    print json_encode($lastvalue)."\n";

TrystanLea:
I think we will need to keep monitoring this one to see if the issue appears again, it took me ages to find the failure point, and Im sure I needed all the checks I added to sort the issue highlighted here (https://community.openenergymonitor.org/t/input-list-doesnt-load/3797/49:) but Im happy if your proposed change is sufficient.

For reference this is what I had previously in the input_model.php 

    $lastvalue['value'] = $lastvalue['value'] * 1; 
    $row['value'] = (float) $lastvalue['value'];
    if (is_nan($row['value'])) $row['value'] = 0;

The feed model only had:

    $lastvalue['value'] = $lastvalue['value'] * 1;
    $lastvalue['value'] = (float) $lastvalue['value'];

chaveiro:
On redis hmget documentation it can only return an array (can be empty) or false if redis has no data (eg: after a flush).
For the array value to be NaN it means somewhere it is being wrongly written.
The redis input:lastvalue is only changed by the follow files:
- Modules/input/input_controller.php 
- Modules/process/process_processlist.php  <- Not used
- scripts/input_queue_processor.php
- scripts/phpmqtt_input.php:

So the reason that it might return NAN comes from one of the above files.
Anyway, i've forced it to NaN as a test and got no error from PHP. I'm using php7.

TrystanLea:
Thanks @chaveiro, I will flag this change up on the thread where we where discussing it and see if it comes back.

This check does result in a json_encode bug for me:

    $lastvalue['time'] = ($lastvalue['time'] ? $lastvalue['time'] : null);

is there any reason we cant use:

if ($lastvalue['time']) {
    $lastvalue['time'] = (int) $lastvalue['time'];
} else {
    $lastvalue['time'] = null;
}

chaveiro:
Where does that json encoding error comes?

TrystanLea:
Are you asking the initial source? pre redis? or how the error is displayed?

chaveiro:
Follow your recommendation, see https://github.com/emoncms/emoncms/pull/694

PR #695, opened 2017-07-22
Update admin_main_view.php

AndyTaylorTweet:
Updated to ad the ability to parse in versioning information via JSON.

### DIFF
--- a/Modules/admin/admin_main_view.php
+++ b/Modules/admin/admin_main_view.php
@@ -299,8 +299,8 @@ if ( @exec('ifconfig | grep b8:27:eb:') ) {
               }
               if (isset($emonpiRelease)) {
                 echo "<tr><td class=\"subinfo\"></td><td>Release</td><td>".$emonpiRelease."</td></tr>\n";
-                echo "<tr><td class=\"subinfo\"></td><td>File-system</td><td>Set root file-system temporarily to read-write, (default read-only)<button id=\"fs-rw\" class=\"btn btn-danger btn-small pull-right\">"._('Read-Write')."</button> <button id=\"fs-ro\" class=\"btn btn-info btn-small pull-right\">"._('Read-Only')."</button></td></tr>\n";
               }
+                echo "<tr><td class=\"subinfo\"></td><td>File-system</td><td>Set root file-system temporarily to read-write, (default read-only)<button id=\"fs-rw\" class=\"btn btn-danger btn-small pull-right\">"._('Read-Write')."</button> <button id=\"fs-ro\" class=\"btn btn-info btn-small pull-right\">"._('Read-Only')."</button></td></tr>\n";
 }
 
 // Ram information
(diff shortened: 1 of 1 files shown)


Paul-Reed:
That looks good Andy. I've just added a further module.json file to the app module to assist with testing (I'll create v1.0.0 tonight), and that also has updated the admin page.

Do we need to keep the same format in both 'emoncms modules' & 'php modules'? ie dashboard(1.0.0) or dashboard v1.0.0

I see that the PR failed Travis, I think that maybe an issue with PHP 7.0 rather than your PR (I seem to recall recent discussions between @chaveiro & @TrystanLea 

Also, I'm picking up a browser error which existed **before** your PR;
`Notice: Undefined variable: emonpiRelease in /var/www/emoncms/Modules/admin/admin_main_view.php on line 300` (line 295 in emoncms current master) so will look at that separately.

AndyTaylorTweet:
I believe the emonpiRelease can be solved by checking if the file in /boot exists before trying to use its name as a variable. If you want me to take a look at that - I can do so.

Formatting the output to suit - thats kinda up to you guys, I put the version string in brackets right after the module name, because that was quick and simple, if you want it to do something else, you can amment the PR to do what you want or guide me and I'll adjust it, the code is there to be used however you want :)

Paul-Reed:
Andy
If you'd like to sort the emonpiRelease error, that would be great. Thanks.

Regarding format of the displayed version, I think if it's the same as the PHP entries, it would be good for consistency, eg Dashboard v1.0.0

Paul

AndyTaylorTweet:
Formatting modified to match the PHP versions lower in the page.
Now looking at the other issue.

AndyTaylorTweet:
And now added a check to see if the variable "$emonpiRelease" exists before its used.
Paul - please test that for me and see if it works correctly for you.

Paul-Reed:
Thank you @AndyTaylorTweet 
The fix has resolved the issue OK, and the addition of the module versions on the admin page is a nice addition, and well executed.

Paul

glynhudson:
Nice work guys. This is awesome :+1: 

>I see that the PR failed Travis, I think that maybe an issue with PHP 7.0 rather than your PR (I seem to recall recent discussions between @chaveiro & @TrystanLea

This has been fixed by #697

glynhudson:
Works great.

Would it be possible to extend to grab the module name (as displayed in the admin list) from `name` in `module.json` if the file exists and if not fall back to the old method? 

I've just added a separator to better distinguish between modules in the list: dc78f639c71f9ed06ee192fea30681fe6b7b6010

selection_065

glynhudson:
When we're ready to go these changes will need to be merged into stable branch on Emoncms core and Dashboard module. We should probably also make a `stable` branch on the other modules.

AndyTaylorTweet:
Yes, working up a PoC now.

PR #700, opened 2017-07-24
Add module.json to core modules?

Paul-Reed:
As a number of the modules are included in the main emoncms repo (& unlikely to be moved) it has created a mismatch of capitalized and lower-case names in the module list.
By adding a module.json file to each of the core modules too (minus the versioning), we can create capitalized 'friendly' name for all modules.
Let me know if you want me to progress this.

Example;

Before -
event

After -
event1

### DIFF
--- a/Modules/admin/module.json
+++ b/Modules/admin/module.json
@@ -1,0 +1,4 @@
+{
+    "name"         : "Administration",
+    "version"      : ""
+}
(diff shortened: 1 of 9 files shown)


glynhudson:
Fine by me, makes sense. After all they are modules even if they are in the core. No reason why they can't be versioned also...maybe a step too far atm. Good to have to option for the future. I would be happy for you to merge this. I'll leave it open for 24/48hrs to get some more feedback.

PR #701, opened 2017-07-26
initial support for Apache Cassandra

mabi:
Experimental development for Apache Cassandra. Feed engine works. Table structure and code to be improved.


Modules/feed/engine/CassandraEngine.php
```diff
@@ -0,0 +1,248 @@
...
+    /**
+     * Create feed
+     *
+     * @param integer $feedid The id of the feed to be created
+     * @param array $options for the engine
+     *
+     * {@inheritdoc}
+     *
+     */
+    public function create($feedid,$options)
```
TrystanLea: Can you add $feedid = (int) $feedid; here to help with input sanitisation?

TrystanLea: I've fixed in latest commit

TrystanLea:
Thanks Mabi, I havent been able to test with Cassandra but I have tested to check it doesnt affect a standard emoncms system and quickly reviewed the engine code and added extra input sanitisation

mabi:
Thank you @TrystanLea. I am testing the code with a single emonpi, so far so good. I will improve the code and implement missing features (eg cvs export). I think Cassandra is well suited for this kind of data.

PR #709, opened 2017-08-14
Device dialogs and dynamic templates for modules

adminde:
Hi,

I (again) hope this is not too arbitrary.
As a big fan of all these device template functionalities, I was hoping to maybe be able to do something worthwhile here with my time (and maybe add some features to further improve our [device driver solution](https://github.com/isc-konstanz/emonmuc) integration).

As a small proposal in this pull request:
 - Add a device configuration dialog with a template-selection sidebar both in the device, as well as the input view
 - Adapt existing templates to group them e.g. by "group" and "origin", to make device selection easier for hopefully increasing template numbers in the near future ;)
 - Make template functions dynamically providable for all modules by looking for *module_template.php* files in all modules

I hope I tested the proposed changes thoroughly enough and you maybe have the time to take a look at them


TrystanLea:
Thanks @adminde, this looks great!
Still working though trying to get my head around all of the changes. 

I noticed an issue with deleting a device from the inputs view, it doesnt seem to delete the inputs in the way I coded it, it may be an issue with using a single implementation dialog? The behaviour I am after is where deleting the device in the inputs list deletes the inputs for that device at the same time

Thanks a lot

Trystan

TrystanLea:
Is there a particular reason that you needed both the group and origin properties in the device specification? I would be inclined to simplify and keep it to one grouping option, perhaps just the 'group' property?

TrystanLea:
Id also be interested to hear more about your reason for:

"Make template functions dynamically providable for all modules by looking for module_template.php files in all modules"

How does this help with integration with OpenMUC?

Thanks

Trystan

adminde:
Well to be honest, I completely missed the part with the automatic input deletion.
I just added it in the last commit :)

Regarding OpenMUC, I was hoping to use the json templates and the whole device concept, to ease up the configuration process of the OpenMUC data structure.
As the basic principle, OpenMUC is split into 3 classes - drivers, devices and channels. Drivers implement specific communication protocols and can register several devices, which correspond to "connections" e.g. to specific hardware devices. Each device can register channels that can be read or written to. They commonly represent e.g. registers of energy meters and in my emoncms-implementation, the channels post their read values to inputs.
This whole configuration can be pretty complex, but is pretty much the same for specific hardware.
In Konstanz for example, household main energy meters are commonly from Elster (Type AS1440, AS1500, etc. depending on consumption or if bidirectional measurement is necessary or not), which can communicate e.g. via serial IEC62056-21, ModBus, etc. ... their necessary configurations could easily be pre-configured via templates.

Which is, why I thought of this split into group and origin^^ it seemed the most comfortable or "intuitive" to me, to browse device templates for this specific scenario e.g. like "Energy meters" (group) -> "Elster" (origin/manufacturer) -> "AS1440".

chaveiro:
Hi @adminde  Name, Location and Type aren't really enough?

adminde:
Well, sure one category could be enough (for now) ;)
This is just a proposal, but let me try to defend it. In the end, the amount of categories is irrelevant for the template-implementation, as only the type is used to identify the json file.

My proposal with _groups_ and _origins_ originated actually really in the variety of energy meters and smart plugs we use here at ISC, for which I hope to provide templates in the future.
With just one category, e.g. **Energy Meters**, the list could still be pretty overwhelming pretty fast, like:
- Energy Meters
    - B+G e-tech DRS110M
    - B+G e-tech DRS110ME
    - B+G e-tech DRS155D
    - B+G e-tech DRS155DE
    - ...
    - Elster A1350
    - Elster AS1440
    - Elster AS1500
    - ...
- Temperature Sensors
    - ...

I just thought, that with about 15 templates or more in one category, device selection could get pretty confusing and a non-optimal design, so why not think big ;)
Personally, I dont feel like having this additional (at the moment "unnecessary") category reduces the usability or stirs up confusion, as I kinda prefer structure. So why not having it already

But I can easily remove it, if you guys agree on thinking it unnecessary but like everything else^^

TrystanLea:
Thanks for fixing input deletion @adminde 

I've made a couple of modifications which I think improves how the device dialog looks and makes the device selection categories more intuitive.

devicelayout

- I think a smaller overall dialog size is nicer so I dropped the width to 60% with a 20% left margin.
- I then changed the table to a multi line set of device properties.
- I added the template heading to indicate that this is key device setting
- By swapping the properties for origin and group around it now makes much more sense to me as all the OpenEnergyMonitor devices are grouped under the OEM heading, I then set the origins to EmonTx, EmonPi etc.
- I changed the order so that OpenEnergyMonitor is always at the top ;)

Now OEM device templates look like this:

    {
	"name": "EmonPi: HomeEnergyMonitor",
	"group": "OpenEnergyMonitor",
	"origin": "EmonPi",

I wonder if we change the naming to 

{
	"name": "EmonPi: HomeEnergyMonitor",
	"manufacturer": "OpenEnergyMonitor",
	"model": "EmonPi",

Would that be more appropriate?

TrystanLea:
@adminde I like the modularisation of the device dialog, that's really nice, a good step forward.

I have not reviewed the implementation of load module_template. 
@chaveiro as the original author of the device module are you happy with the modified implementation?
https://github.com/emoncms/emoncms/pull/709/commits/c49e5ccb41e71c529f9794073acfa6b06e86d2f9

TrystanLea:
I still dont quite understand the need for the option to load alternative versions of Modules/device/device_template.php?

Do you envisage there being completely different implementations of this class?

adminde:
Well shame on me, I shoud've have elaborated this... right now, the template list will be requested from every available *_template.php file of each model, similar to the process model.
Function calls regarding device templates, e.g. the initializtion of a template, will then be handled by the originating *_template.php class. By doing so, I want to enable the implementation of custom device template processing for modules.
In essence, this would be great for us, as we could write our own templates which additionally specify e.g. serial port parameters, to communicate with specific hardware :)

TrystanLea:
@adminde ok that sounds great, if @chaveiro is happy, Im happy to go forward with this.

Did you have any thoughts on my previous post and proposed changes? including change from group, origin to manufacturer, model?

adminde:
You are absolutely right, my version had a lot of whitespace^^ I did this because I had in mind to implement some sort of "options" declaration for template json files in the future, where specific parameters can be declared in the json files, configured in the device dialog and passed to the generic template class, to provide a bit more flexibility for the template implementation.
Hence, I like your changes, as this is far from being done and the whitespace is just unnecessary ;)
Maybe the description input could be reduced in size to fit the others...

And yea, this categorization is the hard part imho^^
I initially just though of a 1st tier grouping, to get the user in the right direction, before confronting him with manufacturer names, like
- Smart Plugs
- Energy Meters
- Electric Vehicles
- Heat Pumps
- Temperature & Humidity Sensors

But I agree - the OpenEnergyMonitor products do not easily fit into such categories and should be promoted prominently^^
Maybe I can convince you of a more "generic" 1st tier, something like:
~~~
{
"name": "EmonPi: HomeEnergyMonitor",
"category": "OpenEnergyMonitor",
"group": "EmonPi",
~~~
which could then look like:
- OpenEnergyMonitor
    - EmonPi
    - EmonTx
    - ...
- Raspberry Pi
    - GPIO
    - 1-Wire
- Smart Plugs
- Energy Meters
    - Elster
    - ...
- Electric Vehicles
    - OpenEVSE
- Heat Pumps

etc. ;) in the end its very hard to have a convincing user guidance

Paul-Reed:
Many people may not choose to use these features, so I'm surprised it's
being added to emoncms core, instead of an optional module(s).
Has this been considered? Is it possible?

TrystanLea:
@Paul-Reed the main reason for the integration of the device module in the core (at this point) is that these changes integrate into the input module and interface itself. So that a group of inputs become a device. This is part of the development I outlined here https://community.openenergymonitor.org/t/development-devices-inputs-and-feeds-in-emoncms/4281 

@adminde development is an extension of this.

TrystanLea:
@adminde 

Happy for something more generic like:

    {
        "name": "EmonPi: HomeEnergyMonitor",
        "category": "OpenEnergyMonitor",
        "group": "EmonPi",

or your initial proposal with origin as the first level and group as the second?

    {
        "name": "EmonPi: HomeEnergyMonitor",
        "origin": "OpenEnergyMonitor",
        "group": "EmonPi",

I dont mind :)

adminde:
I just updated the code for the dialog to have a little less unnecessary whitespace and altered available templates to fit the _category/group_ scheme (and always have the OpenEnergyMonitor category prominently at first place^^) as discussed.

device_dialog

TrystanLea:
Great thanks @adminde I will merge this in, thankyou for the contribution

PR #710, opened 2017-08-18
Input improvements

TrystanLea:
emoncms input improvements

- AES 128bit Encryption route
- removal of .json requirement
- ability to use subaction for node name, e.g: input/post/emontx?csv=100,200,300
- new input/get/nodename api

### DIFF
--- a/Modules/input/Views/input_api.php
+++ b/Modules/input/Views/input_api.php
@@ -68,10 +68,18 @@
 </table>
 
 <br>
+<p><b><?php echo _('Input get'); ?></b></p>
+<table class="table">
+<tr><td><?php echo _('List all nodes and associated inputs:'); ?></td><td><a href="<?php echo $path; ?>input/get"><?php echo $path; ?>input/get</a></td></tr>
+<tr><td><?php echo _('List inputs for specific node:'); ?></td><td><a href="<?php echo $path; ?>input/get/emontx"><?php echo $path; ?>input/get/emontx</a></td></tr>
+<tr><td><?php echo _('Fetch specific input from node:'); ?></td><td><a href="<?php echo $path; ?>input/get/emontx/power1"><?php echo $path; ?>input/get/emontx/power1</a></td></tr>
+</table>
+
+<br>
 <p><b><?php echo _('Input actions'); ?></b></p>
 <table class="table">
 <tr><td><?php echo _('List of inputs with latest data'); ?></td><td><a href="<?php echo $path; ?>input/list"><?php echo $path; ?>input/list</a></td></tr>
-<tr><td><?php echo _('Get inputs configuration'); ?></td><td><a href="<?php echo $path; ?>input/get_inputs"><?php echo $path; ?>input/get_inputs</a></td></tr>
+<tr><td><?php echo _('Get inputs configuration (last time and value not included)'); ?></td><td><a href="<?php echo $path; ?>input/get_inputs"><?php echo $path; ?>input/get_inputs</a></td></tr>
 <tr><td><?php echo _('Set input fields'); ?></td><td><a href="<?php echo $path; ?>input/set?inputid=0&fields={'description':'Input Description'}"><?php echo $path; ?>input/set?inputid=0&fields={'description':'Input Description'}</a></td></tr>
 <tr><td><?php echo _('Delete an input'); ?></td><td><a href="<?php echo $path; ?>input/delete?inputid=0"><?php echo $path; ?>input/delete?inputid=0</a></td></tr>
 <tr><td><?php echo _('Clean inputs without a process list'); ?></td><td><a href="<?php echo $path; ?>input/clean"><?php echo $path; ?>input/clean</a></td></tr>
--- a/Modules/input/Views/input_api.php
+++ b/Modules/input/Views/input_api.php
@@ -111,6 +111,8 @@ td:nth-of-type(2) { width:4%;}
 7. Verify the result. The result is a base64 encoded sha256 hash of the json data string.
 </p>
 
+<p><a href="https://github.com/emoncms/emoncms/blob/input-improvements/docs/input_encrypted.md">PHP Example source code</a></p>
+
 <table class="table">
 <tr><th>Description</th><th>Method</th><th>Example</th></tr>
 <tr><td><?php echo _('URL and POST parameters:'); ?></td><td>POST</td><td>URL: /input/encrypted, POST BODY: username=USERNAME&data=IV+CIPHERTEXT</td></tr>
--- a/Modules/input/Views/input_api.php
+++ b/Modules/input/Views/input_api.php
@@ -17,30 +17,29 @@
 <table class="table">
     <tr><td><?php echo _('The input list view'); ?></td><td><a href="<?php echo $path; ?>input/view"><?php echo $path; ?>input/view</a></td></tr>
     <tr><td><?php echo _('This page'); ?></td><td><a href="<?php echo $path; ?>input/api"><?php echo $path; ?>input/api</a></td></tr>
-    <tr><td><?php echo _('Input processing configuration page'); ?></td><td><a href="<?php echo $path; ?>input/process?inputid=1"><?php echo $path; ?>input/process?inputid=1</a></td></tr>
 </table>
 
 <h3><?php echo _('Available JSON commands'); ?></h3>
-<p><?php echo _('To use the json api the request url needs to include <b>.json</b>. The "fulljson" format is recommended for new integrations, it uses the PHP JSON decoder and answer is also in json. The "json like" format is based on the CSV input parsing implementation and maintained for backwards compatibility.'); ?></p>
-
+<p><?php echo _('To use the json api the request url needs to include <b></b>. The "fulljson" format is recommended for new integrations, it uses the PHP JSON decoder and answer is also in json. The "json like" format is based on the CSV input parsing implementation and maintained for backwards compatibility.'); ?></p>
+<p><?php echo _('A node name can be a name e.g: emontx or a number e.g: 10.'); ?></p>
 <p><b><?php echo _('Post data (using http get)'); ?></b></p>
 <table class="table">
-    <tr><td><?php echo _('JSON format'); ?></td><td><a href="<?php echo $path; ?>input/post.json?node=1&fulljson={%22power1%22:100,%22power2%22:200,%22power3%22:300}"><?php echo $path; ?>input/post.json?<b>node=1</b>&fulljson={"power1":100,"power2":200,"power3":300}</a></td></tr>
-    <tr><td><?php echo _('JSON like format (deprecated)'); ?></td><td><a href="<?php echo $path; ?>input/post.json?node=1&json={power1:100,power2:200,power3:300}"><?php echo $path; ?>input/post.json?<b>node=1</b>&json={power1:100,power2:200,power3:300}</a></td></tr>
-    <tr><td><?php echo _('CSV format'); ?></td><td><a href="<?php echo $path; ?>input/post.json?node=1&csv=100,200,300"><?php echo $path; ?>input/post.json?<b>node=1</b>&csv=100,200,300</a></td></tr>
-    <tr><td><?php echo _('Assign inputs to a node group'); ?></td><td><a href="<?php echo $path; ?>input/post.json?node=1&csv=100,200,300"><?php echo $path; ?>input/post.json?<b>node=1</b>&csv=100,200,300</a></td></tr>
-    <tr><td><?php echo _('Set the input entry time manually'); ?></td><td><a href="<?php echo $path; ?>input/post.json?time=<?php echo time(); ?>&node=1&csv=100,200,300"><?php echo $path; ?>input/post.json?<b>time=<?php echo time(); ?></b>&node=1&csv=100,200,300</a></td></tr>
+    <tr><td><?php echo _('JSON format'); ?></td><td><a href="<?php echo $path; ?>input/post?node=emontx&fulljson={%22power1%22:100,%22power2%22:200,%22power3%22:300}"><?php echo $path; ?>input/post?<b>node=emontx</b>&fulljson={"power1":100,"power2":200,"power3":300}</a></td></tr>
+    <tr><td><?php echo _('JSON like format'); ?></td><td><a href="<?php echo $path; ?>input/post?node=emontx&json={power1:100,power2:200,power3:300}"><?php echo $path; ?>input/post?<b>node=emontx</b>&json={power1:100,power2:200,power3:300}</a></td></tr>
+    <tr><td><?php echo _('CSV format'); ?></td><td><a href="<?php echo $path; ?>input/post?node=mynode&csv=100,200,300"><?php echo $path; ?>input/post?<b>node=mynode</b>&csv=100,200,300</a></td></tr>
+    <tr><td><?php echo _('Set the input entry time manually'); ?></td><td><a href="<?php echo $path; ?>input/post?time=<?php echo time(); ?>&node=1&csv=100,200,300"><?php echo $path; ?>input/post?<b>time=<?php echo time(); ?></b>&node=1&csv=100,200,300</a></td></tr>
+    <tr><td><?php echo _('Node name as sub-action'); ?></td><td><a href="<?php echo $path; ?>input/post/emontx?fulljson={%22power1%22:100,%22power2%22:200,%22power3%22:300}"><?php echo $path; ?>input/post<b>/emontx</b>?fulljson={"power1":100,"power2":200,"power3":300}</a></td></tr>
 </table>
 
 <p><b><?php echo _('APIKEY'); ?></b><br>
 <?php echo _('To post data from a remote device you will need to include in the request url your write apikey. This give your device write access to your emoncms account, allowing it to post data.'); ?></p>
 <table class="table">
-    <tr><td><?php echo _('For example using the first json type request above just add the apikey to the end like this:'); ?></td><td><a href="<?php echo $path; ?>input/post.json?node=1&json={power:200}&apikey=<?php echo $user->get_apikey_write($session['userid']); ?>"><?php echo $path; ?>input/post.json?node=1&json={power:200}<b>&apikey=<?php echo $user->get_apikey_write($session['userid']); ?></b></a></td></tr>
+    <tr><td><?php echo _('For example using the first json type request above just add the apikey to the end like this:'); ?></td><td><a href="<?php echo $path; ?>input/post?node=1&json={power:200}&apikey=<?php echo $user->get_apikey_write($session['userid']); ?>"><?php echo $path; ?>input/post?node=1&json={power:200}<b>&apikey=<?php echo $user->get_apikey_write($session['userid']); ?></b></a></td></tr>
 </table>
 
 <p><b><?php echo _('Bulk data'); ?></b></p>
 <table class="table">
-<tr><td><?php echo _('You can provide data using bulk mode'); ?></td><td><a href="<?php echo $path; ?>input/bulk.json?data=[[0,16,1137],[2,17,1437,3164],[4,19,1412,3077]]"><?php echo $path; ?>input/bulk.json?data=[[0,16,1137],[2,17,1437,3164],[4,19,1412,3077]]</a></td></tr>
+<tr><td><?php echo _('You can provide data using bulk mode'); ?></td><td><a href="<?php echo $path; ?>input/bulk?data=[[0,16,1137],[2,17,1437,3164],[4,19,1412,3077]]"><?php echo $path; ?>input/bulk?data=[[0,16,1137],[2,17,1437,3164],[4,19,1412,3077]]</a></td></tr>
 </table>
 <ul>
 <li><?php echo _('The first number of each node is the time offset (see below).'); ?></li>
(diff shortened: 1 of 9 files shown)


boblemaire:
encrypted code looks good and tests OK with IotaWatt encrypted input.

chaveiro:
The concept is nice and useful.
I just ask why make encrypted a new API and not an option to the existent APIs? 
Eg.: Differentiating the transport layer (plain vs encrypted) from the data layer seemed more logic and is extensible to new API formats (message pack for example).

TrystanLea:
@chaveiro I did hope we could do that and I'm open to suggestions, the main reason I decided on the dedicated route was that the format is slightly different from input/bulk as the sentat / offset parameters are not present, its not as easy to add these parts into the encrypted section of the message.

Making this encrypted method available for other API's would have the same challenge, which parameters do you encrypt etc

TrystanLea:
I guess it could be possible to encrypt the whole post body perhaps and place the username in the url? We would then need to have a way to decode the post body into its parameters..
How would you flag up that the message being sent is an encrypted one? that was the other reason for having the input/encrypted specific API

mabi:
I think that encryption should be done at the transport layer ie https and the message should be signed with hmac to ensure authenticity

TrystanLea:
@mabi @chaveiro would you be able to give detail on how your proposed solutions would work?

- How would you signal to emoncms that the request is of the encrypted type? Is there a standard header property you would use, to identify the nature of the request?

- would you then use the username in the URL as the identifier? or would you also put this in the header?

- would you then encrypt the full POST body including parameter names e.g encrypting all of:

    data=[[123,"emontx",100,200,300]]&sentat=123

instead of 

    data=IV&CIPHER-TEXT

chaveiro:
@TrystanLea Yes, idea is to allow to secure any supported emoncms API on a new 'channel' but maintaining the same APIs formats. 
Parameters as a new field seems a nice solution. I also suggest using devicekey and deviceID instead of username. (in future we could have a way to change devices owners).

See my test example, it is basicaly a secure proxy to the existing APIs and can be  implementing on emonCMS as a module, or by invoking the controller in the end :
```

<?php

$deviceid = "1";
$devicekey = "1234";
$enc_type= 'aes-256-ctr';

/*
$time = time();
$data = array();
$data[] = array($time-30,"emontx",120.3,58.8,408.2);
$data[] = array($time-20,"emontx",120.3,58.8,408.2);
$data[] = array($time-00,"emontx",120.3,58.8,408.2);
$data = json_encode($data);
*/

$data='node=1&fulljson={"power1":100,"power2":200,"power3":300}';
$api="post"; // post or bulk

//-----------------------------------
// Encrypt data
$iv = openssl_random_pseudo_bytes(16);
$encryptedData = openssl_encrypt($data, $enc_type, hex2bin($devicekey), OPENSSL_RAW_DATA, $iv);
$base64EncryptedData = rtrim(strtr(base64_encode($encryptedData), '+/', '-_'), '=');

$base64EncryptedDataAll= $deviceid.":".$enc_type.":".$api.":".$base64EncryptedData;

// Develop a sha256 hash of the data to check against the reply
$sha256 = hash("sha256", $data, true);
$sha256base64 = str_replace(array('+','/'),array('-','_'), base64_encode($sha256));


print "<b>ENCODED PAYLOAD (POST to http://emoncms.org/input/secure.json):<br></b>";
print($base64EncryptedDataAll);

//-----------------------------------
// decrypt - This runs on the server
print("<b><br><br>Received data on the secure API, (data is obtained via \$HTTP_RAW_POST_DATA):<br></b>");
$payload=explode(":",$base64EncryptedDataAll);
$rcv_deviceID= $payload[0];
$rcv_enc_type= $payload[1];
$rcv_api= $payload[2];
$rcv_data= base64_decode(str_replace(array('-','_'),array('+','/'),$payload[3]));

$rcv_decrypted = openssl_decrypt($rcv_data, $enc_type, hex2bin($devicekey), OPENSSL_RAW_DATA, $iv);
$rcv_devicekey=$devicekey; // TODO: get $devicekey from $rcv_deviceID

print "rcv_deviceID => " . $rcv_deviceID;
print "<br>rcv_enc_type => " . $rcv_enc_type;
print "<br>rcv_api => " . $rcv_api;
print "<br>rcv_decrypted=> ";print($rcv_decrypted);


//-----------------------------------
// example, send request localy (could also invoke controller directly)
// Send request
print "<b><br><br>Example, send request localy (could invoke controller directly)<br></b>";
print 'curl --data "'.$rcv_decrypted.'" "http://emoncms.org/input/"'.$rcv_api.'".json?devicekey='.$devicekey.'"';


$ch = curl_init();
curl_setopt($ch,CURLOPT_URL,"http://emoncms.org/input/".$rcv_api.".json?devicekey=$devicekey");
curl_setopt($ch,CURLOPT_POST,1);
curl_setopt($ch,CURLOPT_POSTFIELDS,$rcv_decrypted);
curl_setopt($ch,CURLOPT_RETURNTRANSFER,true);
$result = curl_exec($ch);
curl_close($ch);

// Check that result matches sha256base64 of request data
if ($result==$sha256base64) echo "ok\n";

```

boblemaire:
@chaveiro @TrystanLea 
If the problem scope is going to be expanded to be a secure alternative to HTTPS for all of the APIs, we should probably put a lot more thought into it.  My general feeling is that the problem of inadequate resources to do HTTPS on IOT devices will go away within a couple of years, and that HTTPS and crypto will probably evolve to respond to new challenges uncovered by very sophisticated sponsored hackers.  To me this is an interim solution, and as soon as possible I'd like to get on the HTTPS bus and go wherever it goes.

This protocol is specific to one input format, bulk, but in my opinion its the most versatile.  Regardless of where a solution is implemented, anyone attempting to use it will have to do some programming.

That said, there are a couple of things here that I would like to follow up on:

Although I'm satisfied to use AES-128-CBC, I imagine that there could be an instance where that is not the best choice, or a fatal vulnerability could be discovered, so allowing the ability to specify an alternative algorithm would be useful and is pretty easy to do within the proposed format - just add another keyword in the plaintext post.

The deviceid/devicekey is another way to accomplish what username/apikey accomplishes.  I'm wondering what the rational is.  I agree that it's better to not send the username in plaintext, especially if its an email address, so using another identifier would accomplish that.  I don't know the internals, but suspect it might require another table in the Emoncms database.  The devicekey appears to be a device specific apikey.  That has some associated overhead in maintaining the keys per device and communicating them securely. In the event that the client is compromised and the key is exposed, wouldn't it be a bigger problem if it could be used to send any Emoncms API?  So I guess what I'm saying is I think the idea of a device specific apikey is a better idea if it has limited usefulness to an input only API.

All that said, do you see any vulnerabilities in this new protocol beyond exposing the username and using AES-128-CBC?

chaveiro:
Hi @boblemaire , dont get me wrong, the idea of a secure alternative to https is great for embedded systems that do not process the capacity to do https.
Making this security layer extensible with a modular implementation that allows any API format to use the added security without supporting https is great for everyone and a nice technical achievement of emoncms.
If you require a new api (but i doubt json wont fit), no problem define it and use the added security layer on top of it.

boblemaire:
no problem @chaveiro, to the extent you can convince @TrystanLea that there is a need to have a general purpose low-footprint alternative to HTTPS, I'm fine.  I do feel some complicity in having put forward this protocol and opened the door to a lot more potential work for Trystan.  

Perfect is the enemy of good.

mabi:
@TrystanLea my proposal was to sign the whole POST body using HMAC and include the hash in the Authorization header along with the userid (or another client identifier)
Authorization: <userid>:<HMAChash>
This does not completely address the https alternative that @boblemaire needed, as the message is in clear and can be read, but it prevents abuse from unauthorized clients. To identify this kind of messages the server should check if the header is present and fall back to API key if it is missing, but only if configured to do so

chaveiro:
@mabi, HMAC just ensures that the message is trustworthy but does not encrypt the data. Is that your proposal?

boblemaire:
@mabi , @chaveiro ,

I seem to be missing part of the conversation.  Is there some back channel communication that I'm not aware of?  I don't see the proposal that includes sending the post body in the clear or a statement of the problem that adding HMAC to authenticate the encrypted data solves.  These two things seem to be contradictory.

Looking past this context issue, in a general sense, how would a HMAC in the header work?  Does the sender hash the message body and encrypt that with the shared symmetric key (apikey in the pull request, devicekey in @chaveiro's proposal) or is there a private/public key to be established to do the HMAC in the typical way?

@TrystanLea and I had a conversation about server side authentication and came to the conclusion that if the server could successfully Json parse the data array using the apikey associated with the plaintext username, the message could be assumed to be valid and should be posted.  I suppose its possible that a corrupted encrypted string could decode into a valid Json array, but I have to ask why anyone would do that?  Even in that case, the SHA256 return would be invalid and the client would resend the data.

chaveiro:
Hmac is also usefull to validate non json data on the server without effort. 
Or a CRC of the decrypted payload can do the same.

TrystanLea:
Ok, I think I have an idea of how to implement this, the first step is the abstraction of access to POST and GET parameters.

I've made a start on this here:
https://github.com/emoncms/emoncms/compare/input-improvements...input-improvements-v2?expand=1

The next step is to decode the post body of the encrypted message into the $input_params array
https://github.com/emoncms/emoncms/blob/input-improvements-v2/index.php#L31

I need a way of telling emoncms that the request contains an encrypted post body, does anyone have any recommendations? is there a header property we could set that could specify the whether it is encrypted and the encryption type use perhaps?

    Authorization: <userid>:<HMAChash>
    Encryption: AES-128-CBC

TrystanLea:
@boblemaire no other back channel for communication, the above is the full discussion.

TrystanLea:
Initial implementation of encrypted post body implemented:
https://github.com/emoncms/emoncms/blob/input-improvements-v2/index.php#L140

TrystanLea:
The prototype implementation requires a userid and secure URL parameter e.g:

    http://localhost/emoncms/input/post.json?userid=1&secure=AES-128-CBC

We can set the user id in the header as discussed above. Any thoughts on the secure flag and type?

mabi:
@TrystanLea I suggest using the standard Content-Encoding: header if it is not difficult, have a look at https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Encoding and http://httpwg.org/http-extensions/draft-ietf-httpbis-encryption-encoding.html

chaveiro:
Hi @TrystanLea I've thought a little after my initial test code and think to have solved all requirements.
This can be made modular and usable for any api pretty simple not requiring a change to any emoncms file.
Non standard headers are really not needed and there are some proxies that will filter them out.

How about create a new secure api module eg "modules/secure" and we specify within the parameter of this new API what 'regular' emoncms API the message is direct to.

So there we will need2 API, one for POST other for GET, the complete request of each would be something like this :
```
POST /input/secure HTTP/1.1
Host: localhost:80
Content-Length: 1234

plain_api_path:device_id:encrypted_data:crc_of_decoded_data

```
Example for GET APIs:
```
GET /input/secure?api={}&device_id={}&data={}&crc={} HTTP/1.1
Host: localhost:80
Content-Length: 0
```

Let me know what you think.
If you agree on this approach, i could make a concept module if you could wait some time, or help you out doing it. This is the kind of a core thing that must be made right the first time, or it will bite us back later...

TrystanLea:
Thanks @chaveiro I can see the potential benefit of a separate module but at the same time I think my integrated proposal above is quite elegant, did you have a chance to look at it?. The API url is the kept the same, the only thing we need to specify is the content encoding to be aes128cbc as @mabi suggests. This tells emoncms to then look for the authorisation header and decode the iv+chipher text in the post body.

TrystanLea:
@boblemaire suggested limiting the extent to which we make this encrypted option available, given that if there are flaws in the implementation of it, it could potentially open a hole in emoncms - which is also something I am thinking about.

chaveiro:
I've missed @mabi post, now that i've read it, i like it, but only if we follow that standard strictly so we will have support by other and libraries for arduino / raspbery.
https://tools.ietf.org/html/draft-ietf-httpbis-encryption-encoding-09
That encoding allows for transferring encrypted data together with information necessary to decrypt them when somebody knows the key. 

This demonstrates the implementation :
https://github.com/tpeczek/Lib.Net.Http.EncryptedContentEncoding
https://tpeczek.com/2017/02/supporting-encrypted-content-encoding.html
https://tpeczek.com/2017/03/supporting-encrypted-content-encoding.html

The nice thing is that the decoding keyId is contained within the encrypted payload! (dont need to specify a userid/device parameter)
To downside of this standard is that it's only usable on POST/PUT, not GET parameters.

So, you are making have to make a commitment: only allow encrypt on POST, or  impair the possibility to differentiate POST from GET fields on emoncms (maybe not really a problem)...
https://github.com/emoncms/emoncms/compare/input-improvements...input-improvements-v2?expand=1


This might be usefull also for implementation details:
https://github.com/beverloo/php-ece
https://dzone.com/articles/supporting-encrypted-content-encoding-in-httpclien

TrystanLea:
Ok latest version of input-improvements-v2 uses headers: authorization and content-type.
https://github.com/emoncms/emoncms/tree/input-improvements-v2

    <?php

    $userid = 1;
    $apikey = "APIKEY";

    $data = "node=emontx&data=100,200,300";

    // Encrypt data
    $iv = openssl_random_pseudo_bytes(16);
    $encryptedData = $iv.openssl_encrypt($data, 'AES-128-CBC', hex2bin($apikey), OPENSSL_RAW_DATA, $iv);
    $base64EncryptedData = rtrim(strtr(base64_encode($encryptedData), '+/', '-_'), '=');
    $hmac = hash_hmac('sha256',$data,$apikey);

    // Send request
    $ch = curl_init();
    curl_setopt($ch,CURLOPT_URL,"http://localhost/emoncms/input/post");
    curl_setopt($ch,CURLOPT_POST,1);
    curl_setopt($ch,CURLOPT_HTTPHEADER, array(
      "Authorization: $userid:$hmac",
      "Content-Type: aes128cbc"
    ));
    curl_setopt($ch,CURLOPT_POSTFIELDS,$base64EncryptedData);
    curl_setopt($ch,CURLOPT_RETURNTRANSFER,true);
    $result = curl_exec($ch);
    curl_close($ch);

    print $result;

TrystanLea:
They key lines are L142 to L184:
https://github.com/emoncms/emoncms/blob/input-improvements-v2/index.php#L142

TrystanLea:
Thanks @chaveiro will read through the links.
I think only allowing encrypted data in post body is fine.

boblemaire:
Is there an advantage or benefit to the HMAC using the unencrypted data as opposed to the Base64encoded (raw) post data?  Seems to me that the sooner the message can be authorized, before any decoding or decrypting, the better.

boblemaire:
@TrystanLea 
I don't see what happens to the response.  Is there a mechanism in this to authenticate the response?

chaveiro:
@boblemaire that spec does specify encrypted response the same way see https://tools.ietf.org/html/draft-ietf-httpbis-encryption-encoding-09#page-7
To acknowledge correct decryption of the message is not specified by the spec, but we should do it with http error status code: 200 OK, 403 Forbidden.

chaveiro:
@TrystanLea teh header format of that spec is similar o JWE (JSON Web Encryption ) header.
 I've found some implementation is PHP that might facilitate coding, here: https://github.com/kelvinmo/simplejwt/blob/3e98a1a01a22c87d9fd7ddafd51ae14044a7b41d/src/SimpleJWT/JWE.php

TrystanLea:
@boblemaire 
"Is there an advantage or benefit to the HMAC using the unencrypted data as opposed to the Base64encoded" Im not sure about this one, happy for anyone more knowledgeable to comment.

" Is there a mechanism in this to authenticate the response?" - with the intention being to verify that the receiver was able to decode and generate the correct hash - to check for man in the middle attacks? Which the HMAC alone cannot verify.. No modified implementation does not implement this yet.. I will add this back in

Not sure where to start with the links you provided @chaveiro, could you summarise the relevant information and its implications for my v2 implementation concept above?

@mabi It would be great to have your input on my v2 implementation concept above.

TrystanLea:
I think we should limit the use of the encrypted mode to specific API's to begin with, i.e: input/post and input/bulk, so that we introduce this gradually

mabi:
@TrystanLea your v2 implementation is what I had in mind, the only vulnerability that I can see is a replay attack, but receiving exactly the same data should not cause any harm.
Including the request (current) timestamp in the hmac would stop this, but if the clocks are not synchronized this can be a problem.

TrystanLea:
Thanks @mabi ok great yes that could be a tricky one to stop.

I've abstracted the parameter loading code a bit so as to avoid overcomplicating index.php.
https://github.com/emoncms/emoncms/commit/23a6047ea59229548d3377c7e7905f790a89d74e

TrystanLea:
Actually thinking about it a user can use an encrypted body to authenticate with emoncms, non-encrypted GET parameters and request URL could then be used to write to other emoncms API end points.. It should be as secure as an API key login but its a bit messy..

The emoncms $session variable is written to in the param class here (via a global which is not ideal either): 
https://github.com/emoncms/emoncms/blob/input-improvements-v2/param.php#L79

TrystanLea:
@boblemaire I have re-introduced the sha256 response

boblemaire:
@TrystanLea 
I'll get on this tonight.  I need to make sure I can generate a compatible HMAC with the tools I have on the ESP8266.  They are not the openssl toolkit.

Regarding the HMAC using the raw (base64encoded) POST body vs the decrypted body, I brought that up because it might be simpler and more straightforward.  If the message isn't authentic, you might run into problems getting to the point where you can check the HMAC.

In the current implementation, I had changed the timestamp to be absolute time, in an effort to eliminate the need to parse the body.  If you are doing that anyway now, I suppose we should revert to the preexisting format where the time entry in each data array is relative to either a &time= or &sentat= field.  I use the &time= in the plain-text version so that works for me.

Lastly, I've been running the original secure post live on Emoncms for 10 days now.  Can you give me a heads up when you intend to change it so I can revert to the unsecure GET?

boblemaire:
@TrystanLea 

About replay attacks.  This is one of the things I had in mind when I said the protocol was flawed in the writeup, and my thinking was what @mabi just expressed.  I don't see any harm in re-posting the same data, at least not as I understand the storage engines that I use. There are implications for cumulative feeds like Kwh, but we've already looked at that with respect to data recovery and I think you only accept chronologically new data to those feeds.  

As I understand it, another common use of replay attacks is in DOS attacks.   While there is no way to stop these or any other messages from reaching the server,  the notion of somehow time stamping the transaction could help to reduce the amount of resources such a transaction uses.  As it stands, the TLS handshake for a HTTPS transaction probably consumes as much resource as receiving and decoding these encrypted transactions.  I don't know what network monitors do to detect and stop DOS attacks, but I think recognizing identical transactions should be part of it.

boblemaire:
@TrystanLea 
I have the basics working.  I can generate a compatible SHA256 HMAC and cobbled together a test php script to perform all of the essential elements.  

1) When using the hash_hmac function in php the key must be a binary string - so as in the decrypt function it must be specified as 

> hex2bin($apikey)

2) the Authorization header has a standard format, and the HTTPclient class takes over to enforce it.  I haven't worked this out completely yet.  If I change the name to something else, it works fine, so it's a special case somewhere.

3)  I changed back to the standard bulk post protocol with 

> time=.....&data=[[reltime,node,data1,data2,etc]]


I've discontinued using the old method in Emoncms.org.

TrystanLea:
Thanks @boblemaire I will make that change.

Im still trying to figure out what to do about the issue where the encrypted post body can be used to authenticate any emoncms API call.. and whether its possible with this implementation to restrict it to input/post and input/bulk in a sensible way..

boblemaire:
@TrystanLea 
I have been experimenting with the Authentication header and have hit a dead end in testing.  The problem is that the Authorization header is not passed through to my test php script.  I have checked and it is being sent by the IotaWatt along with the other headers.  No doubt the authorization is being filtered out by the server.  My hosting service is Linux/Apache.

If you are confident that all of the various server types running Emoncms will pass the Authorization header through, it will work.  Otherwise, "missing authorization header" will be the response and it will not work.  When I rename the Autorization header to something like "X-Authorization", it all works fine.

It looks as if there are system level uses for the Authorization headers, like in configuring passwords to view directories.  I also came across some pretty complex uses of it in AWS where they appear to be doing something similar to what we are doing, but with much more complex syntax and flexibility.  The difference is that they have complete control over all of the servers that run AWS.

Is there a reason to stick with the formal name "Authorization" or can we use something like "X-Authorization" to avoid this problem?

With regard to the encrypted post body providing authentication for any Emoncms call, I can't help but note the irony of expanding the protocol to be more general purpose and then discovering that it is flawed because it is too general purpose.

Not sure if this will help, but take from it what's useful:

I like the idea of using "Content-type" to specify encrypted content rather than a special URI.
I like the idea of using another header to specify the username ("Authorization" or "X-Authorization")
I like the idea of including a HMAC in the authorization.
What would happen if the URI (/input/bulk/) were included in the HMAC?
Can you restrict to just POST transactions?
When you encounter the "Content-type" as aes128cbc, can you check if its a post, and then check the URI against a list of supported actions that use this protocol? (input/bulk/) - throwing an invalid request or something if not?
That leaves the door open to use the protocol for other post transactions by adding support to the approved list, or possibly maintaining a per-site list of allowed transactions.


I'm ok to test my code when these issues get resolved.  The HMAC, encrypting and SHA256 return are working.  I still think its cleaner to HMAC the raw post body, and I can easily change that if you decide to go that way.

TrystanLea:
Thanks @boblemaire 
Just to check, is 

    $headers = apache_request_headers();

not returning 'Authorization'?

TrystanLea:
"I can't help but note the irony of expanding the protocol to be more general purpose and then discovering that it is flawed because it is too general purpose." - yes quite right!

boblemaire:
I wasn't getting the headers that way, but I changed to that method with the same result:
"Authorization" does not pass through but "X-Authorization" works fine.

chaveiro:
Told you about Non standard headers some post ago: "**some proxies that will filter them out.**" :)

I really strongly advise against reinventing the wheel here.
Some bright minds already thought about this and even created standards.
Later i've written about JWE, but how about JWT there are libs available on php and arduino already.
Take a look at this  :
http://openid.net/developers/jwt/
https://github.com/nov/jose-php
https://github.com/yutter/ArduinoJWT

And some insights:
http://odino.org/securing-your-http-api-with-javascript-object-signing-and-encryption/
https://github.com/kelvinmo/simplejwt
https://github.com/rmcdaniel/angular-codeigniter-seed/blob/master/api/application/helpers/jwt_helper.php

TrystanLea:
@boblemaire thanks for the clarification. @chaveiro my initial understanding was that 'Authorization' was a standard header? and my implementation so far is very close to what you and @mabi proposed.

If this gets much more complicated I will be inclined to move back to our original implementation of a dedicated API for the time being. Im low on time at the moment.

boblemaire:
@chaveiro 
Actually, the problem seems to be using _standard_ headers.  I'm not having any problems using non-standard headers. 

I'm still haven't burned through the summer reading list that you posted last week so it would be helpful if you could summarize how you see this JWE, JWT concept working in our environment, and explain the relative merits over what we are doing.

I spent a long time trying to understand the proposed Encrypted Content-Encoding for HTTP that you referenced last week, only to find it is a protocol for a completely different application and actually relies on TLS for underlying transaction security.

mabi:
@boblemaire @TrystanLea the Authorization header is standard (see my previous MDN link) and already used and working in emoncms as an option for the APIKEY exchange in the feed API (I just tested)

> Add the HTTP header: "Authorization: Bearer APIKEY"

boblemaire:
@mabi yes I understand that Authorization is a standard header.  My issue is that it doesn't pass through to my test pho script that I put on my iotawatt.com server.  All other headers pass to pop fine. When I rename authorization to X-Authorization, it passes through fine and works.  So not being very knowledgeable about the various server environments that emonCMS might be installed on, I am asking if this could become a problem.  I think it might be a problem if I installed it on my server and tried to use this protocol. 

As I surf the web reading about the Authorization header, I see all kinds of uses of it, for a variety of purposes.  The "Basic" variant is described as an open protocol that should be used with HTTPS, which is unavailable and the reason we are doing this in the first place.  

So no problem if you say it will work everywhere except my server, but I'm just suggesting we use a header name without so much historical baggage.

TrystanLea:
I found that Authorization was not accessible via $_SERVER, hence why I used: apache_request_headers();
I wonder if Authorization is available in $_SERVER for you @boblemaire

TrystanLea:
> UPDATE: I still think its cleaner to HMAC the raw post body, and I can easily change that if you decide to go that way. Thinking about it, authorizing the base64encrypted body upfront also validates that it transmitted correctly. It is what the sender sent. If the subsequent base64decode or decrypt fails, the error is an encoding error by the sender and can be reported as such. With the HMAC authorizing the decoded/decrypted body, any failure is ambiguous. It can be a transmission failure or the data was not properly encrypted and base64encoded.

That sounds sensible

    // HMAC generated from encrypted data
    $hmac2 = hash_hmac('sha256',$base64EncryptedData,hex2bin($apikey));
    if (!hash_equals($hmac1,$hmac2)) {echo "authorisation error"; die; }

TrystanLea:
Actually thinking more about it, its then hard to verify that the decoded data is correct? 
Im not sure what check to add without getting into complex verification to verify the output of:

    $dataString = @openssl_decrypt(substr($encryptedData,16), 'AES-128-CBC', hex2bin($apikey), OPENSSL_RAW_DATA, substr($encryptedData,0,16));

TrystanLea:
Ok I have a solution for initial restriction of this encrypted method to the input/post and input/bulk API's. Essentially there is now an array with allowed API's which is used to restrict the access to this encryption method for the time being while still making it possible for us to develop and prove the generic 'transport layer' for wider adoption at a later date

    private $allowed_apis = array("input/post","input/bulk");
    ...
    // Temporary restriction on allowed api's for encrypted method
    $allowed_apis = array_flip($this->allowed_apis);
    $api = $this->route->controller."/".$this->route->action;
    if (!isset($allowed_apis[$api])) return false; 

See lines: https://github.com/emoncms/emoncms/blob/input-improvements-v2/param.php#L49

TrystanLea:
Testing very roughly the implication of this added code, per input request

        $allowed_apis = array_flip($this->allowed_apis);
        $api = $this->route->controller."/".$this->route->action;
        if (!isset($allowed_apis[$api])) return false; 
        
        // Decode encrypted parameters
        $headers = apache_request_headers();
        
        if (isset($headers["Content-Type"]) && $headers["Content-Type"]=="aes128cbc") {

Im not measuring any noticeable slow down, I can make 1000 un-encrypted input/post requests in 2.2s with or without the encryption method code

TrystanLea:
Im generally happy now with the above implementation, if we can sort/clarify the header issue and the hmac_hash location I think we can move forward and merge this in.

boblemaire:
I'll give $_SERVER a try as soon as I can look up how to use it.  Crash course in php.

> Im generally happy now with the above implementation, if we can sort/clarify the header issue and the hmac_hash location I think we can move forward and merge this in.

Why don't you call the HMAC hash location.  I see your point with validating that the sender correctly encrypted and base64 encoded.  I can do it either way.

I can send the Authorization header as described.  My question was whether we understood what was happening on my server enough to have confidence that won't be an issue on any of the variety of servers that Emoncms runs on.  I personally prefer to use a non-standard header like X-Authorization, but I can generate and send either.  What happens after that was my question.  Once again, your call.

mabi:
@boblemaire dont't get me wrong, my comment was just to highlight the fact that that header was already successfully used in emoncms index.php so maybe it is an issue just on some servers
`$apikey = str_replace('Bearer ', '', $_SERVER["HTTP_AUTHORIZATION"]);`
obviously if it is not working for you, that is a show stopper :-)

boblemaire:
@TrystanLea  No go on the $_SERVER[HTTP_AUTHORIZATION"] (if that's how it's used).

@mabi  

> obviously if it is not working for you, that is a show stopper :-)

Not necessarily, I really don't know much about these servers other than how to upload a php script and access it.  If it looks like I know what I'm doing here, it's an illusion.  I don't run Emoncms on my server.  I'm just running a php script to validate the output from my ESP8266 IotaWatt.  In the end, that code will send input/bulk POSTs to Emoncms servers, so I want to understand that what I'm seeing won't be a problem we end up scratching our heads over.

boblemaire:
@trystan, @mabi,
Just to verify I'm not doing something stupid, can you scan this code? It works if I change the header to X-Authorization.

```
<?php

	//foreach (getallheaders() as $name => $value) {
	//	$headers[$name] = $value; }
	
	$headers = apache_request_headers();
	
	// Fetch authorization header
	if (!isset($headers["Authorization"])) {echo "missing authorization header"; die; }
	$authorization = explode(":",$headers["Authorization"]);
	if (count($authorization)!=2) {echo "authorization header format should be userid:hmac:apikey"; die; }
	$userid = $authorization[0];
	$hmac1 = $authorization[1];
	$apikey = "01d0b420e10bdbb374be371938445605";
	
	
			// Fetch encrypted data from POST body
			
	$base64EncryptedData = file_get_contents('php://input');
	if ($base64EncryptedData=="") {echo "no content in post body"; die; }
	// The base64 is converted from "URL safe" code to standard base64 (RFC2045 etc),
	// then it is decoded into the binary encrypted data
	$encryptedData = base64_decode(str_replace(array('-','_'),array('+','/'),$base64EncryptedData));
	// The binary encrypted data is decrypted using the apikey.
	// Note that the first 16 bytes of the encrypted data string are the IV and
	// the actual data follows
	$dataString = @openssl_decrypt(substr($encryptedData,16), 'AES-128-CBC', hex2bin($apikey), OPENSSL_RAW_DATA, substr($encryptedData,0,16));	
			
		// HMAC generated from decoded data
		
	$hmac2 = hash_hmac('sha256',$dataString,hex2bin($apikey));
	
	if (!hash_equals($hmac1,$hmac2)) {echo "invalid data"; die; }

		/* now we develop a sha256 hash of the data to send back in the reply as proof
			that the recipient was able to decode the message. */
	
	$sha256base64_response = str_replace(array('+','/'),array('-','_'), base64_encode(hash("sha256", $dataString, true)));
	
	echo $sha256base64_response;

		/* The Emoncms php input handler should respond 200 ok and send the sha256 string as data */

?>
```

TrystanLea:
> Why don't you call the HMAC hash location 

I vote for checking the HMAC hash location after decryption so that we can validate the decrypted output.

@boblemaire can you try to see what happens if you place this in a .htaccess file next to the test decryption script that you have? (I noticed its present in the emoncms .htaccess file)

    # ------------------------------------------------------------------------------
    # Rewrite auth header. Fix Authorization: Bearer
    # https://cweiske.de/tagebuch/php-apache-authorization.htm
    # ------------------------------------------------------------------------------
    <IfModule mod_setenvif.c>
      SetEnvIf Authorization "(.*)" HTTP_AUTHORIZATION=$1
    </IfModule>

boblemaire:
@TrystanLea Not sure if I did that right but nothing changed.  I'll stipulate that there is most likely some configuration option that will make my server work.  No problem if you say Emoncms will pass it through.  I'll just keep this in mind if the protocol appears to not work because of "missing authorization header" when going to another instance of Emoncms.

TrystanLea:
@boblemaire could you try changing your test code to:

<?php

	// Fetch authorization header
	if (!isset($_SERVER['HTTP_AUTHORIZATION'])) {echo "missing authorization header"; die; }
	$authorization = explode(":",$_SERVER['HTTP_AUTHORIZATION']);
	if (count($authorization)!=2) {echo "authorization header format should be userid:hmac:apikey"; die; }
	$userid = $authorization[0];
	$hmac1 = $authorization[1];
	$apikey = "f8e185293ed0ec7d6fec4a4ded8db767";
	
	
			// Fetch encrypted data from POST body
			
	$base64EncryptedData = file_get_contents('php://input');
	if ($base64EncryptedData=="") {echo "no content in post body"; die; }
	// The base64 is converted from "URL safe" code to standard base64 (RFC2045 etc),
	// then it is decoded into the binary encrypted data
	$encryptedData = base64_decode(str_replace(array('-','_'),array('+','/'),$base64EncryptedData));
	// The binary encrypted data is decrypted using the apikey.
	// Note that the first 16 bytes of the encrypted data string are the IV and
	// the actual data follows
	$dataString = @openssl_decrypt(substr($encryptedData,16), 'AES-128-CBC', hex2bin($apikey), OPENSSL_RAW_DATA, substr($encryptedData,0,16));	
			
		// HMAC generated from decoded data
		
	$hmac2 = hash_hmac('sha256',$dataString,hex2bin($apikey));
	
	if (!hash_equals($hmac1,$hmac2)) {echo "invalid data"; die; }

		/* now we develop a sha256 hash of the data to send back in the reply as proof
			that the recipient was able to decode the message. */
	
	$sha256base64_response = str_replace(array('+','/'),array('-','_'), base64_encode(hash("sha256", $dataString, true)));
	
	echo $sha256base64_response;

		/* The Emoncms php input handler should respond 200 ok and send the sha256 string as data */

and test alongside a .htaccess file with the following:

<IfModule mod_setenvif.c>
  SetEnvIf Authorization "(.*)" HTTP_AUTHORIZATION=$1
</IfModule>

I didnt realise @chaveiro had added that in so that $_SERVER['HTTP_AUTHORIZATION'] works. Which I realise it does now on my system.

boblemaire:
@TrystanLea 
That's what I did.  I added a file postecho.htaccess in the same directory as postecho.php.
I tried using $_SERVER["HTTP_AUTHORIZATION"] in the php.  No luck.`

```
# ------------------------------------------------------------------------------
# Rewrite auth header. Fix Authorization: Bearer
# https://cweiske.de/tagebuch/php-apache-authorization.htm
# ------------------------------------------------------------------------------
<IfModule mod_setenvif.c>
  SetEnvIf Authorization "(.*)" HTTP_AUTHORIZATION=$1
</IfModule>
```
end the file...

```
<?php

	//foreach (getallheaders() as $name => $value) {
	//	$headers[$name] = $value; }
	
	$headers = apache_request_headers();
	
	// Fetch authorization header
	if (!isset($_SERVER[HTTP_AUTHORIZATION])) {echo "missing authorization header"; die; }
	$authorization = explode(":",$_SERVER[HTTP_AUTHORIZATION]);
	if (count($authorization)!=2) {echo "authorization header format should be userid:hmac:apikey"; die; }
	$userid = $authorization[0];
	$hmac1 = $authorization[1];
	$apikey = "01d0b420c10bdbb374bef71938445605";
	
	
			// Fetch encrypted data from POST body
			
	$base64EncryptedData = file_get_contents('php://input');
	if ($base64EncryptedData=="") {echo "no content in post body"; die; }
```
etc.

boblemaire:
@TrystanLea  OK, I got it working with $_SERVER["HTTP_AUTHORIZATION"].

I was doing the .htaccess file wrong.  Still no clue about the file content, but got the correct name, in the correct directory (apparently).

It still doesn't work with $headers["AUTHORIZATION"], so it might be better if you used $_SERVER["AUTHORIZATION"].

It also looks to me as if the header name is case sensitive, so we have to be together on that.

Does this mean that others using this protocol on their servers will need to have a .htaccess file?

chaveiro:
Ok so i've made a working demo with JWE, its fully complaint with the google spec.
Uses this PHP lib: https://github.com/Spomky-Labs/jose
This is the best explication i found  on JWT, JWE JWS : https://medium.facilelogin.com/jwt-jws-and-jwe-for-not-so-dummies-b63310d201a3

The reason to use A128GCM is because it uses a shared key and of this: https://github.com/httpwg/http-extensions/blob/0f3ff05b3679f9ead35c1be800723e73915b6aaf/draft-ietf-httpbis-encryption-encoding.md

You can try the code bellow here: http://dev.gtronica.com/encrypt/jose/my/test.php
Code first encrypts a payload using a vault key, then decrypts it using the passed index key to look up the original key at the server :

```
<?php
require dirname(__FILE__) . '/../vendor/autoload.php';
use Jose\Factory\JWEFactory;
use Jose\Factory\JWKFactory;
use Jose\Loader;

echo "<pre>";

$data='node=1&fulljson={"power1":100,"power2":200,"power3":300}';
// This is on the server database
$vault = [
  '1'=> "saH0gFSP4XM_tAP_a5rU9ooHbltwLiJpL4LLLnrqQPw",
  '2'=> "aH0gsdfiha3b9857hglkgng08019324gfghpogpoiqa",
];


// We create our key object (JWK) using an octet key.
// Additional parameters ('kid' and 'use') are set for this key.
$key = JWKFactory::createFromValues([
    'kty' => 'oct',
    'k'   => $vault[1],   // the vault key
    'kid' => '1',         // the vault index of the used key
]);


$jwe = JWEFactory::createJWEToCompactJSON(
    $data,
    $key,
    [
        'enc' => 'A128GCM',
        'alg' => 'dir',
    ]
    );

echo "<br>JWE compact encoded<br>";
echo $jwe;
// The above is what the client send to the server
echo "<hr>";
echo "<hr>";

//______________________________________________________
// Lets decode a paylaod now
$payload="eyJlbmMiOiJBMTI4R0NNIiwiYWxnIjoiZGlyIiwia2lkIjoiMSJ9..3Qp3kbGsfPwxv_eW.21NZfE_naOTDmPKrtRfyXU-Xz258dZgwwqismKbnRbkMol5eg6aPpvy$

// Obtain the key from  our  vault by the given kid
$loader = new Loader();
$jwt = $loader->load($payload);
$header = $jwt->getSharedProtectedHeaders();
$decryptkey = $vault[$header['kid']];
//echo "decryptkey=".$decryptkey ;

// The payload is then decrypted using our key.
$key = JWKFactory::createFromValues([
    'kty' => 'oct',
    'k'   => $decryptkey,
]);

// Now decrypt the payload
$jws = $loader->loadAndDecryptUsingKey(
    $payload,          // The input to load and decrypt
    $key,              // The symmetric or private key
    ['dir'],           // A list of allowed key encryption algorithms
    ['A128GCM'],       // A list of allowed content encryption algorithms
    $recipient_index   // If decrypted, this variable will be set with the recipient index used to decrypt
);

echo $jws->getPayload();
//var_dump($jws);
```

chaveiro:
A quick note about the headers @boblemaire: 
 "Authorization: Bearer APIKEY" is standard but
 "Authorization: APIKEY" is not.

@TrystanLea  how about using this logic to signal a jwt encoded/encrypted version of an API, wont it seems more logical? :)
Plain API-> http://host/input/post.json
JWT API -> http://host/input/post.**jwt**

TrystanLea:
Thanks @chaveiro, this is an interesting concept, code appears pretty straight forward and I will read up on the link you provided.

I noticed it uses AutoLoader and Jose dependencies. Incorporating these if I am not mistaken would require quite a step for emoncms, there was a bit of a discussion about this in the discussion here with @inverse regarding redis encapsulation https://github.com/emoncms/emoncms/pull/536

I'm wary that we risk expanding this development issue considerably and I don’t really have time available at the moment for this. It may be that we need to think more about long term development pipeline for emoncms in regards to questions such as AutoLoader and JWE rather than do all of this now.

inverse:
@TrystanLea you could always include the dependencies pre-packaged for users and downloadable as a zip. 

Having composer manage dependencies and autoloading will revolutionise this project ;)

TrystanLea:
Thanks @inverse. Something to sit down and spend time looking at some day :) 
In the mean time, not sure how to load with composer to test (or without for that matter). Any idea on the steps @chaveiro, sorry I ran out of time this week.

TrystanLea:
I've opened another pull request from the input-improvements-v2 branch
https://github.com/emoncms/emoncms/pull/718

TrystanLea:
See new pull request with Transport layer AES 128bit Encryption feature only:
https://github.com/emoncms/emoncms/pull/722

PR #711, opened 2017-08-27
CSS improvements

franzo15:
Better visualization in mobile environment


Paul-Reed:
IMO breaking the footer into 2 lines does not look good, apart from the obvious issue in the screenshot below, it breaks up the sentence & also takes up more real-estate as the footer would be over twice as thick.

thermo

PR #712, opened 2017-08-29
fix dashboard menu overflow

thib66:
Partially fixed the dashboard menu overflow

### DIFF
--- a/Theme/basic/theme.php
+++ b/Theme/basic/theme.php
@@ -9,7 +9,7 @@
   Part of the OpenEnergyMonitor project:
   http://openenergymonitor.org
   */
-  global $ltime,$path,$fullwidth,$menucollapses,$emoncms_version,$theme,$themecolor,$favicon;
+  global $ltime,$path,$fullwidth,$menucollapses,$emoncms_version,$theme,$themecolor,$favicon,$menu;
 
 ?>
 <html>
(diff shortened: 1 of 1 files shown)


thib66:
before cd324c0 : 
image
the dashboard is truncated (missing title)
after cd324c0 : 
image
the whole dashboard is displayed

should partially fixed the issue described here : 
https://github.com/emoncms/dashboard/issues/75
and
https://community.openenergymonitor.org/t/unusable-web-interface-in-full-screen-mode/3380

chaveiro:
The real fix should avoid dual lines menu and deal with removing the legend when the menu has no space, like when you resize the window smaller.

thib66:
@chaveiro yes of course it would be nice to avoid dual line but the behaviour is managed with bootstap v2.3.2 and the following part of the code in theme.php : 
                        **_@media (min-width: 980px) and (max-width: 1200px){
                          .menu-text {
                            display: none !important;
                          }
                        }_**
and
                            **_@media (max-width: 1200px){
                              .menu-text {
                                display: none !important;
                              }_**
So the behaviour concerning the display of legend of the actives dashboard depends on two absolute values : 980px and 1200px and the number of active dashboards and /or the total length of all the active dashboards is not considerated.
So I propose two options:
- create an extra option in the settings.php in order to have the ability to hide all the legends of the active dashboards (only the star icons would be displayed) regardless of the number of active dashboards - should be easy to implement
- max-width and min-width options dynamically calculed according to the number of active dashboards and /or the total length of all the active dashboards - more difficult
@Paul-Reed @TrystanLea : Any other idea?
Regards,
Aymeric

Paul-Reed:
This has been raised many times previously and also debated in the forum. Unfortunately there has never been an consensus of how to deal with it.
I do however agree with Chaveiro that the menu should never be allowed to flow onto 2 lines, as it creates other formatting problems.

Paul

chaveiro:
Sorry late reply. Are you able to figure out how to do that option 2 (calculate max-width and min-width  dynamically)?

Other option exists, limit the number of published dashboards from the dashboards menu, at any given time to a safe number.

thib66:
Hello,
the idea is to calculate the lenght of each title in order to sum them. Depending on the total length, the max-width and min-width would be adjusted.
Regards,
Aymeric

thib66:
Hello, just for your information here is a screenshot of the previous dashboard using boostrap 3.3.7:
image

thib66:
There is an issue with the actual use of bootstrap in emoncms as we want to display published dashboards side by side whereas the concept of bootstrap is to use grids and when there is not enought space to display elements on a unique line, some are displayed below the others.
If we want to have absolutely one line, the list of published dashboards via a pulldownlist should be strongly envisaged.
Looking forward your comments,
Regards,
Aymeric

Paul-Reed:
...which is exactly what we have got already.
The concept of the shortcuts (as I recall) was to allow 1 click access to the most accessed dashboard, and the rest accessed via the drop down, but some users are adding too many, hence the problem.

dash

Paul-Reed:
What are others opinion of imposing a limit of say 3 dashboard **shortcuts**, and clipping the (displayed) dashboard names to a sensible number of digits.

Although wouldn't be popular with everyone, at least it would stop the overflow and malformation of the dashboard.
IMO the framework design of the dashboard needs to be consistent. otherwise the dashboard looks buggy, and does not reflect the hard work of it's contributors. 

Paul

TrystanLea:
If we could measure the length of the dashboard shortcuts and if less than screen width + right hand menu items, hide the shortcuts and instead move to the dropdown, that would be a nice solution, it should be achievable with jquery javascript if not in css

thib66:
Hello,
with edac540 we are able to mesure the length of the dashboard shortcuts and the other shortcuts of the menu as well. An algorithm actually computes the values for the css media property. 
Could you tell me if this algorithm gives satisfaction?
Actually, a line is displayed above the menu in order to display the number of shortcuts and the total length for each type of menu in order that you can check the algorithm is working fine : 
image

thib66:
Hello Again,
I have also verified that localization is correctly supported : 
image
image

thib66:
Hello,
Is there any volunteers to test the algorithm?
Regards,
Aymeric

Paul-Reed:
Aymeric, can we exchange email addresses?
I have a temporary email address rossoreed@gmail.com to start the exchange.
I'll delete this post shortly.

Paul

Paul-Reed:
Hi Aymeric, just trying your PR and it seems to work very well, ensuring that the menu does not overflow.
The idea of creating the info line above the menu makes testing functionality is also a nice touch, thank you.

Within the algorithm, how are you calculating space taken by the icons, and the padding between the different elements. Or are you just calculating a word count?
Obviously a lot of dashboards with short names will take more space than a few dashboards with longer names, despite their word count being the same, due to the icons & padding between them.

Could you leave the info line above the menu in place for now please, and hopefully we can get some further feedback from @chaveiro @glynhudson or @TrystanLea before progressing.

Paul

thib66:
Hello Paul,
in order to integer the space taken by the icon and the padding, in the fomula a calculation of number of items is added to the total lenght of shorcuts and also a coefficient (see parts in bold)  : 
$maxwidth1=intval((($sumlength1+$sumlength2+$sumlength3+$sumlength4+$sumlength5)+**($nbshortcuts1+$nbshortcuts2+$nbshortcuts3+$nbshortcuts4+$nbshortcuts5+1)*6**)*85/9);
$maxwidth2=intval((**$nbshortcuts1+$nbshortcuts2+$nbshortcuts3+$nbshortcuts4+$nbshortcuts5+3**)*6*75/9);
Regards,
Aymeric

Paul-Reed:
Very nice work Americ!
Sorry it's taken so long to merge.

Paul

PR #713, opened 2017-08-31
Adjust sun-theme dropdown menu colors

adminde:
Hi guys,

I noticed that the colors of the dropdown menu were still blue for the sun theme of pull request #688 and I felt the urgent need to fix that ;)

The checks will most probably fail again, as **!important** is still considered bad^^


TrystanLea:
Great thanks @adminde

PR #716, opened 2017-09-09
Update LinuxInstall.md

wthomson:
Prompt user to create the html directory if it doesn't exist.

### DIFF
--- a/docs/LinuxInstall.md
+++ b/docs/LinuxInstall.md
@@ -63,12 +63,21 @@ First cd into the /var/www directory:
 
     cd /var/www/
 
-Set the permissions of the html directory to be owned by your username:
+Change the owner of the html directory to the current user:
+
+    sudo chown $USER html
+    
+If you get an error message "No such file or directory" you'll need to create the directory,
+
+    sudo mkdir html
+    
+then change the owner:
 
     sudo chown $USER html
 
-Cd into html directory
+Change into the html directory. 
 
+    sudo mkdir html
     cd html
 
 Download emoncms using git:


Paul-Reed:
Isn't the html directory automatically created by apache2 
Under what circumstance would it not exist if the installation guide was followed?

Paul

wthomson:
I thought so too. But when I installed Apache 2.2, there was no html directory.

glynhudson:
It's always been created for me. What system did you install apache on? Is this the  latest version of apache?
 Maybe you could add a note to create the directory if it doesn't exist?

wthomson:
I *did* include a line that says:
If you get an error message "No such file or directory" you'll need to create the directory,
It's not the latest version. I did a apt-get install apache2 during a manual build and it gave me version 2.2

docs/LinuxInstall.md
```diff
@@ -63,12 +63,21 @@ First cd into the /var/www directory:
...
+    sudo mkdir html
+    
+then change the owner:
 
     sudo chown $USER html
 
-Cd into html directory
+Change into the html directory. 
 
+    sudo mkdir html
```
Paul-Reed: The html directory will have already been created either by apache or by L72.
No need to create it again.

wthomson:
Weird. It wasn't there when I did the manual build. When I went to change ownership of the file it gave me an error saying the file/directory did not exist.

PR closed.

Paul-Reed:
As Glyn said above, there is no harm in adding a note to create if it doesn't exist, but your PR contains a typo where you run `$ sudo mkdir html` twice.
Once on line 72 and again on line 80.

If you wish to amend, I'll merge your PR into the guide as a safeguard.

Paul

PR #717, opened 2017-09-09
Update LinuxInstall.md

wthomson:
Add Servername localhost to /etc/apache2/apache2.conf to prevent
"Could not reliably determine the server's fully qualified domain name, using  x.x.x.x for ServerName" error message.


Paul-Reed:
Thank you.

Paul

PR #718, opened 2017-09-13
Input improvements v2

TrystanLea:
v2 implementation of original input-improvements pull request, see full discussion here:
https://github.com/emoncms/emoncms/pull/710

- Transport layer AES 128bit Encryption, example of use documented here: https://github.com/emoncms/emoncms/blob/input-improvements-v2/docs/input_encrypted.md
- Removal of .json requirement in API
- Ability to use subaction for node name, e.g: input/post/emontx?csv=100,200,300
- New input/get/nodename api


TrystanLea:
It might be useful to note that the extent of the Transport layer AES 128bit Encryption is handled by code in param.php from Line 56 through to L101. Which is only a small part (~6%) of the changes in this pull request which also contains a lot of re-factoring of the emoncms input code and the new input/get/nodename API.

https://github.com/emoncms/emoncms/blob/input-improvements-v2/param.php#L56

TrystanLea:
Clarification: An altered version of this pull request without the Transport layer AES 128bit Encryption was merged. I will now reintroduce a branch with this feature and create a new pull request...

TrystanLea:
See new pull request with Transport layer AES 128bit Encryption feature only:
https://github.com/emoncms/emoncms/pull/722

PR #736, opened 2017-11-13
Add ability to get account name from the userid

pb66:
This is similar to the `get_username($userid)` function except it returns the My Profile name rather than the user name.
This can be used in a bespoke Theme to add a friendly name (including spaces) to the nav bar for example.

### DIFF
--- a/Modules/user/user_model.php
+++ b/Modules/user/user_model.php
@@ -440,6 +440,14 @@ class User
         return $row['username'];
     }
 
+    public function get_name($userid)
+    {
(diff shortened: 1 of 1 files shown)


pb66:
I have been using this for a while now so that I can have a friendly site name on the lefthand side of the main navbar. This has been very useful as I deploy the same dashboards for each site this client monitors and when looking at a webpage there was no indication of what user account you are looking at.

TrystanLea:
great, thanks Paul

PR #738, opened 2017-11-14
Adjust admin view to also accommodate mariadb

pb66:
Removed the "MySQL" prefix to the currently installed version as the installed mysql DB maybe mariaDB not Oracle MySQL. Also changed the section name from database to MySQL as the core database is always mysql, regardless of whether it is Oracle or MariaDB.


TrystanLea:
thanks again!

PR #739, opened 2017-11-14
Correction to swift_required.php path

pb66:
This comes up a lot on the forum. The PEAR PECL installer installs the swift_required.php file to the same location as the Swift folder not in the Swift folder, since the PEAR PECL installer is the recommended install method this is where emoncms is most likely to find it.
(for example forum mention see https://community.openenergymonitor.org/t/emoncms-org-email-reports/4519/6?u=pb66)


TrystanLea:
thanks Paul!

PR #742, opened 2017-11-16
Fix missing $userid in input redis function

adminde:
Hi guys,

this is just a very small pullrequest to fix a bug in `load_input_to_redis($inputid)`, where 
`$this->redis->sAdd("user:inputs:$userid", $row->id);` called a `$userid` object that is undefined.
This only occurred if you by chance call something extraordinary like */emoncms/input/process/get?inputid=1* immediately after flushing redis ;)

Additionally, I added a public function to get the last timevalue to the input model as I thought it cannot hurt and I possibly could like to use it someday^^

### DIFF
--- a/Modules/input/input_model.php
+++ b/Modules/input/input_model.php
@@ -348,11 +348,40 @@ class Input
                 $lastvalue = (float) $lastvalue;
             }
             return $lastvalue;
-        } else {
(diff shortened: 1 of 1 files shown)


TrystanLea:
Thanks @adminde

PR #743, opened 2017-11-16
Fix load_input_to_redis

adminde:
Well, the german word is verschlimmbessern for what happened here^^
I accidentally slipped a `fetch_array` instead of an `fetch_object` in, with #742 - breaking everything as bad as the missing userid did :)
Sorry for that.


pb66:
This fixes the "no processlist issue" thanks

PR #745, opened 2017-11-21
Break out the version to a separate file.

pb66:
Although I do not think "module.json" is the right name for this file we need consistency across all repos so that we can create a standardized update method. Ideally we need the index.php to read this file rather than have the index.php file updated at every new tag/version.

### DIFF
--- a/module.json
+++ b/module.json
@@ -1,0 +1,4 @@
+{
+    "name"         : "emonCMS",
+    "version"      : "9.8.20"
+}


pb66:
I also think this might be a good place for the `$config_file_version = "10"` settings.php file version info to live perhaps?

pb66:
I've been working on an automated updater that uses the "modul.json" files we have discussed and seen in the app and dashboard modules.

Here I propose the samething but for the core emoncms repo too. I would like to see all the files use the same name so I question whether "module.json" should be revised to "version.json" before it is too widely rolled out.

I'm happy to do the PR's including the minor edits in the admin module to make the switch.

Over the last week of so we have seen 10 revisions to the index.php due to updating the version details, a separate file will result in less commit history on the index.php file and users can make local changes (as I have) without having to rewrite those changes at every update regardless of whether there is any real change to the index.php.

Any thoughts? @TrystanLea @Paul-Reed @glynhudson @chaveiro

Paul-Reed:
Not sure what the benefits are to now calling it version.json instead of module.json, but no objections here, so long as there is consistency.
The name 'module.json' was the preference of @glynhudson so maybe his view maybe stronger than mine.

Paul

glynhudson:
Hi @pb66 ,

Good idea. I agree the version number would be better in it's own file

As to the naming conversion `module.json` vs `version.json` I have no particular preference. I guess 'version' would maybe be more intuitive when used on Emoncms core.

My original idea for `module.json` is that is could be extended in the future to include more info about each module e.g. name, description, dependencies etc. 

Could we not use `version.json` for Emoncms core and `module.json` for the modules? Maybe they could be interchangeable? I.e the same process that reads and processes the file the would honour either filename?

pb66:
DO NOT MERGE !!!

There seems to be an unexpected side effect to logging in that needs investigating.

glynhudson:
> I think that's messy and bloating any updating script to either test for all the possible file names or to make a decision based on what type of software is being updated. How would it tell?

Ok sure, let's go for switching to `version.json`. It's still quite early days and `module.json` has hardly been adopted further than the repos we have control over. As you said earlier it would not be difficult to change. 

> I am making the updating utility pretty generic so it can also be used for firmware and other software such as emonhub, perhaps it might even get considered for the emonSD/ emonPi.

Great, yes I will certainly consider it. The emonPi update is currently nice and simple but does have it's limitations. 

> The alternative I guess, is to use completely different files, I would have preferred a plain text file with just the version alone in it and would be happy to follow that route if you want a emoncms module specific file. I have worked to accommodate using json instead so we have one file but a separate txt file would be easier for me.

I think using json has it's advantages to be able to encapsulate more information. Since we already have a working solution using json let's stick with json, it's the chosen solution for many platforms e.g. [package.json](https://github.com/node-red/node-red-nodes/blob/master/package.json) for node-RED nodes.

glynhudson:
Sounds good. I have asked @TrystanLea to take a look and comment.

pb66:
I've hit a bit of a hurdle with the "2 part" emoncms modules such as emailreport and postprocess etc.

Currently these modules are installed elsewhere and a sub-folder gets symlinked from emoncms/Modules, The modules.json is located in the folder that gets symlinked which is easy for  emoncms to find for populating the admin page.

Any version file needs to be located in the root directory of the module for easy location, both locally and online at the GitHub repo. I have toyed with several solutions and find that for the version.modules data file to remain as one file there is no way around this as we cannot look for a module specific path.

I had already started installing these modules in a specific folder (eg /home/pi/emoncms_modules ) to allow the same "update all in folder" approach without unintentionally updating other git repos that happen to also be installed in the home_dir, prior to now I considered this good practice and found it a much tidier approach to just installing everything to the home dir.

With this 2nd Modules folder it would be pretty straight forward to make the admin page look in 2 locations (emoncms/Modules and emoncms_modules) for installed modules, this would allow other repo's to be included with emoncms even though they are no strictly "emoncms modules" eg emoncms/useful_scripts.

The path to the 2nd folder is already catered for in the settings.php https://github.com/emoncms/emoncms/blob/master/default.settings.php#L71-L72
```
    // For use with emoncms modules that require installation in an alternative directory to emoncms/Modules
    // $homedir = "/home/username/emoncms_modules";
```
although perhaps a change of name to something more appropriate $remotedir, $securedir, $alternativedir or $nonwwwdir etc, since it needn't even be in the home dir. 

The alternative is to turn things on their head and put the "non-www" folder inside the modules folder, install to the emoncms/Modules folder and then move/symlink it out so the version/module file is in the root and remains there for the emoncms/admin page. But I expect this is a security risk if users do not move those sub-folders, not to mention the headache of changing everything.

So the options seem to be 
1 separate version and module files
2 put the files at the root and get emoncms/admin to check 2 folders

there was a 3rd option of symlinking ./version.json to ./www_sub_folder/version.json inside the module so that the symlinked sub-folder also appears to have a copy of version.json, but IMO this is getting messy and only works on Linux.

Any objection to the admin page looking in 2 locations? I already have my installer and updater scripts doing this and there is no escaping emoncms needs a folder outside the www path.

Paul-Reed:
Rather than creating alternative locations, wouldn't it make for a cleaner solution for the emailreport and postprocess modules to be updated so that can be installed and run from within `Modules`, so that there is some commonality with the other modules?

Paul

pb66:
That would be cleaner yes, but I think these modules have been designed this way specifically to keep parts of it out of the www path so they are not exposed to the network.

I'm not suggesting adding another location, just tidying up and formalizing the fact we already have more than 1 location, to 2 specific and managed locations, and then extending the admin page's module search to include that fact.

Paul-Reed:
Yes, I get that, but to have modules/data scattered around the OS is not a good design principle, and adding an additional  function to pull things together seems like using a sticky-plaster to stop the hemorrhage. 
Node-red, by example contains all modules, data & credentials in just one directory, so making backups, restorations and problem solving so much easier.

pb66:
_"but to have modules/data scattered around the OS is not a good design principle,"_
Absolutely, I couldn't agree more, hence my suggestion to round up all those that cannot go in emoncms/Modules to one (other) folder. I'm not informed enough to challenge the need for 2 locations, just that IF they cannot all reside in one folder, it should be restricted to 2 and both should be looked at by the admin page.

I'm all for moving towards a single location but I don't think that's going to fly, even if possible, there will be little or no appetite to change what we have. In the meantime...

I can most likely implement a change to search in 2 places myself, where as restructuring several modules and ensure security is beyond me. I do not see this a "sticking plaster" as it is simply encompassing the 2 types of module (network exposed and non-network exposed), if in time they can be merged, great we'll only have one location. but there have been modules that cannot be installed to emoncms/Modules for many years so I do not expect that to change overnight, but who knows?

Paul-Reed:
_I'm all for moving towards a single location but I don't think that's going to fly, even if possible, there will be little or no appetite to change what we have. In the meantime..._
@glynhudson @TrystanLea your thoughts please.

Paul

Paul-Reed:
Should this 'standard' folder also include the emoncms data files too, as well as the settings files?

This would make the backup/restore process easier, and in many respects replicate the model used by node-red.
Node-red runs a test at start-up, to find out if the settings file exist in `/home/pi/.node-red` and if it does, then node-red uses those settings & paths. If not, it uses a default settings file.
Emoncms could run a similar test, and also include the path for emonpi's.

/home/pi/.emoncms
/home/pi/.emoncms/Modules/ _(or should that be /home/pi/.emoncms/**m**odules/)_
/home/pi/.emoncms/settings/
/home/pi/.emoncms/data/

Paul

borpin:
How would it manage non-standard installations?

Paul-Reed:
_How would it manage non-standard installations?_

In what way, can you give an example?

borpin:
DietPi does not use ```/home/pi```. Windows installations?

Not suggesting it won't, but I do worry that there is a risk of being sucked into everything revolving around the RaspberryPi and Rasbian.  I do think it is a good idea though.

As there is no 'install' script for Emoncms (unlike Node-Red), this folder path cannot be set at install time.

_it would fall back..._

I'd suggest the opposite in that, instead of falling back to the install folder, the data directory location is set within the install folder (could include it in the administration page).  There is of course already some path data within the ```settings.php``` file.

TrystanLea:
Happy with original pull request here and the use of the version.json file name. Can we separate out the wider question about module locations for now? and place the version.json in the public_html folder for the emailreport and postprocessing modules.

Then come back to the wider question in our discussion here https://community.openenergymonitor.org/t/move-data-settings-file-to-a-common-location/6464/19

TrystanLea:
or should we come back to this pull request after sorting out the latter?

pb66:
_"and place the version.json in the public_html folder for the emailreport and postprocessing modules."_

I am trying to write a generic updater that is not specific to emoncms modules, for what I am trying to achieve the version MUST be in the root of the folder/module, we can symlink it later, but the updater needs to find a version in the root of the on-line repo to establish if an update is available, I'm not happy with the path changing for different modules or repos, many repos are not even going to have a "public_html" folder, that is generic to certain emoncms modules (usefulscripts doesn't have one) and none of the firmware repos or the emonpi or the emonhub repos will have "public_html" folders.

Serving "PHP" from a "public_**html**" folder isn't entirely clear and I can see that getting changed along the road, we could end up with a string of places to check for the version file and that not only adds extra code and extra processing etc, it also adds confusion and opens the door to issues arising from multiple version files. 

Having the version at https://github.com/<user-or-organisation>/<repository>/blob/<branch>/version for all repos makes for a very reliable and simple system.

TrystanLea:
I realise this is very similar to the pull request I have just merged from @emrysr.

Can you share your automated updater @pb66 that might help with understanding what your trying to do?

PR #762, opened 2017-12-24
Fixed tooltip lookup for stacked charts and a few Codacy tweaks

CDuffers:
Lines 269-274 are the real change, the rest is swapping single for double quotes.


glynhudson:
👍

PR #772, opened 2018-01-10
Extended menu length to prevent confusion like …

CDuffers:
As per the confusion cause here https://community.openenergymonitor.org/t/backup-option-not-available/4698
And also because the scroll bar is not appropriate for such a small menu, I've extended the menu length.

Before:
image

After:
image


TrystanLea:
thanks!

PR #775, opened 2018-01-12
copy emoncms info in the clipboard

thib66:
image

### DIFF
--- a/Modules/admin/admin_main_view.php
+++ b/Modules/admin/admin_main_view.php
@@ -169,7 +169,6 @@ pre {
 table tr td.buttons { text-align: right;}
 table tr td.subinfo { border-color:transparent;}
 </style>
-
 <h2><?php echo _('Administration'); ?></h2>
 <table class="table table-hover">
     <tr>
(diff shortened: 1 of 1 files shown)


Paul-Reed:
Should that be 'Copy to clipboard', instead of 'Copy in clipboard'?

Paul

thib66:
Hello Paul, 
I've just changed the label.
Thanks for pointing the error.
Regards

Paul-Reed:
Thanks Aymeric, It would be great if we could encourage users to paste this info in the forum, when describing issues.
Maybe other items of system info could also be included, such as OS, Server, MYSQL version etc

Paul

thib66:
Hello Paul,
I will add a 'Copy to clipboard' button for each category (OS, Server, MYSQL version etc)
Aymeric

Paul-Reed:
But that would mean that users would have to keep switching windows between emoncms & the forum to copy/paste the information.
I was thinking more of just one button, which would quickly & easily copy 'key' system information, and allow them to then paste it in forum posts etc.

I merged your first commit as a proof of concept, to stimulate some comments from other developers/users, as to what information is required, and will raise a forum post later today.
Can we see what forum feedback we get before further development please.

Paul

Paul-Reed:
Forum post added - https://community.openenergymonitor.org/t/copy-system-info-to-clipboard/6232 for comments.

Paul

PR #776, opened 2018-01-13
Update to phpmqtt_input to accept JSON object

borpin:
**Not done much PHP work so all comments gratefully received**

Modified phpmqtt_input to accept a validated JSON name:value object. Used the InputMethods class to actually inject the data.

Updated some of the indenting as it did not line up (helped me work out what was what).

.gitnore updated to ignore VSCode folder.

TODO - look for a 'time' name in the input data, put it into $time and remove from JSON object. (help with this appreciated from an expert in PHP - I'll work it out eventually though!).

TODO - update comments / help

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ emoncms.log
 .idea
 .vagrant/
 .env
+.vscode/*
 
 # Ignore all modules except core ones
 Modules/*
--- a/scripts/phpmqtt_input.php
+++ b/scripts/phpmqtt_input.php
@@ -194,8 +194,22 @@
                 // JSON is valid - is it an array
                 $jsoninput = true;
                 $log->info("MQTT Valid JSON found ".$value);
-                #TODO If JSON check to see if there is a time value else set to time now.
-                $time = time();
+
+                #If JSON check to see if there is a time value else set to time now.
+                if (array_key_exists('time',$jsondata)){
+                    $time = $jsondata['time'];
+                    if (is_string($time)){
+                        if (($timestamp = strtotime($time)) === false) {
+                            $log->warn("Time string not valid ".$time);
+                        } else {
+                            $time = $timestamp;
+                        }
+                    } else {
+                        //Do nothings as it has been assigned to $time as a value
+                    }
+                } else {
+                    $time = time();
+                }
             } else {
                 $jsoninput = false;
                 $time = time();
--- a/scripts/phpmqtt_input.php
+++ b/scripts/phpmqtt_input.php
@@ -98,6 +98,9 @@
         $device = new Device($mysqli,$redis);
     }
     
+    require_once "Modules/input/input_methods.php";
+    $inputMethods = new InputMethods($mysqli,$redis,$user->get_timezone($mqttsettings['userid']),$input,$feed,$process,$device);
+
     $mqtt_client = new Mosquitto\Client();
     
     $connected = false;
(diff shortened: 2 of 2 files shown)


TrystanLea:
Thanks @borpin I will look over your code and test here in the coming week

borpin:
Latest commit relates to issue #777 and allows a 'time' value to be used when processing the JSON message.

Currently not deleted from the input data.

TrystanLea:
Sorry for the delay @borpin. I've had a good look through the pull request now, there was a lot in there to take in at once. I see your using the 'process_node' function rather than passing the data to the in-built processing stage, do you think you would be able to use the in-built processing stage instead?

TrystanLea:
Is it worth also having a discussion re accepted formats? I know @pb66 bought this up in the past. So far we support 

    basetopic/emontx/power 10
    basetopic/emontx 100,200,300

your suggestion adds (if I read it correctly?):

    basetopic/emontx {"power1":100,"power2":200,"power3":300}

and the option to set a time:

    basetopic/emontx {"time":UNIXTIME, "power1":100, "power2":200, "power3":300}

Should we also support?

    basetopic {"nodeid":"emontx", "time":UNIXTIME, "power1":100, "power2":200, "power3":300}

and even:

    basetopic [
        {"nodeid":"emontx", "time":UNIXTIME, "power1":100, "power2":200, "power3":300}.
        {"nodeid":"emontx", "time":UNIXTIME, "power1":100, "power2":200, "power3":300}
    ]

borpin:
To be more accurate rather than it being emontx it is

`basetopic/nodeid {"time":UNIXTIME, "power1":100, "power2":200, "power3":300}`

so I do not see any benefit of concatenating the data further as this reduces the MQTT connections significantly and it is rare that more than one node reports at exactly the same time I'd suggest.  It is also usually true that a node will report all the data elements in one go (although I currently have 2 x CO2 sensors reporting separately and independently into the same node which breaks the paradigm of node = HW device).

> I see your using the 'process_node' function

That was because I simply used the same method the HTTP API employed 😄 

https://github.com/emoncms/emoncms/blob/a82d54fc5805dc3b5a31248d8526e2786f2f772f/Modules/input/input_methods.php#L101 

Before the PR is merged, I will

- remove the case sensitivity of 'time' key
- add an INFO message re the time actually used so it should be more obvious if the setting of the time fails.

TrystanLea:
I've created a development branch for these changes called mqtt_input_json and have merged in your changes there. I've got a couple of modifications I would like to make and I think this is probably the easiest way to do that as I could not seem to send a pull request to your fork
https://github.com/emoncms/emoncms/commit/87f76690f6779cb0bf8af30d093736b3b932eda1

TrystanLea:
ok this is probably a bit hard to follow, the git compare is not particularly clear. This is what I meant by using the inbuilt processing rather than input methods
https://github.com/emoncms/emoncms/pull/788/files
It results in less lines added overall and avoids having two processing routes

borpin:
> It results in less lines added overall and avoids having two processing routes

Achieves the same end. 

> I could not seem to send a pull request to your fork

I'd need to give you access I think.

I was going to make some minor changes to the logging around setting the time so it was clear what had happened but I now can't work out how to!  May be too much info for the log although there is no other way to be sure the time set has been taken.

```
                    } else {
                        //Do nothings as it has been assigned to $time as a value
                        $log->info("Valid time string used".$time);
                    }
                } else {
                    $log->info("No time element found in JSON - System time used");
                    $time = time();
                }
            } else {
                $jsoninput = false;
                $log->info("No JSON found - System time used");
                $time = time();
            }
```

borpin:
I added the log bits in via another PR #789 but I suspect that may have really confused things!

borpin:
I have added in making the search for the time key case insensitive.

I also realised that the case where there is a time string but it is not valid did not result in a time value being set.

TrystanLea:
"Achieves the same end." yes certainly, I've merged my suggested changes in https://github.com/emoncms/emoncms/pull/788

borpin:
Changes commited via #788 so closed

PR #779, opened 2018-01-18
copy to clipboard server and client information - step 2

thib66:
Add 2 buttons in order to copy information related to server and client configuration

### DIFF
--- a/Modules/admin/admin_main_view.php
+++ b/Modules/admin/admin_main_view.php
@@ -214,6 +214,7 @@ if(is_writable($log_filename)) {
                     <br>
                     <button id="getlog" type="button" class="btn btn-info" data-toggle="button" aria-pressed="false" autocomplete="off"><?php echo _('Auto refresh'); ?></button>
                     <a href="<?php echo $path; ?>admin/downloadlog" class="btn btn-info"><?php echo _('Download Log'); ?></a>
+                    <button class="btn btn-info" id="copylogfile" type="button"><?php echo _('Copy to clipboard'); ?></button>
 <?php } ?>
                 </td>
             </tr>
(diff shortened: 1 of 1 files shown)


thib66:
image
image
client information has been added in order to retrieve the version of the browser, the screen resolution, ...
the "details" tag has also been added
I cannot test the MQTT and Redis part . Is there a volunteer to test this in order to indicate if the render is good?

Paul-Reed:
Yes of course, I'll volunteer.
Do we need separate buttons to copy the information? It would mean that users have to make repeated actions to copy/paste the information.
I was anticipating that a single button would copy all of the relevant information, and the `<summary>` tags would be inserted to split up the various sections into Server Information, Client Information, etc.

Paul

Paul-Reed:
I've updated https://community.openenergymonitor.org/t/copy-system-info-to-clipboard/6232/19 with the result.

Paul

thib66:
Hello, 
if you click a Copy to Clipboard button, the data is copied with <summary>, <details>,  <br /> (should be used for the forum)
if you ctrl + click  a Copy to Clipboard button, the "raw" data is copied (only carriage return)
Server and client information are now merged in the copy.
Aymeric

Paul-Reed:
Thanks Aymeric, it's looking much better now, and nice to be able to paste the log in much the same way.
I've updated the forum post, but not merged - at least until after the [repository has been fixed.](https://community.openenergymonitor.org/t/updating-system-libraries-used-by-emoncms/6274/38)

Paul

Paul-Reed:
Just looking through the code Aymeric, there's an awful lot of code duplication to produce the information. Is there no way that the existing code (which is used to layout the page 'Server Information') can be reused to prepare and format the data that is copied to the clipboard?
That would probably also address [this comment by borpin](https://community.openenergymonitor.org/t/copy-system-info-to-clipboard/6232/23)

Paul

thib66:
Hello,
Yes there is a duplication of the code and this is wanted and this is necessary.
Look at the copy / paste of the existing code made by borpin and you will see there are truncations.
Regards,
Aymeric

thib66:
Paul look also to the code of pb66 sent 4 days ago : 
image

thib66:
Paul, another thing.
if you reduce the browser window size you get this for the post of pb66 : 
image
all the data are truncated. This is not acceptable. 
the behaviour could be explained by the fact table tags are used whereas boostrap is probably loaded by the forum page and the table are not correctly managed.
Regards,
Aymeric

pb66:
@Paul, Can we get a feature branch set up for this so we can try it and perhaps explore some alternatives.

I wasn't aware the html was getting truncated, but that isn't just a forum issue, if you reduce a browser window down beyond a certain point the same happens and I've also noticed the lines are incomplete when the page is narrow enough to wrap the text as the text goes to the edge of the page, under the scroll bar.

If the code is to be duplicated and there is a raw text function, then we should explore a better way of displaying a forum friendly version. There seems little point in having both a raw text and an un-formatted text if there is little difference between them and a screenshot is clearer.

Perhaps we need to look at fixing the html table so it can be made portable or look at other ways of formatting the data for pasting in the forum. Lets not allow the desire to have collapsible elements or retain the pipe separators dictate whether the data is easily read on the forum, that has to be the primary goal or it's pointless pursuing.

Paul-Reed:
`copy2clipboard` feature branch now added, and PR merged.

Paul

PR #786, opened 2018-02-03
Add options for changing settings file location

Paul-Reed:
Enables the settings.php file to be moved outside of HTML route.
See [this quick video.](http://www.youtube.com/embed/ksIDDfz7igM?rel=0&autoplay=1)
This PR accomodates the following paths, but others could be very easily added for different operating systems, such as DietPi, etc;

- home/pi/.emoncms/
- home/pi/
- or compatibility fallback of emoncms installation directory

Main advantages are;
- Makes it easy to import/export as most things are in one directory
- Moves settings file out of html path
- Makes it easy to delete emoncms & reinstall afresh from git
- Backwards compatible

### DIFF
--- a/process_settings.php
+++ b/process_settings.php
@@ -14,11 +14,28 @@ defined('EMONCMS_EXEC') or die('Restricted access');
 
 require_once('Lib/enum.php');
 
-// Check if settings.php file exists
-if(file_exists(dirname(__FILE__)."/settings.php"))
+// Locate settings.php file location
+// RaspberryPi
+        if(file_exists("/home/pi/.emoncms/settings.php"))
+        {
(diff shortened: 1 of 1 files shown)


Paul-Reed:
Submitted for discussion at this stage.
If this is the direction that we would want to pursue, I would suggest that I update the raspberry pi installation guide to reflect this, and possibly create a small bash script (in /scripts) to create the new directory structure, set ownership, and copy across default.settings.php to the new structure (as settings.php). This would cut several steps out of the installation guide.

Paul

pb66:
Hi Paul - I've replied on the forum thread https://community.openenergymonitor.org/t/move-data-settings-file-to-a-common-location/6464?u=pb66

PR #790, opened 2018-02-08
Wrong variable used

borpin:


borpin:
I am sure I am doing something wrong here but this is the only way I could find of submitting this change.

PR #792, opened 2018-02-08
make time key case insensitive and add logging

borpin:
Make the search for a time key case insensitive and add in additional logging.

If time key has a string value, if that string is not valid use system time.


borpin:
@TrystanLea I think I have got this right at last (in terms of actually creating the PR etc).

TrystanLea:
great looks good @borpin thanks a lot

PR #793, opened 2018-02-09
Multigraph localization

thib66:
the result with French localization : 
sans titre 2
sans titre 4


thib66:
image

thib66:
image

Paul-Reed:
Aymeric, what is the purpose of vis_langjs_builder.php in scripts please.

Paul

thib66:
Paul, vis_langjs_builder.php is needed to activate the localization done on multigraph_edit.js
Regards,
Aymeric

thib66:
vis_langjs_builder.php is used to complete vis_langjs.php. Please look at the comments inside the vis_langjs.php file.

Paul-Reed:
Thanks Aymeric. I know what the `vis_langjs_builder.php` does, I was just surprised to see it in the scripts folder, as it's not used by 'users', and only used in development, I assumed it would be placed in the [useful scripts repo](https://github.com/emoncms/usefulscripts) along with other administrative tools.
However... I later noticed that we already have the `dashboard_langjs_builder.php` in the emoncms scripts folder, so yes, I can see why you've placed it there.
The development team have had a discussion recently about where scripts should be located, so it may at some stage be moved, but for now, we'll leave it there.

Thanks

Paul

Issue #791, opened 2018-02-08
"Error creating user" at the first register on a fresh installtion


belba:
If I follow the installation instruction for Ubuntu (https://github.com/emoncms/emoncms/blob/master/docs/LinuxInstall.md) I meet the point where I have to make the first reistration for the admin-user.

At this point I got only the "Error creating user" message. No further hints in any logfile I could found (apache, mysql,syslog).

I played also with $stmt->errno and $stmt->error statements around line 225 in Modules/user/user_model.php - with no any success.

The problem is also in discuss at the community forum from Opennergymonitor:
https://community.openenergymonitor.org/t/emoncms-error-creating-user/6490

any ideas?

belba:
@geoffrey2312 
please check if in the table "users" the column "tags" has default null. If not try following SQL command:

ALTER TABLE `users` CHANGE `tags` `tags` TEXT CHARACTER SET utf8 COLLATE utf8_general_ci NULL DEFAULT NULL;

More abot this issu you can find under 
[https://community.openenergymonitor.org/t/emoncms-error-creating-user/6490](url)

geoffrey2312:
Huzzah - Worked perfectly thank you.

MatthiasMattiza:
Having the same problem. worked perfect! Thanks a lot.

pb66:
The problem should have been fixed by PR #796 and this issue closed then, can you please confirm you have had the issue with code cloned/downloaded/updated in the last 4 days?

If yes please provide more details on your setup as we will need to look at this again.

If no I think we can close this issue.

geoffrey2312:
I had this issue on Monday I spun up an Amazon EC2 with Ubuntu server and had thus issue.

TrystanLea:
Im not 100% sure that this is still a fixed issue..

If I change the user_schema from:

    'tags' => array('type' => 'text', 'default'=>NULL),

to:

    'tags' => array('type' => 'text', 'default'=>NULL, 'Null'=>true),

and recreate the tags column it seems to work again on my laptop test system. It does work fine however on an emonpi, without this change

TrystanLea:
I've created a pull request with my proposed changes https://github.com/emoncms/emoncms/pull/852

glynhudson:
@geoffrey2312 @pb66 could you check that PR #852 fixes this issue for you? Should it be merged?

takkaria:
I'm still experiencing this issue.  Looks like a bug in dbschemasetup.php.

[...]

PR #796, opened 2018-02-12
user registration fails if no default for 'tags' is set

rostwald:
If 'tags' has no default value, the sql query for user registration will
silently fail.

Also see this thread regarding this issue:
https://community.openenergymonitor.org/t/emoncms-error-creating-user/6490/19
Thanks to @belba who found the culprit of the error.


pb66:
Fixes #791

PR #800, opened 2018-02-15
Input process list - tidy reordering arrows

CDuffers:
Changed alignment of the first down arrow within the process reorder list within "emoncms/input/view" when editing the input process list so that it will align to the right hand column and thereby make it easier to follow.

Also the hover over text for the down arrow has been corrected from "up" to "down", example of both:

image


Paul-Reed:
Thanks Clive.

PR #802, opened 2018-02-15
fix visibility of progress bar with French language

thib66:
Before this PR, 
there is an issue with progress bar when language is set to French in the user profile (and probably for other languages): 
sans titre 2
the bar is not visible due to a wrong formatting of the width value below : 
sans titre 4
this PR fix this issue

### DIFF
--- a/Modules/admin/admin_main_view.php
+++ b/Modules/admin/admin_main_view.php
@@ -315,13 +315,13 @@ if ( @exec('ifconfig | grep b8:27:eb:') ) {
 if ($system['mem_info']) {
               $sysRamUsed = $system['mem_info']['MemTotal'] - $system['mem_info']['MemFree'] - $system['mem_info']['Buffers'] - $system['mem_info']['Cached'];
               $sysRamPercent = sprintf('%.2f',($sysRamUsed / $system['mem_info']['MemTotal']) * 100);
-              echo "<tr><td><b>Memory</b></td><td>RAM</td><td><div class='progress progress-info' style='margin-bottom: 0;'><div class='bar' style='width: ".$sysRamPercent."%;'>Used:&nbsp;".$sysRamPercent."%&nbsp;</div></div>";
+              echo "<tr><td><b>Memory</b></td><td>RAM</td><td><div class='progress progress-info' style='margin-bottom: 0;'><div class='bar' style='width: ".str_replace(',','.',$sysRamPercent)."%;'>Used:&nbsp;".$sysRamPercent."%&nbsp;</div></div>";
               echo "<b>Total:</b> ".formatSize($system['mem_info']['MemTotal'])."<b> Used:</b> ".formatSize($sysRamUsed)."<b> Free:</b> ".formatSize($system['mem_info']['MemTotal'] - $sysRamUsed)."</td></tr>\n";
 
               if ($system['mem_info']['SwapTotal'] > 0) {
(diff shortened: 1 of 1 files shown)


Paul-Reed:
Aah right... because many languages use a comma as a decimal separator instead of a dot, which is upsetting the display of the bar.
I assume that is your understanding too Aymeric?

Paul

thib66:
Hello Paul,
yes, that is my understanding.
Aymeric

Paul-Reed:
I'm not at home to look myself right now, but Is it not possible to obtain the value **before** it is translated, so that there would be no need to apply this.
Aren't we obtaining a value, translating it, and then hacking it back to usable value again.

Paul-Reed:
I'm not at home to look myself right now, but Is it not possible to obtain the value **before** it is translated, so that there would be no need to apply this.
Aren't we obtaining a value, translating it, and then hacking it back to a usable value again.

@AndyTaylorTweet - was this feature your addition?

AndyTaylorTweet:
@Paul-Reed Yes this is code I committed a long time back, thib66 is correct that it is floored in that localisation is able to replace the decimal point with a comma - that really is a problem, however I don't believe that string substitution is actually the best cure.

I believe (but I need to test it) that it would be better to pre-format the variable into the correct format instead;

`number_format($theFloat, 2, '.', ''); //swap the output to use the correct decimal place character.`

The other option that is not so nice it to enforce the use of GB number formatting, thats fine for US/GB but other parts of the world do not follow the same standards and may well not appreciate the enforced use here.

`setlocale(LC_NUMERIC, 'en_GB'); //override the decimal separator`

So I would suggest replacing line 317 for example;
`$sysRamPercent = sprintf('%.2f',($sysRamUsed / $system['mem_info']['MemTotal']) * 100);`

With:
`$sysRamPercent = number_format(round(($sysRamUsed / $system['mem_info']['MemTotal']) * 100, 2), 2, '.', '');`

That should do the math, round the output to 2 decimal places, and then format the output to use "." and not "," and enforce the 2 decimal places on the output (so that 10.1 is outputted as 10.10).

One issue I had with this, is that I can't persuade my machine to replicate the initial issue, setting my Pi to Fr localisation and my machine / browser etc all still rendered correctly - any other suggestions to help me replicate to the issue would be great so that I can't test the above PoC before blindly submitting a pull.

@thib66 are you able to test the above to see if that cures your issue ?

thib66:
@ynyreds Hello,
I would prefer to keep the French format for $sysRamPercent except when this variable is used in the width property.
regards,
Aymeric

thib66:
@ynyreds disregard my message. my apologies

AndyTaylorTweet:
@thib66 ahh you want the cherry and the bun :)
Paul doesn't like the string substitution because its not always a match - and I can't disagree with that.

We could add another variable used only for table generation;
To save doing the math more than once, we also need to stage out the creation;

`$sysRamPercentRaw = ($sysRamUsed / $system['mem_info']['MemTotal']) * 100;
$sysRamPercent = sprintf('%.2f',$sysRamPercentRaw);
$sysRamPercentTable = number_format(round($sysRamPercentRaw, 2), 2, '.', '');`

Now you get the RAW calculated output that can be re-used without doing the math again, plus the formatted output for use in the data displayed to the user, plus the unformatted output for use to generate the table.

Does that work better for you ?

thib66:
Hello,
thanks for your support.  I have good results with dbee0cd

Paul-Reed:
Thanks Aymeric & Andy.

Issue #777, opened 2018-01-13
Non Numeric date strings not processed correctly if older date/time


borpin:
It appears to me that Emoncms currently looks like it processes the 'time' passed in non numeric string format when it does not if that date/time is not 'now'.

If a non numeric time value is passed as a string, that is close to time now, the data gets processed implying the date/time field has been accepted.  If the time is set to less than now (e.g. 10s), the data is not processed and the input is shown as n/a. No idea why a string close to 'now' gets through; those are my observations.

If a time in seconds is used, that is less than now, it is processed correctly.

This behavior has been tested on the Input API and was observed as part of the modifications to the phpmqtt_input work I am doing.

I have put a test into the phpmqtt_input to check and convert a string date, but this should be more deeply embedded.  I attempted to do this, but it seems the integer casting of time objects in other places may be causing problems later on (a guess I admit).

[From PHP docuemtation](http://us2.php.net/strtotime)
````php
if (is_string($time)){
    if (($timestamp = strtotime($time)) === false) {
        $log->warn("Time string not valid ".$time);
    } else {
        $time = $timestamp;
    }
}
````

TrystanLea:
Hello Brian, Im not sure that I follow, where does a non numeric time values processed with strtotime come in to emoncms previous to your pull request?

borpin:
Hi Trystan, I picked up from this thread https://community.openenergymonitor.org/t/uri-not-acceptable/5928/6 that there was a perception that the API accepted an ISO format time string.  This appears to be confirmed by the API not rejecting a string value if it is 'now' but anything older does get rejected (input simply does not appear).

If the API is not designed to accept a string (and looking at the code this is doubtful), it should at least not accept a string or produce a warning.

Looking at this again, I think bulk could be left as it is, but changing the main API is I think quite simple
I think they key line is this one, and doing something similar to the above code should work.
https://github.com/emoncms/emoncms/blob/c400a14e319a154c8bb8901640ad7ff9f4032ca7/Modules/input/input_methods.php#L56

However, I don't know enough about the overall code structure to be sure.

HTH, Brian

TrystanLea:
Im not entirely sure why a date time string that is equivalent to the present moment worked ok, I would have thought all strings will convert to 0 by the integer type casting.

> If the API is not designed to accept a string (and looking at the code this is doubtful), it should at least not accept a string or produce a warning.

I think I would prefer this rather than accepting a new format.

TrystanLea:
[...]

PR #803, opened 2018-02-16
Update time validation for API and MQTT input

borpin:
Following on from issue #777 I have added in some validation of the time input value for both the API (less bulk input) and MQTT input.

This allows for the time be in ISO string, seconds and milliseconds.  It also checks the time is valid such that it is in the past and not future (if there is a TimeZone issue for instance).

I updated the validation I had previously done in the MQTT_input as I realised it was not robust.

Both sets of code are identical so could be included as a function somewhere (which is where my php skills end I'm afraid).

### DIFF
--- a/Modules/input/input_methods.php
+++ b/Modules/input/input_methods.php
@@ -39,7 +39,7 @@ class InputMethods
     public function post($userid)
     {   
         // Nodeid
-        global $route,$param;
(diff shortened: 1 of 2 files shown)


borpin:
I think I have tried to be too clever and added in a check to see if the time is in the future. Not doing what I hoped so need to revisit this.

PR #804, opened 2018-02-17
API time input format and validation modifications

borpin:
Re issue #777 
**phpmqtt_input.php**
Updated validation of time input to MQTT to make more robust. Check for a numeric value first before a string value.

**input_methods**
Used same time validation method in input_methods so both input methods (API & MQTT) allow the same time formats and JSON data structure. Check for a time parameter element first. In decoding JSON data, if no time parameter set, look for one in JSON data.

Updated API help page.

I have not been able to test any impact to bulk upload method. I do not think there will be any.


borpin:
The commit ce609d0 occurred because I update my repository from upstream.  It does not seem to have impacted the actual PR.

borpin:
I'll close this and re issue as it has all got too messy for me with Git

PR #805, opened 2018-02-18
Added aggregation functions (DMY and average) to FEED ENGINE:MYSQL ...

isfh-ohrdes:
…to be compatible to FEED ENGINE:PHPFINA and other Modules (e.g. graph, app)


glynhudson:
Appologies for the slow response. Any thoughts on this @TrystanLea ?

TrystanLea:
Looks great @isfh-ohrdes sorry for the slow merge

PR #807, opened 2018-02-19
remove ntp, ufw from package list; clean up sudo-code

rostwald:
ntp and ufw aren't related to the functionality of emoncms; i.e. they are not dependencies. NTP already was removed from the Raspberry Pi installation instructions with 6eddfdd1ee6111607a5bbb4d14502dc8c8951381

General cleanup of sudo-code blocks to make them more readable. Removed unnecessary linking of the site configuration as this is handled by `a2enmod`

### DIFF
--- a/docs/LinuxInstall.md
+++ b/docs/LinuxInstall.md
@@ -10,11 +10,11 @@ You may need to start by updating the system repositories
 
 **For Ubuntu 14.04:**
 
-    sudo apt-get install apache2 mysql-server mysql-client php5 libapache2-mod-php5 php5-mysql php5-curl php-pear php5-dev php5-mcrypt php5-json git-core redis-server build-essential ufw ntp -y
+    sudo apt-get install apache2 mysql-server mysql-client php5 libapache2-mod-php5 php5-mysql php5-curl php-pear php5-dev php5-mcrypt php5-json git-core redis-server build-essential -y
 
 **For Ubuntu 16.04:**
 
-    sudo apt-get install apache2 mysql-server mysql-client php libapache2-mod-php php-mysql php-curl php-pear php-dev php-mcrypt php-json git-core redis-server build-essential ufw ntp -y
+    sudo apt-get install apache2 mysql-server mysql-client php libapache2-mod-php php-mysql php-curl php-pear php-dev php-mcrypt php-json git-core redis-server build-essential -y
 
 ### Install PHP pecl dependencies
 
--- a/docs/LinuxInstall.md
+++ b/docs/LinuxInstall.md
@@ -38,18 +38,21 @@ You may need to start by updating the system repositories
 Emoncms uses a front controller to route requests, modrewrite needs to be configured:
 
 ```
- sudo a2enmod rewrite
- sudo sh -c "echo '<Directory /var/www/html/emoncms>' >> /etc/apache2/sites-available/emoncms.conf"
- sudo sh -c "echo '  Options FollowSymLinks' >> /etc/apache2/sites-available/emoncms.conf"
- sudo sh -c "echo '  AllowOverride All' >> /etc/apache2/sites-available/emoncms.conf"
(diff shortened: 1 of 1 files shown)


Paul-Reed:
The install guide will be used by users with a wide range of programming experience, so most of the guides have been written so that commands are simply copied & pasted line by line, into the cmd prompt.
I'm wondering if less experienced users would be confused by lines 43-51??

Paul

borpin:
Couple of things.

I agree with Paul that the original instructions, as listed, are easier for a new user to understand and that is important.

Secondly, the Linux based instructions have a number of flavours in the documentation, and they *must* all be consistent IMHO - a lack of consistency has been a flaw in the past.  Consistency makes keeping the different flavours up to date easier.

Simply removing ufw and ntp is fine as they are not strictly dependencies, but if the user base is to grow, it is important to cater for the lowest denominator and that is a new user.  There is still a place for some basic help with time sync and firewall to explain their importance, but perhaps these should be broken out to a separate document.

pb66:
"I'm wondering if less experienced users would be confused by lines 43-51??"

I actually think heredocs are the right way to do the task, I use them in my installer script, but ( as I have [just said on the forum](https://community.openenergymonitor.org/t/install-dependencies/6530/42?u=pb66
)) I do not think we should encourage users to use `sudo su`.

```
cat <<EOF >> /etc/apache2/sites-available/emoncms.conf
 <Directory /var/www/html/emoncms>
	Options FollowSymLinks
	AllowOverride All
	DirectoryIndex index.php
	Order allow,deny
	Allow from all
 </Directory>
EOF
```
is IMO much cleared than 
```
sudo sh -c "echo '<Directory /var/www/html/emoncms>' >> /etc/apache2/sites-available/emoncms.conf"
sudo sh -c "echo '  Options FollowSymLinks' >> /etc/apache2/sites-available/emoncms.conf"
sudo sh -c "echo '  AllowOverride All' >> /etc/apache2/sites-available/emoncms.conf"
sudo sh -c "echo '  DirectoryIndex index.php' >> /etc/apache2/sites-available/emoncms.conf"
sudo sh -c "echo '  Order allow,deny' >> /etc/apache2/sites-available/emoncms.conf"
sudo sh -c "echo '  Allow from all' >> /etc/apache2/sites-available/emoncms.conf"
sudo sh -c "echo '</Directory>' >> /etc/apache2/sites-available/emoncms.conf"
```

as you can c;early see the block of text you are transplanting.

But the content is also looking to be changed as we decide the location of the emoncms folder (not in "html") and we improve the vhost conf, there is no reason to (and in fact we shouldn't ) edit the apache2.conf file. The servername can be defined in the vhost, editing the apache.conf can cause headaches when serving multiple sites, and it can block (or get overwritten by) updates to that file via the package manager, if we can keep it as stock (which we can), we should.

"remove manual linking of the site configuration from site-available to
site-enabled; that's the job of 'a2ensite' (see a2ensite manual page)"

Definitely, I had missed that both were present, I think that is a result of a merging separate guides perhaps as using `a2ensite` has never really been fully embraced (no matter how much I bang on about it :-) ) until recently the links were always done manually.

TrystanLea:
> I actually think heredocs are the right way to do the task

Yes that does look much cleaner. I agree about not encouraging sudo su, Im sure that's not recommended practice?

rostwald:
I've reverted the usage of `sudo su` as suggested; see added commit #0a9e6cb

Paul-Reed:
I've just made some minor changes in https://github.com/emoncms/emoncms/pull/807/commits/d0a0cbde7bcbbf3ce1804464d5c2c4450e2ab916 as using tabs to indent the heredoc commands causes unexpected results when pasted.

heredoc

I realize that the heredoc operator can be changed to accommodate tabs, but spaces have been found to be more reliable.

Paul

Paul-Reed:
Merged, thank you.
@borpin Your comments have been noted, and hope that we can move towards greater consistency across the install guides, to make it easier for all.

Paul

PR #808, opened 2018-02-20
Fix test for emonpi

Paul-Reed:
Fixes https://community.openenergymonitor.org/t/error-after-update-to-v9-8-28/6629/2
I'm sure there is a more elegant way of writing this fix!
Also, it's untested on emonpi (as I don't have one!)

Paul


glynhudson:
Just tested on an emonPi and it works. Happy to merge when you are

glynhudson:
Thanks 👍

PR #820, opened 2018-04-07
Added Index as another option in schema

cagabi:
I am working in a project in which it's not possible to have a primary key in one of the tables so I wanted to index one of the columns.
This is an example of how the schema file of a module may look like:

```
$schema['distribution_distributions'] = array(
    'date' => array('type' => 'date', 'Null'=>false, 'Index'=> true),
    'itemid' => array('type' => 'int(11)'),
    'quantity' => array('type' => 'int(11)'),
    'distribution_point_id' => array('type' => 'int(11)'),
);


TrystanLea:
Thanks @cagabi, sorry for the slow update on merging this, I now see that there is a conflict, Im not sure what the best way is to merge this part and retain the error response, I wonder if you could check? @emrysr any ideas?

<<<<<<< HEAD
            if ($query && $apply && !$mysqli->query($query)) $operations['error'] = $mysqli->error;
=======
            if ($query && $apply) {
                $mysqli->multi_query($query);
                while($mysqli->more_results()){ // We need to loop all the results to avoid errors in future single queries
                    $mysqli->next_result();
                }
            }
>>>>>>> ac7f0254bf3ea8c7165fff58dda1e074c4315da6

cagabi:
I am looking into it

cagabi:
Sorted now, I have checked by manually corrupting the query statement and the error is displayed in the screen

PR #821, opened 2018-04-09
Modify time validation for API and MQTT input

borpin:
Fix for issue #777 and a replacement PR for #804.

Validate the time input value for both the API (less bulk input) and MQTT input, to allow time to be in ISO string and seconds.

Updated validation of time input to MQTT to make more robust. Used same method in input_mthods so both input methods allow same time format.

Updated API help page.

Not tested any impact to bulk upload method. Should be none.


glynhudson:
Can this be merged @TrystanLea ?

TrystanLea:
Thankyou @borpin

PR #822, opened 2018-04-09
Prevent MQTT Client exception

borpin:
Issue - subscribing to a topic when the MQTT client was not connected caused an exception; not fatal but not nice.  This has been noticed on a number of topics recently on the Forum.

[Documentation](http://mosquitto-php.readthedocs.io/en/latest/client.html)

The onConnect callback is only called if there is a CONNACK from the MQTT server.

The onConnect callback assumed than any call to it meant a successful connection.  This is not correct, only a response of '0' means success, so $connected now set as such.

The reconnect try, did not check the connected flag before attempting to subscribe to the topic.  By doing so the exception can be avoided.

### DIFF
--- a/scripts/phpmqtt_input.php
+++ b/scripts/phpmqtt_input.php
@@ -123,15 +123,17 @@
             try {
                 $mqtt_client->setCredentials($mqtt_server['user'],$mqtt_server['password']);
                 $mqtt_client->connect($mqtt_server['host'], $mqtt_server['port'], 5);
-                $topic = $mqtt_server['basetopic']."/#";
-                //echo "Subscribing to: ".$topic."\n";
-                $log->info("Subscribing to: ".$topic);
-                $mqtt_client->subscribe($topic,2);
+                // check and see if the onConnect callback was successful.
+                if ($connected) {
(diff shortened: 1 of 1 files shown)


TrystanLea:
Hello @borpin just tested this here and it doesnt seem to work for me. If I restart mqtt_input it does not detect any inputs?

borpin:
Odd. What do you see in the logs?

borpin:
Yes, it was because the callback takes a little while so the code never got back into the subscription code.  Split the connection and subscription up and added a variable that will take the topic ID on successful subscription and 0 if not subscribed.

I have also added in #823 code to create a persistent connection.

Without this, it can be seen that the ID returned on subsequent subscriptions (forced by restarting the mqtt_input daemon) increases.  With it, the ID is the same.  Using node-red, and using a QOS of 2, retained = true, data that was in the MQTT buffer, comes across (at least one value).

A loss of connection to the MQTT broker has been tested by stopping the MQTT broker (on a remote machine).

TrystanLea:
Hello @borpin I've uploaded a version of this that fixes the conflict to test here:
https://github.com/emoncms/emoncms/pull/822

I will close this PR

PR #832, opened 2018-04-30
fixed get_name() in user model

cagabi:
The get_name () method was trying to fetch the wrong column from the table. EMonCMS doesn't use this method anywhere but we use it in MyHomeEnergyPlanner, that's why I spotted it


glynhudson:
Thanks 👍

PR #839, opened 2018-05-13
decrease keepalive sensitivity to reduce disconnections

pb66:
See https://community.openenergymonitor.org/t/a-question-over-whether-emoncms-is-compatible-with-mqtt/7180/12?u=pb66

I suspect setting a keepalive of "5" is actually resulting in MORE disconnections than a more relaxed setting of "60". But it may be better to go for "180" or more to accomadate setups with just a couple of emonTH's.

**PLEASE DON'T MERGE WITHOUT DISCUSSION AND CONSIDERATION OF THE CLIENT ID PR TOO!**

### DIFF
--- a/scripts/phpmqtt_input.php
+++ b/scripts/phpmqtt_input.php
@@ -122,7 +122,7 @@
             $last_retry = time();
             try {
                 $mqtt_client->setCredentials($mqtt_server['user'],$mqtt_server['password']);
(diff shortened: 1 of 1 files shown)


glynhudson:
This also sounds great. Shall I merge into master to open up the testing pool? All (or the majority) of emonPi/emonBase/emonSD users are running Stable. I can't see it causing any adverse issues

borpin:
Is working for me with the keepalive removed completely so it falls back to the default.

PR #841, opened 2018-05-16
Use device view by default

adminde:
Hi guys,

as usual, I hope this is not too arbitrary. This small PR contains 3 "minor" propositions, starting with me taking a liking in the new device view, following a disgracing amount of time it took me to understand how to enable it again via the settings 😄 I'd propose here to opt-out the new view instead of an opt-in.

Further, I noticed the auth check failing for non-redis installations as a {"success": false ... } gets returned and checked against "no devices", which results in a ever-present "allow" button.
At last, I'd propose a json response to fulljson bulk requests.

Hope this is no spark for difficult discussions and thanks for your work as always^^


TrystanLea:
Great idea @adminde I will come back to this once device-integration is merged in the device module :)

TrystanLea:
Thanks @adminde

PR #846, opened 2018-05-21
Localization vis in dashboard

thib66:
Without PR : 
image
With PR : 
image


thib66:
The dialog box configure item is now localized with options coming from widgets of dashboard module and options coming from visualisations of vis module

thib66:
Should be applied with #158 of emoncms/dashboard repository

glynhudson:
Thanks :+1:

PR #848, opened 2018-05-24
Added max_value and min_value processes

cagabi:
Addition of 2 new processess to limit the value of an input before logging it to a feed. User can define the maximum and/or minimum values allowed


TrystanLea:
Great thanks @cagabi

PR #849, opened 2018-05-24
Hooking system

cagabi:
In [this forum post](https://community.openenergymonitor.org/t/emoncms-org-terms-of-use-and-privacy-policy/7437/2?u=cagabi) it came up the need of a way to tell not core modules to delete user's data when the user deletes his/her account. 

The original implementation of "delete user" looks for the userid in handcoded tables but that will not work for other modules not included so far. In my case I need to implement this functionality for the task module. So the solution is a hooking system (this pull request :). An implementation of a hook can be found in [the group module](https://github.com/emoncms/group/blob/master/group_hooks.php#L20)

### DIFF
--- a/Modules/user/deleteuser.php
+++ b/Modules/user/deleteuser.php
@@ -46,7 +46,7 @@ function delete_user($userid,$mode) {
             if ($mode=="permanentdelete") $redis->del("writeapikey:$apikey");
         }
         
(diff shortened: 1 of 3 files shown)


cagabi:
@TrystanLea beware that I have removed the handcoded tables as I am planning to write hooks for MHEP but there is one table _mhep_library_access_ which you use in the _library_dev_ branches but is not used in _master_. Maybe I should put it back into the delete user function??

TrystanLea:
Thanks @cagabi looks great, happy for the handcoded tables to be removed if your adding them in to the modules directly.

cagabi:
Great, I'll work today in the delete user hook for MHEP and do it for the library_dev branch too

PR #850, opened 2018-05-30
Fix regexp for retrive mosquitto 1.5 version number

gablau:
Fix regexp for retrive mosquitto 1.5 version number


glynhudson:
Thanks 👍

PR #853, opened 2018-06-05
Feature/issue 840 migrate to mosquitto

emrysr:
installed mqtt following this guide:
https://github.com/emoncms/emoncms/blob/master/docs/RaspberryPi/MQTT.md


Modules/process/process_processlist.php
```diff
@@ -686,13 +692,30 @@ public function wh_accumulator($feedid, $time, $value)
...
+            $this->mqtt->onConnect(function($responseCode, $message) use ($log, $topic, $value) {
+                /*
+                0 	Success
+                1 	Connection refused (unacceptable protocol version)
+                2 	Connection refused (identifier rejected)
+                3 	Connection refused (broker unavailable )
+                */
+                //log errors connecting to mqtt
+                if ($responseCode > 0) $log->info($msg.$message);
+                $this->mqtt->publish($topic, $value);
```
borpin: Should the publish be inside an else i.e. only publish if there is a successful connection?  I suspect it will raise an exception if you try a publish with no connection.

Equally, if there is no connection, you should not disconnect.

I think, that if the connection fails the first time, the data will be lost as it will not retry.

emrysr: Hi Borpin, thanks for looking over this.
We've moved to the [Mosquitto MQTT](https://github.com/mgdm/Mosquitto-PHP) client to make use of the callbacks and event-driven operations used in the library to avoid the issues you raised. Installed via PECL. 

` pecl install Mosquitto-alpha

I've called the Mosquitto::publish() method within the onConnect() callback. This will only attempt the publish once a successful connection is established.

The Mosquitto::loopForever() function will attempt re-connection if disconnected for whatever reason. And the onConnect() callback will disconnect once the messages have been published.

If there is an unsuccessful connection an error is sent to the error log.

I hope this helps. Let me know if you think I've missed anything.

Thanks again for your help - I had to double check that it was doing what is was supposed to be doing :)

borpin: Hi @emrysr,

> We've moved to the Mosquitto MQTT

Yes, I suggested we did so.

> I've called the Mosquitto::publish() method within the onConnect() callback. This will only attempt the publish once a successful connection is established.

The onConnect() callback is called even if there is an unsuccessful connection (hence values other than 0).  The Mosquitto::publish() should only therefore be called if a CONNACK of 0 is received.

If this happens then as you say the Mosquitto::loopForever() function should reattempt the connection (does this need to time out?).

I suggest that if you want to disconnect on a sucessful publish, use the onPublish callback.

What is way beyond me, in PHP terms, is, once a connection has been made, does there need to be a disconnect() or could it simply be left to run.  Making multiple connections is daft if you do not need to but I really do not know if the asynchronous nature of this actually work across separate procedural calls.

I've got #822 open to modify the mqttinput script as there was a similar issue in that the subscribe was happening before there was a connection and causing an exception.  However, I have subsequently found (but not had time to do anything with) that a simpler mechanism (to flags etc) is to subscribe in the onConnect() where CONNACK is 0.

emrysr: Apologies @borpin I should have looked you up before I sent you the last message - I'm new!! :-\

Good spot! I've just changed it to only publish() where CONNACK is 0. else error logged.

Firing the disconnect() event within the publish() event would work fine in most cases. However in the few cases where it doesn't the loopForever() would cause the script to hang.

The error log should provide an accurate description of what needs to be fixed to allow the script to connect successfully.

https://github.com/emoncms/emoncms/commit/98565cfe903328b91d8aae6bb6681fa722c6c246

let me know if you think I've missed something. Thanks.

Issue #760, opened 2017-12-21
Feed CSV export interval


cagabi:
According to the description in the modal:

_Selecting an interval shorter than the feed interval (or Auto) will use the feed interval instead._

But selecting "Auto"  sets the interval to 1s not to the feed interval.

Once we are there, it would be good to include all the possible intervals when setting up a PHPFINA log to feed process: (5s, 10s, *15s, 20s,* 30s, 60s, 2m, 5m, 10m, 15m, 30m, 1h, 1d) as interval options in the feed export in the drop down(missing 15s and 30s)

emrysr:
Added the above intervals and set the 'auto' option to 10s

https://github.com/emoncms/emoncms/pull/858

PR #858, opened 2018-06-06
added 2 additional  intervals and altered auto to 10s

emrysr:


chaveiro:
Please let auto be 1 second or it will break existent reports based on sql engine.

TrystanLea:
Hello @chaveiro how do the existent reports work, why is auto = 1s needed?

pb66:
@TrystanLea  an you please confirm the intended functionality, the "Progress thread" you recently updated on the forum states _"Feed CSV export interval auto should be minimum of 10s"_ will 5s feeds created before the 10s minimum was rolled out, still be able to be downloaded in full?

The quoted documentation in #760 

 _"Selecting an interval shorter than the feed interval (or Auto) will use the feed interval instead."_

sounds like the correct way to do things rather than an auto minimum of 10s regardless of feed selected, especially as many users may not know the interval off-hand.

Don't forget this is the standard emoncms repo (not just emoncms.org) and the emonPi data is currently posted every 5secs and the 5sec fixed interval feeds are still available to self hosting, so you are setting the auto minimum to double the standard emonSD interval at 10s.

chaveiro:
Hi, i remember doing that code. The auto was created specialy for mysql like feeds where you dont have an underlaying interval. It will impact that feeds as they can have an arbitrary interval with the minimum being 1 sec (integers for the timestamp).  By chosing an export interval other than 1 you get a rouded number to the interval you set if i remember correctly and that is not the intended behaviour.
Maybe the fix here should be on the phpfina feeds engine export code?

emrysr:
#863 - I've re-inserted the 1s interval as part of another fix:
https://github.com/emoncms/emoncms/pull/863

The auto/default is still set to 10s but you can select 1s if required.

image

chaveiro:
Hi, if a proper fix that works for all feeds engines cant be provided, i'd say to just remove the auto from the list as it will be missleading on mysql exports.

PR #859, opened 2018-06-06
added °C to feedlist units

emrysr:


pb66:
Hi Emrys, Will it be possible to use free text units as well as (or instead of) selecting from a predefined list?

Just to name a few more litres, cubic meters, gallons, £, $ (and many other currencies, both minor and major units) £/ph (plus other currencies, units and periods) ppm, mph, kph, mm/day (plus other increments and periods) mV, mA, VA, Mb, Gb (and other units of disk/ram usage) pulses events dB Hz seconds millis etc etc just to name a few off the top of my head.

PR #867, opened 2018-06-08
added unique tag:name check in feed->set()

emrysr:
checked for unique tag:name in the controller.


TrystanLea:
Diolch @emrysr, testing here, I cant change the tag to a tag that already exists where the name is different. It should look for the combination of tag:name rather than name and tag separately. E.g the following are valid, even though the tag is repeated and there are other feeds with the same name. 

    emontx:power1
    emontx:power2

    emonpi:power1
    emonpi:power2

emrysr:
I've fixed the check for unique tag:name combinations

PR #869, opened 2018-06-10
Improved "Pi" section in the administration panel

gablau:
Improved the "pi" section in the administration panel:
- added exact Model, through revision codes. (https://www.raspberrypi.org/documentation/hardware/raspberrypi/revision-codes/README.md)
- added SoC and serial number, through /proc/cpuinfo
- added GPU temperature, requires user "www-data" in the "video" group, otherwise it will not be shown (cmd "sudo usermod -G video www-data")
- added current filesystem status ro / rw, via "mount" command

Preview of changes made see the red dots:
emoncms_admin_mod_show

### DIFF
--- a/Modules/admin/admin_main_view.php
+++ b/Modules/admin/admin_main_view.php
@@ -340,6 +340,7 @@ if ( @exec('ifconfig | grep b8:27:eb:') ) {
               if(strpos($gputmp, 'temp=' ) !== false ){
                 $gputmp = " - GPU: ".str_replace("temp=","", $gputmp);
               }
+              else $gputmp = " - GPU: n/a"."(to show gpu temp you have to launch this command from the console \"sudo usermod -G video www-data\" )";
                echo "<tr><td class=\"subinfo\"></td><td>Temperature</td><td>CPU: ".$cputmp.$gputmp.RebootBtn()."</td></tr>\n";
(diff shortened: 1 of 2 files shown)


glynhudson:
Nice work! This is really great. Displaying the exact Pi version will be very useful for debugging other users systems and especailly as we start to move to the new RasPi 3B+

I first tested on a 'stock' emonSD raspi image, would it be possible to add a more verbous and or tidy message to be displayed when the GPU temperature cannot be displayed? Maybe the command `sudo usermod -G video www-data` could be displayed to promp the user? 

screenshot 2018-06-12 at 00 04 44


After setting up the www-data permissions (`"sudo usermod -G video www-data`) and restarting Apache reading the GPU temperature worked great: 

screenshot 2018-06-12 at 00 10 55

gablau:
I fixed the problem, but I did not try it.
See if it's ok, otherwise you can correct the message to give to the user.

Regarding the buttons to enable / disable writing the file system, since we know the status of the file system, how about if we hide the button that is not needed?

glynhudson:
Thanks a lot, that's great 

>Regarding the buttons to enable / disable writing the file system, since we know the status of the file system, how about if we hide the button that is not needed?

Yes, great idea. It's best if you open a new PR for this. 

Cheers,

PR #879, opened 2018-06-14
Update readme.md

wthomson:
Move the paragraph that describes the difference between emoncms.org and self-hosted emonCMS to the top of the description.
With the move to paid subscriptions on emoncms.org, there has been an increase in questions (and general confusion) about the differences between the two, or even the fact the two ARE different.


Paul-Reed:
That looks much better Bill

glynhudson:
Looks good, thanks 👍

PR #880, opened 2018-06-15
File system ro / rw buttons

gablau:
Enable the file system ro / rw buttons according to its status


TrystanLea:
Hello @gablau looks like this pull request is combined with the translation concept, any chance to separate it out? thanks

gablau:
Hi Trystan, 
very strange :-) 
close this and see PR #897

PR #881, opened 2018-06-15
Feature/issue 687 editing process list type

### LINKED ISSUE #687: Possible improvement to editing process list items.
When (re?)editing a process type for an item in a process list the arg resets, ideally this should only occur when the arg type changes ie a different process type is selected. 

Currently if you wish to change a `+feed` to a `-feed` for example, you also need to reselect the feed from the feed list as it will automatically reset back to the first feed in your list.

emrysr:
fixes #687


TrystanLea:
gret, diolch @emrysr

PR #885, opened 2018-06-18
vis options optimization

thib66:
Before : 
image
labels truncated
image
color picker not correctly displayed on FireFox
After : 
image
image


TrystanLea:
Thanks @thib66 works great!

PR #890, opened 2018-06-20
[Proposal] Removal of the gettext requirement - DO NOT MERGE

gablau:
Hi,
This is a proposal to remove the requirement of gettext, make the translation system leaner and faster and easier to use.

Advantages:
- no complicated installation of gettext, generation of locale, etc.
- structure of the translation folders is simplified
emoncms_dir_language
- translation cache in redis
emoncms_redis_language
- simplification of the code writing of the views with the new functions `<?php echo _('My account'); ?>` becomes `<?php _e('My account'); ?>`

In this PR you can see the translation sample of the User module, in addition to the changes to the few main files

If you think it is a good solution and you would like to implement it, I will be happy to help you.

Let me know.
Gabriele


TrystanLea:
Thanks @gablau, its a nice idea to simplify translation support. For some reason I cant test your code now? could you try again?

gablau:
Hi @TrystanLea ,
of course, I forked the emoncms repository, then applied my patch again.
Tested and working on my RPI with emonSD 26Oct17
Find everything here: https://github.com/gablau/emoncms

Let me know if you need a new PR.
Gabriele

reedy:
If this is moved forward, I'd highly recommend not adding the lib like this, but infact bring it in use composer

https://github.com/php-gettext/Languages#with-composer

Will make the diff a lot cleaner

TrystanLea:
Yes definitely keen not to include the library in the repo if we can. I still need to get up to speed with using composer. If its not too much work for either of you to rework this pull request using composer that would be great.

reedy:
> Yes definitely keen not to include the library in the repo if we can. I still need to get up to speed with using composer. If its not too much work for either of you to rework this pull request using composer that would be great.

Rebasing it it is probably the most difficult part.

The only problem we kinda have atm is when/where to load the composer libraries. If a project has an "autoloader", it's easier, but when you're using PHP files for each different file (nothing wrong with that), there's little common code.

Basically, you just need to be calling `require __DIR__ . '/vendor/autoload.php';` in the "right" places

PR #891, opened 2018-06-21
correct db update URL in docs

mabi:
The upgrade instructions explain how to update the DB even when not being able to login. The URL was not correct, fixed by this commit.

### DIFF
--- a/docs/Upgrading.md
+++ b/docs/Upgrading.md
@@ -76,7 +76,7 @@ Add the following line to the bottom of *settings.php* to enable a special datab
     
 In your internet browser open the admin/view page and click on the database update and check button to launch the database update script.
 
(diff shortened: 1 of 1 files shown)


pb66:
I think that whole passage needs some clarification. The line immediately preceding that url says 

> In your internet browser open the admin/view page and click on the database update and check button to launch the database update script.

so 'http://localhost/emoncms/admin/view' would be correct to open the admin/view page.

the admin/db url should be added rather than replace admin/view link and overall, perhaps the "what to do if you are unable to log in" part should come after the "how to do it normally" has completed rather than in the middle of it.

mabi:
I agree, I'll update the doc as per your suggestion

glynhudson:
Thanks :+1:

Issue #515, opened 2016-05-03
Detect version of emoncms from API


JumpMaster:
To keep track of changes to the API it should be possible to query the version of emoncms from the API.  This should either be returned with every request, have a dedicated request, or a known request such as feed list should return the version.

chaveiro:
A specific api for emoncms version info is recommended.
But i'd like APIS to not be versioned and changes to APIs ensure backward compatibility.
Meaning for example if you pass a new parameter to an old api that did't suported it at that time, the response just be empty or would not toke usage of that parameter.
What is your usecase?

JumpMaster:
Hi Chaverio,

I would like to use a newer API that was recently implemented but currently have no way of knowing if it is safe to use. If the Android app could get an emoncms version or API version I could use that to decide which API calls are compatiable.

Kev

chaveiro:
What newer API ?

JumpMaster:
For example /feed/data.json now has the mode option instead of using interval.  

e.g.
https://emoncms.org/feed/data.json?apikey=APIKEY&id=3&start=START&end=END&interval=86400&skipmissing=1&limitinterval=1

can be replaced with

https://emoncms.org/feed/data.json?apikey=APIKEY&id=3&start=START&end=END&mode=daily

TrystanLea:
This is now implemented https://github.com/emoncms/emoncms/pull/895

PR #893, opened 2018-06-25
Feature/issue 515 showing version

emrysr:
You can now get the version by accessing /version   or by getting the file /version.txt

related to #515


glynhudson:
Thanks, this works well. How about dropping the date? What do you think @TrystanLea ? The version number links to the github emoncms releases page where each release is date stamped. This would improve the mobile view since currently on emonPi the footer overflows onto two lines. 

Also do you think it's still needed to display "low write" if we're running redis? I guess this is useful but not really part of the version number. The admin page already shows if Redis is running or not

glynhudson:
I've just made the change to remove the data and have pushed to `version-number` branch and have created a new PR: https://github.com/emoncms/emoncms/pull/895. I will close this one. 

I don't think it's possible for me to push commits to your branch?

PR #895, opened 2018-06-25
Version number file & remove date

glynhudson:
- Breakout version number to a file @emrysr
- Remove date from version number, the version number still links to the github release page where each version has changelog and date stamp. 

Related to https://github.com/emoncms/emoncms/pull/893


chaveiro:
I'm advocate of release dates paired with the version numbers.
Why not keep the release date as additional metadata of the file?

pb66:
Breaking out the version to a separate file is a welcomed change. 

I agree the date is superfluous since any version relates to a specific release date, it's is simply not needed to show that info at the foot of every page, the link back to github expands on the simple version number with a date and additional notes, commits and changelog etc.

The simplicity of the version file content is IMO important (see https://github.com/emoncms/emoncms/pull/745) although the previous trend was towards a combined module.json/version.json file, I strongly favor a separate plain file containing a raw version number only  and that is what I currently use in my emoncms updater scripts.

Personally I would like to see the new file name reduced to just "version" not "version.txt" so that in addition to using the new version api to get the currently installed version, we can also compare it to the latest **available** version by simply comparing the results of http://myserver.com/emoncms/version to the results of https://raw.githubusercontent.com/emoncms/emoncms/version-number/version.txt (swapping out "version-number" for "stable" or "master" or pointing the url at your personal repo etc), a extension-less "version" file would offer an endpoint  consistent to the emoncms "version" api.

With regards to "low-write" in the footer, This is a different thing to versions and dates, it is a mode not a version and whilst I agree it may not be needed (as per the version date), it does help a great deal when users post screen shots on the forum, perhaps a slight formatting difference could replace the words low-write.

also with ref to #893. "low-write" doesn't directly indicate Redis. You can have Redis running without it being a low-write install, low-write refers to "buffered-write" not Redis, but info about both Redis and Buffered-write (low-write) are available on the admin page.

Paul-Reed:
I don't disagree with having a version file, but shouldn't there be some consistency with the modules 'module.json' file format?
ie, all called version.txt or module.json.

IMO we can lose the date - serves no real purpose.

glynhudson:
My thinking was that getting rid of the date would make the footer fit on a single line when viewed on a mobile device. The date of the release is a single click away. Clicking the version number links to the github releases page. 

screenshot_20180625-230658

TrystanLea:
Consistency with module.json is fine with me, e.g:

{
    "name"         : "Emoncms",
    "version"      : "9.8.1"
}

Im aware we are adding two potential overheads though, one loading a version file on every page load and two parsing json, I would likely turn the feature off on emoncms.org but for emonpi installs its unlikely to be of any significance.

pb66:
As much as I would like to see a single file suit both functions, it feels like we're trying too hard to make that work.

A version file **MUST** be in the root of the repository to make querying the latest available version consistent across all repo's (not just emoncms or emoncms modules but ALL repos), the module.json is currently located in the root of the folder that gets symlinked into the www path and it's name varies from module to module. Updater and install scripts for Linux based installs (not just emoncms.org and emonPi's, but the other users too) will most likely be bash which doesn't naively support JSON, and anything other than version number is just noise, parsing is therefore an unnecessary complication, we know it's "emoncms" because we are looking in the emoncms folder or the emoncms repo for the version file. Likewise with the modules.json file, it's in the apps folder so we know it's the apps module.

IMO the version file should be in the repo root and just contain the raw Semantic version (major.minor.patch) nothing else. That _can_  then be symlinked into the www path so it _could_ be used for both jobs _if_ we wanted, but if you really want the module.json to be located anywhere but the root of the repo (eg the sync, post processing and email modules) and you want a JSON format to include other less necessary info then lets keep them as separate things. A module.json just doesn't work very well for the version info for installing and updating. Remember - K.I.S.S.

TrystanLea:
> As much as I would like to see a single file suit both functions, it feels like we're trying too hard to make that work

What are both functions, is this not one function, providing access to emoncms version number?

TrystanLea:
I hope Im following everyones comments above ok with this merge, I think the general agreement is that a .txt version file in the root repository makes sense and is an improvement on the previous index.php implementation.

Issue #888, opened 2018-06-18
Incorrect "en-EN" locale code used in emoncms?


pb66:
I am trying to understand some user profile code in emoncms and noticed that the default locale code seems to be "en_EN" is that intended/correct?

It doesn't seem to be a standard locale code, should we not be using "en-GB" as the default? 

(I'm new to JS and not great with PHP so apologies if I ask a stupid question)

Whilst looking at using a user account timezone in client run javascript I found the JS toLocaleString() function uses a locale to format the date and time string eg

```
var acct_time = new Date().toLocaleString("en-GB", {timeZone: "Australia/Sydney", timeZoneName: "long"})
document.getElementById('acct_time').innerHTML = acct_time;
```

so I was just pondering on whether the session['lang'] would be suitable/available to use (in addition to adding a new session['timezone'] variable) and found this code which threw me a bit.

https://github.com/emoncms/emoncms/blob/9.8.30/locale.php#L62-L93

This won't effect using a user account (session?) timezone directly, but I was thinking that displaying the current time in both session and browser timezones might be useful to the viewer.

emrysr:
Hi @pb66 ,

I added the session variables here:
https://github.com/emoncms/emoncms/pull/896/files#diff-3a8983b5e260461fe91544f3ff97de77R733

Not yet sure what will make use of them?

pb66:
When I was trying to figure out how to make the user account timezone available to dashboards and apps etc, I noticed the "lang" variable was already in the session data and I figured that was the way to go. I have since worked out that because dashboards and apps are also available without being logged in, either publicly or using read or write apikeys in the urls, this will not work for all viewing instances.

So I now do not know if that data is of any real value in the session data as any fix for non-logged in viewing will still be available for logged in viewing, I do not know why the "lang" was already there either.

But thanks for fixing the en-GB locale code, I haven't tried it yet but it looks good :-)

PR #896, opened 2018-06-25
removed en_EN

emrysr:
related to #888 

The region part of "**en-EN**" is not an ISO 3166 region/country.

Following: [the emoncms guide - Using gettext on Ubuntu](https://github.com/emoncms/emoncms/blob/master/docs/gettext.md)

I've changed the references to **en-EN** to **en-GB**. This language selection is also available in the profile page. This is done by re-naming all the locale directories within each module.

Doesn't have an adverse effect on all users using en-EN as the available translations rely on an English index to find the translation in the .mo files

Changing the default to en-US can be done in the **settings.php** file, however there are currently no translated strings within the locale directories for each module - the English version will be used.

eg.
`_('Hello World') ` will return '**Hello World**' if no alternate versions are found.


Modules/user/user_model.php
```diff
@@ -711,8 +714,12 @@ public function get($userid)
 
     public function set($userid,$data)
     {
+        global $default_language;
+        $default_locale = !empty($default_language) ? $default_language : 'en_GB';
+        $default_timezone = 'Europe/London';
```
emrysr: might be a good idea to put this in settings.php too?

Modules/user/user_model.php
```diff
@@ -721,10 +728,10 @@ public function set($userid,$data)
         $bio = preg_replace('/[^\p{N}\p{L}_\s-.]/u','',$data->bio);
         $language = preg_replace('/[^\w\s-.]/','',$data->language);
         $tags = isset($data->tags) == false ? '' : preg_replace('/[^{}",:\w\s-.]/','', $data->tags);
-        
         $startingpage = preg_replace('/[^\p{N}\p{L}_\s-?#=\/]/u','',$data->startingpage);
         
-        $_SESSION['lang'] = $language;
+        $_SESSION['lang'] = !empty($language) ? $language : $default_locale;
+        $_SESSION['timezone'] = !empty($timezone) ? $timezone : $default_timezone;
```
emrysr: not sure if these are used by anything ... any ideas where this might be usefull?

TrystanLea:
Thankyou @emrysr, Im struggling to test this here, trying to switch language, I realise our documentation here might be out of date https://github.com/emoncms/emoncms/blob/master/docs/gettext.md#4-install-gettext, did you manage to get this working with different languages?

PR #900, opened 2018-06-27
added $default_emailto to allow triggered input processes to send email

### LINKED ISSUE #764: Emails from MQTT inputs
Hi,

I have spotted an issue where when you update an input via MQTT emails attached to the input process list are not sent but are sent when the input is updated via HTTP.

There is no errors in the log file.

I thought the update to 9.8.27 may have fixed this along with the other MQTT issues but it doesn't seem to have fixed it.

Matt.

emrysr:
fixes #764 

The issue with this that the current email process used the user session to get the users email address.
Publishing via MQTT does not create a user session - so couldn't get a valid email address.

added a variable in settings.php to store a default email address.


Modules/eventp/locale/fr_FR/LC_MESSAGES/eventp_messages.po
```diff
@@ -82,7 +82,7 @@ msgid "Send Email"
 msgstr "Send Email"
 
 #: ../../../eventp_processlist.php:51
-msgid "Send an email to the user with the specified body."
+msgid "Send an email to the user with the specified body. Email sent to user's email address or default set in config."
```
emrysr: hi @thib66 
I have had to change the English string that you translated to better explain how emails are sent. I thought I'd let you know in case you might want to edit the french version?
this is what google translate suggested...
> Send an email to the user with the specified body. **Email sent to user's email address or default set in config.**
> Envoyer un e-mail à l'utilisateur avec le corps spécifié. **Email envoyé à l'adresse e-mail de l'utilisateur ou par défaut défini dans config.**

TrystanLea:
Brilliant, nice spot! @emrysr

Modules/eventp/locale/fr_FR/LC_MESSAGES/eventp_messages.po
thib66: Hello, a better translation would be : 
Email envoyé à l'adresse email de l'utilisateur ou à celle définie par défaut dans la configuration.
Regards,
Aymeric

PR #902, opened 2018-06-29
Fixed password reset bugs

cagabi:
Three bugs fixed:
 - When email could not be sent the  string to write into log file was malformed giving  an 'undefined' error in login block instead of 'An error occured'
 - When email was not be sent , the message shown to the user was "Password recovery email sent!"
 - New password was written into the database before checking $enable_password_reset is enabled and before sending the email. I have changed it to write the password into the database only if $enable_password_reset is true and if the email has been successfully sent

### DIFF
--- a/Modules/user/user_model.php
+++ b/Modules/user/user_model.php
@@ -476,24 +476,19 @@ class User
         
         if ($userid!==false && $userid>0)
         {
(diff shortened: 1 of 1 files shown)


TrystanLea:
Thanks @cagabi

PR #906, opened 2018-07-03
appname  moved from user_model.php to default.settings.php

cagabi:
I wanted the subject of the reset password email to be "MyHomeEnergyPlanner password reset" instead "emoncms password reset"


TrystanLea:
Thanks @cagabi

PR #907, opened 2018-07-04
Hotfix/mqtt connections not closing

emrysr:
answered a few questions raised in previous pr: https://github.com/emoncms/emoncms/pull/905

@pb66 - I've added a client id to the mqtt connection "emoncms".
@borpin - the publish and subscribe calls are now done only done if connected.

### DIFF
--- a/Modules/process/process_processlist.php
+++ b/Modules/process/process_processlist.php
@@ -55,7 +55,7 @@ class Process_ProcessList
                     $id (string) – The client ID. If omitted or null, one will be generated at random.
                     $cleanSession (boolean) – Set to true to instruct the broker to clean all messages and subscriptions on disconnect. Must be true if the $id parameter is null.
                  */ 
-                $mqtt_client = new Mosquitto\Client(null,true);
+                $mqtt_client = new Mosquitto\Client('emoncms',true);
                 
                 $mqtt_client->onDisconnect(function($responseCode) use ($log) {
                     if ($responseCode > 0) $log->info('unexpected disconnect from mqtt server');
--- a/scripts/phpmqtt_input.php
+++ b/scripts/phpmqtt_input.php
@@ -97,8 +97,12 @@
         require_once "Modules/device/device_model.php";
         $device = new Device($mysqli,$redis);
     }
-    
-    $mqtt_client = new Mosquitto\Client();
+    /*
+        new Mosquitto\Client($id,$cleanSession)
+        $id (string) – The client ID. If omitted or null, one will be generated at random.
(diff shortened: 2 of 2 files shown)


TrystanLea:
Thankyou @emrysr, testing here gave the following error:

#2 /var/www/emoncms/scripts/phpmqtt_input.php(120): Mosquitto\Client->loop()
#3 {main}
2018-07-04 11:24:55.099|WARN|phpmqtt_input.php|Not connected, retrying connection
2018-07-04 11:24:55.123|WARN|phpmqtt_input.php|Connecting to MQTT server: Connection Accepted.: code: 0
2018-07-04 11:24:55.124|ERROR|phpmqtt_input.php|exception 'ErrorException' with message 'Undefined variable: mqtt_server' in /var/www/emoncms/scripts/phpmqtt_input.php:173
Stack trace:
#0 /var/www/emoncms/scripts/phpmqtt_input.php(173): exceptions_error_handler(8, 'Undefined varia...', '/var/www/emoncm...', 173, Array)
#1 [internal function]: connect(0, 'Connection Acce...')
#2 /var/www/emoncms/scripts/phpmqtt_input.php(120): Mosquitto\Client->loop()
#3 {main}

the $mqtt_server variable is missing in the global definition for that function. I will make the modification here and merge in.

pb66:
> @pb66 - I've added a client id to the mqtt connection "emoncms".

That's cool, but please see https://github.com/emoncms/emoncms/pull/823 and maybe the https://community.openenergymonitor.org/t/a-question-over-whether-emoncms-is-compatible-with-mqtt/7180?u=pb66 forum post. 

If you are using a client ID you must change the QoS 2 to Qos 0 to avoid emoncms getting swamped with loads of backdated non-timestamped data that emoncms will try and process and timestamp it as "now" regardless of how old it is.

The lack of a client id and patchy connection keep the buffer clean of stale data naturally. adding the client id with QoS 2 will cause massive amounts of input processing after a connection issue and create lots of incorrect data. Depending on whether it's a buffered-write setup or not, either the first or the last data processed for any given whole second (or fixed interval in the case of phpfina) will prevail, so what type of install, how long the outage was and how fast the backlog can be processed will result in wildly different saved data.

TrystanLea:
Thanks @pb66 should we revert the added client id for now, I havent had a chance to review this properly and I would prefer not to add a different issue to these particular changes relating to publish_to_mqtt?

borpin:
Personally I think we are getting a little trigger happy with the merges to the master branch.

Paul-Reed:
> @emrysr  - the publish and subscribe calls are now done only done if connected.

But they are still accumulating in the queue, and when connected, the whole lot of backdated data is released to the broker, which could be hours worth.

Issue #903, opened 2018-07-01
Can MQTT connection happen outside of publish to mqtt process?


TrystanLea:
Can MQTT connection happen outside of publish to mqtt process? see:

See: https://community.openenergymonitor.org/t/mqtt-log-entries/7784/4

TrystanLea:
Perhaps the process model can reuse the existing mqtt connection started by the phpmqtt_input script or if not available e.g when being used as part of the HTTP route, it could start a single publishing client if connect was called in the _construct rather than the publish_to_mqtt method itself

Paul-Reed:
I've moved a couple of my comments here, to try and keep the conversation in one place!!
> In https://github.com/emoncms/emoncms/pull/907 @emrysr wrote;  - the publish and subscribe calls are now done only done if connected.

But they are still accumulating in the queue, and when connected, the whole lot of backdated data is released to the broker, which could be hours worth.

In https://github.com/emoncms/emoncms/pull/908 the clientid has been reverted, but Nick O'Leary (node-red author) got around this for his [pubsubclient mqtt library](https://github.com/knolleary/pubsubclient/blob/master/examples/mqtt_esp8266/mqtt_esp8266.ino) by prefixing a random string with the client ID name, so it can easily be recognized.
ie `EMONCMSclient-3b77e8`

The C++ code is below;

```
    // Create a random client ID
    String clientId = "ESP8266Client-";
    clientId += String(random(0xffff), HEX);
    // Attempt to connect
    if (client.connect(clientId.c_str())) {
      Serial.println("connected");
```

pb66:
from #907 @TrystanLea asked

> Thanks @pb66 should we revert the added client id for now, 

IMO it might be marginally better to have a correctly working QoS 0 implementation than a partial QoS 2 implementation resulting in something similar but not entirely like a QoS 0 set up.

Having a client id is the right thing to do, having a QoS of 0 when there is no timestamp provided with the input data is the right combination regardless of client id.

> I would prefer not to add a different issue to these particular changes relating to publish_to_mqtt?

That's fine to want to have sepearte PR's but we do need to discuss the MQTT as a whole, I have created a mqtt branch so that we can perhaps test several things and try to resolve several things before merging quick hotfixes to the master repo and reducing our options.

in #905 @TrystanLea said

>  Can anyone think of a better way to solve the buffering issue?

[...]

PR #908, opened 2018-07-04
removed clientid from mqtt connect()

emrysr:
Reverted the client ID due to discussions on this issue:
- Previous Pull Request: https://github.com/emoncms/emoncms/pull/907

Comments:
- Community Forum: https://community.openenergymonitor.org/t/a-question-over-whether-emoncms-is-compatible-with-mqtt/7180?u=pb66
- Pull Request: https://github.com/emoncms/emoncms/pull/823

I'll change it when an issue is created. 

thanks for the help @pb66 - I jumped the gun on this one!


TrystanLea:
Thanks @emrysr

pb66:
#903 added a client id in 2 places and this PR only reverts one.

As I understand it the mqtt code in the process module should now be redundant so this PR should have no effect.

The place where a client id with QoS2 and no timestamps will cause a problem is in the phpmqtt script which is still using a client id. That's where it needs reverting or changing to qos0

PR #919, opened 2018-07-12
moved bargraph value hover container to centre bottom

### LINKED ISSUE #673: Bargraph tooltip blocks buttons
I don't personally use the bargraph, but just noticed that when a bar is selected, it's value & date appear in the graph frame top-left, on top of the D, W, M, & Y buttons, meaning that they cannot be selected.
If the graph size is really wide, then there would be sufficient space, but on smaller/medium graphs this presents a problem.
A possible solution may be to ensure that the label disappears on 'mouseout'.
Currently the label appears on 'mouseover' of the respective bar, but is then retained no matter where the mouse is moved.

label

emrysr:
fixes #673 

TEXT WAS OVERLAYING BUTTONS
image

FIXED BUY MOVING TEXT
image

### DIFF
--- a/Modules/vis/visualisations/bargraph.php
+++ b/Modules/vis/visualisations/bargraph.php
@@ -22,7 +22,16 @@
 <script language="javascript" type="text/javascript" src="<?php echo $path;?>Modules/vis/visualisations/common/vis.helper.js"></script>
 
 <div id="vis-title"></div>
-
+<style>
+.stats-container{
+  position: absolute;
+  bottom: 2.3em;
+  width: 100%;
+  text-align: center;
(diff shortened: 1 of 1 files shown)


TrystanLea:
Thankyou @emrysr

emrysr:
sorry @TrystanLea I'd noticed an issue with the tooltip positioning.. I've corrected it in another pull request https://github.com/emoncms/emoncms/pull/920

emrysr:
sorry @TrystanLea I'd noticed an issue with the tooltip positioning.. I've corrected it in another pull request https://github.com/emoncms/emoncms/pull/920

PR #922, opened 2018-07-15
changes from https://github.com/emoncms/emoncms/pull/822

TrystanLea:
See original https://github.com/emoncms/emoncms/pull/822

### DIFF
--- a/scripts/phpmqtt_input.php
+++ b/scripts/phpmqtt_input.php
@@ -105,6 +105,7 @@
     $mqtt_client = new Mosquitto\Client('emoncms',true);
     
     $connected = false;
+    $subscribed = 0;
     $last_retry = 0;
     $last_heartbeat = time();
     $count = 0;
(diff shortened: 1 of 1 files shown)


borpin:
I think it is worth keeping the warning that a connection attempt has been made.

The publish needs to check for a connection else that will raise an exception (not part of my original!).

TrystanLea:
> I think it is worth keeping the warning that a connection attempt has been made.

Did I remove a warning?

borpin:
>Did I remove a warning?

Ah, sorry, I didn't notice you had moved it.  Makes more sense to do it before the actual attempt.

>Let me know if there are any modifications I need to make?

https://github.com/emoncms/emoncms/blob/776384d9ece35468cc08750356f0a9fc688443ba/scripts/phpmqtt_input.php#L143
Strictly this should be 'Loop through all topics with data to publish' - queue implies more than one value for each topic.

https://github.com/emoncms/emoncms/blob/776384d9ece35468cc08750356f0a9fc688443ba/scripts/phpmqtt_input.php#L144-L148

Not able to do it myself from here (or more accurately don't know how to do it in the web interface) but it really needs a 'connected' check before line 144.

```php
if ($connected) {
    $publish_to_mqtt = $redis->hgetall("publish_to_mqtt"); 
    foreach ($publish_to_mqtt as $topic=>$value) { 
         $redis->hdel("publish_to_mqtt",$topic); 
         $mqtt_client->publish($topic, $value); 
    }
}
```
Looks good!

borpin:
@TrystanLea You didn't make the change I suggested above and add in a 'connected' check before publishing.

TrystanLea:
aha, ok will add that in

TrystanLea:
Added in https://github.com/emoncms/emoncms/commit/aa97e9064ed1aec5ddf982923885b568ca8b62e9 thanks @borpin

PR #931, opened 2018-07-21
Fix cookie path restriction when running on domain root

takkaria:
#918 introduced a fix for cookies so that if emoncms is running at e.g. domain.net/emoncms, then cookies it saves would be restricted to that path and below.  However it also broke logins if Emon is running not in a subdirectory because it set a restriction to the path `//` rather than `/`.  This pull request fixes that behaviour.


TrystanLea:
Thankyou @takkaria @emrysr

PR #932, opened 2018-07-24
Feature/issue 917 device module release

emrysr:
Created ui for user to save opt in preferences for beta features. Checks for `$_COOKIE["EMONCMS_BETA_OPTIN"] `which is stored as JSON `{"deviceModule":"true"}`

`$user->setBetaOptIn()` sets a cookie with a expiry 2 years into the future.

Sets the `$ui_version_2` in **index.php** to `true` if the cookie value is "true" to allow the user to toggle on/off the new features. Defaults to false / Off

The JSON object key is created by getting the value of the ` "data-prop"` attribute of any element within the form that has the class "`controls`" in [user/profile/profile.php](https://github.com/emoncms/emoncms/compare/device-support...emrysr:feature/issue-917-device-module-release?expand=1#diff-cd18b294e9e3cb5763244200e1e46ad9R125)

`<div class="controls" data-prop="deviceModule">`

The **reset** button makes the cookie expire and looses all the saved settings (currently only one)

This includes loading indicators and error messages (translatable in the php) - adding possible other beta features opt ins will be easy using this ui. 

image

### DIFF
--- a/Modules/user/profile/profile.php
+++ b/Modules/user/profile/profile.php
@@ -121,7 +121,32 @@ function languagecode_to_name($langs) {
     <div class="span8">
         <h3><?php echo _('My Profile'); ?></h3>
         <div id="table"></div>
-        
+
+        <div id="beta-section" class="well hidden">
+            <h4><?php echo _('Beta Features'); ?>:</h4>
+            <form id="beta-form" class="form-horizontal" style="margin-bottom:.2em">
+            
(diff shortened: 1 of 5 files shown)


emrysr:
image

will not work until database is updated / synced as new table field is required


`if $optIns['deviceModule']` is present and set to true then the  `$ui_version_2` is set to `true` 

this enables all the new features in the UI for the logged in user

Issue #744, opened 2017-11-17
Enhanced input processing display labels 


pb66:
Here is a small example of an identical set of inputs on emoncms.org and locally hosted emoncms (only the input name differ)

![screenshot4](https://user-images.githubusercontent.com/5606042/32945250-99baffc6-cb8a-11e7-8b78-3f3b9fe9ac60.png)

![screenshot3](https://user-images.githubusercontent.com/5606042/32945258-a031b7aa-cb8a-11e7-81f1-2006739a4cba.png)

The emoncms.org labels are IMO more informative, compact, colorful and largely language agnostic, I would like to see a return to those tags, but I've been reluctant to implement local changes that need stashing every time I upgrade and may cause merge conflicts etc.

Would there any interest (or objections?) to me re-implementing this for inclusion in the main repo?

When you have lots of inputs with long processlists it makes a big difference to break the lists up with some color for easyier viewing and also to the size on page. 

<img width="895" alt="2017-11-17" src="https://user-images.githubusercontent.com/5606042/32945442-5289210e-cb8b-11e7-88e6-d42151806244.png">
 
(forum thread https://community.openenergymonitor.org/t/enhanced-input-processing-display-labels/5674?u=pb66)

Paul-Reed:
I've noticed the same, and currently it's harder to visualise processes, as they look so similar, so yes, gets my vote!
I assume you will try and establish a protocol, such as all conditional processes = green, all mathematical processes = blue, etc, etc.

Paul

pb66:
_"I assume you will try and establish a protocol,"_

Try being the operative word there, when we did it last time the code was very different and the process module or ability to drop in additional processing modules didn't exist, so I do not currently know whats possible, but I thought I would ask the question before investing loadsa time in it.

Thanks for the reply, hopefully other will feel the same way.

pb66:
Not wanting to get ahead of myself, but what has also become apparent whilst solely creating processlists for the last 5hrs solid is that a "quick link bar" showing ALL the processes as colored buttons as an alternative to using the drop down selector would make things soooo much faster.

a quick mock up with only a handful of process "quick link buttons" to convey the idea

<img width="943" alt="2017-11-17 1" src="https://user-images.githubusercontent.com/5606042/32948759-474ce1d8-cb98-11e7-9bb1-4f10e05f5c5f.png">

I'm not suggesting we get rid of the dropdown box, just adding "quick links", even if only for the most frequently used processes.

[...]

PR #937, opened 2018-07-27
added short_name to processlist array

emrysr:
am still working on this. uploaded for you to see progress.

@trystan I'm going to use the json file format to store process descriptions as per our dm discussion

having one place describe the processlist will make it easier to read the same information from js and php

associated to issue #744


TrystanLea:
Thanks @emrysr I think we might as well load the json processlist via php as there are multiple processlist.json files, one for each module: process, eventp & schedule. Doing it via the php loader would also allow for validation etc

emrysr:
ok - thanks

emrysr:
@trystan - this work on this branch is finished for now. you can review and merge if you want.

I'll be working on the branch processlist_refactor on the rest of this work

TrystanLea:
this is merged into the processlist_refactor branch

PR #938, opened 2018-07-27
require_once feed and input models in delete_user

cagabi:
After a while we are using in the Group module the delete_user function to fully remove users. We call this function from inside the group_model where the feed and input models have already been required


TrystanLea:
Thanks @cagabi

PR #945, opened 2018-08-01
Feature/issue 917 device module release

TrystanLea:
See original: https://github.com/emoncms/emoncms/pull/932


emrysr:
image

will not work until database is updated / synced as new table field is required


`if $optIns['deviceModule']` is present and set to true then the  `$ui_version_2` is set to `true` 

this enables all the new features in the UI for the logged in user

Issue #655, opened 2017-03-26
Clear all old data within a feed


OHarvey:
Not sure if this has been previously raised, if it has I apologise. 

I would like the ability to clear data that has been logged in a feed. 

A "clear all" would be very helpful to have in the feeds edit menu. 

Thanks.

Paul-Reed:
Agree, with the safeguard of a secondary check - 'This action is irreversible, are you sure that you want to delete this feed's data?' Y/N
This crops up regularly in the forum, and would be a nice addition.

Paul

lxxl:
Yep +1 for this.

robertsLando:
Any news about this feature? Is there any work in progress? 

I need to delete data of a feed before a date, I also would like to know if there is a way to remove all values of a feed without remove and recreate the feed. I think this feature is a must have in a software like emon, the possibility to delete datas isn't supported in anyway, also an API to delete feeds data would be great for full feeds support.

TrystanLea:
@emrysr a couple of pointers to help, we discussed two methods one for the clear all data for a particular feed and the other to delete data before a certain time.

These methods could be implemented in the feed engine's themselves, e.g PHPFina.php: https://github.com/emoncms/emoncms/blob/master/Modules/feed/engine/PHPFina.php

perhaps of the form:

    public function clear($feedid) {
        // clear code...
        // 1. clear feedid.dat
        // 2. set start_time to 0 in feedid.meta, keep interval the same
    }

    public function trim($feedid,$start_time) {
        // trim code...
        // 1. system level trim command? as discussed?
        // 2. change start_time in feedid.meta to new $start_time
    }

thanks a lot!

PR #946, opened 2018-08-01
added relative date buttons to feed trim form

emrysr:
related to issue #655 
added -1year and -2year buttons to simplify the feed trim() feature

image


TrystanLea:
Thanks @emrysr thats really nice!

Issue #952, opened 2018-08-06
Certbot let's encrypt does not work with .htacess


biva:
Hello,
During installation, I tried to activate https with letsencrypt. There was an issue (403 forbidden) during the certbot process(`sudo certbot --authenticator webroot --installer apache`).
I solved it by deactivating the .htaccess (`sudo mv .htaccess .htaccess_bkp`) and re-enabling it after the certbot process was done.
Do you also experiment this?

Config: Debian 9.5, Apache

takkaria:
Cerbot generates files in `.well-known` that they need to be able to access remotely.  At the moment this is not allowed.  To fix this you need to add a line like the following to .htaccess:

```
 RewriteCond %{REQUEST_URI} !=/.well-known/.*$
```

After the robots.txt line in the `<IfModule mod_rewrite.c>` block.

biva:
Thanks @takkaria : shouldn't it be added in the repository? What about after certbot has done its job, does it create a security problem to keep this new line in .htaccess?

takkaria:
Yes it probably should!  I will submit a pull request.

The idea behind Let's Encrypt is that you set up a system to renew the certificate automatically every 12 weeks, so if you don't have that line in .htaccess you won't be able to do that.  There is no security risk in allowing access to `.well-known`, as far as I know.

biva:
@takkaria : should I close this issue or keep it open until it's merged in the main project?

TrystanLea:
Merged thanks

biva:
Hello,
I just tried to renew my certificate with Let's Encrypt: `sudo certbot renew` and the following error appeared:

<code><pre>Domain: emoncms.mysite.com
   Type:   unauthorized
   Detail: Invalid response from
   http://emoncms.mysite.com/.well-known/acme-challenge/jfzeoifjegoirjgzoigjJOIEJFZOIFJEOFIOJEfJEOIfJmfeiJ:
   "<!DOCTYPE HTML PUBLIC \"-//IETF//DTD HTML
   2.0//EN\">\n<html><head>\n<title>403
   Forbidden</title>\n</head><body>\n<h1>Forbidden</h1>\n<p"
</pre></code>

I removed `.htaccess` and it solved the issue. So I think the patch is not enough :(

takkaria:
I can help you out here!  In this section of .htaccess:

https://github.com/emoncms/emoncms/blob/master/.htaccess#L176

can you add the line

```
[...]

PR #955, opened 2018-08-07
Allow access to .well-known for certbot (#952)

takkaria:
This is required to use Let's Encrypt SSL certificates.


TrystanLea:
thanks @takkaria

Issue #950, opened 2018-08-02
Add trim and clear to new feed list UI


emrysr:
currently the new UI shows this modal overlay when you click the delete button:
![image](https://user-images.githubusercontent.com/1466013/43594179-7f597268-9671-11e8-8a08-9de45d9a7cc4.png)

we need to re-integrate the recent changes to the "old" UI ...
![image](https://user-images.githubusercontent.com/1466013/43594217-9f54033a-9671-11e8-9eee-eac55d47b7a1.png)

Paul-Reed:
Is there anyway to catch the error which occurs if there are any existing processes for a deleted feed, or better still, make it not possible to delete the feed if there are any, and display a warning instead.
eg, "This feed cannot be deleted as there are input processes using it. Please delete the processes first and then retry"

PR #957, opened 2018-08-08
added trim() and clear() to new layout

emrysr:
supports multiple selections
checks for associated input process

fixes issue #950


emrysr:
image

PR #972, opened 2018-08-14
Remove test-failing PHP versions from Travis CI

takkaria:
PHP 5.4 is older than the version listed under 'requirements' in the README and the PHPFina tests always fail on it, so I removed it and 5.5 from the Travis php version list and replaced them with 7.1 and 7.2.


TrystanLea:
thanks @takkaria

PR #975, opened 2018-08-15
additional correction

pb66:
See https://community.openenergymonitor.org/t/feeds-are-missing-after-reboot/8193/33?u=pb66
sudo su is required as sudo alone doesn't work


glynhudson:
Thanks :+1:

PR #982, opened 2018-08-20
Feature/match feedlist styling in inputlist

emrysr:
separated the similar js and css from the device list and feed list views.
now uses bootstrap's accordion to manage node visibility
added select all in node option in the header 

@TrystanLea what do you think of this "select all in group"?  I'm not totally sure?
image


TrystanLea:
Epic! thanks @emrysr

PR #994, opened 2018-08-22
Add Git info to administration page

borpin:
Add the Git URL and the branch into the administration module.  When answering queries, it is often useful to know which branch the user is on as some questions may have already been solved in the master branch.

I have taken the URL from ls-remote. Not being an expert on Git this may not be the best method but was the only one I could find to just get the origin URL.


glynhudson:
Nice, good idea. This works well

selection_291

borpin:
Odd, must say I didn't actually try and switch branches 😦

Issue #980, opened 2018-08-20
localhost displayed incorrectly on admin page


danbates2:
Seems to show 127.0.1.1 instead of 127.0.0.1 .
https://community.openenergymonitor.org/t/sbfspot/8245/59

borpin:
Oh shoot. Just opened #995 - sorry.

PR #998, opened 2018-08-23
Feature/issue 980 localhost displaying admin page

emrysr:
fixes issue #980 

This was the simplest fix. adding a ".home" to the gethostbyname() parameter. This then queries the /etc/hosts file for the network name not the localhost name.

This works on my local network. requires further testing on other networks to see if the ".home" suffix is suitable.

I've found a more reliable solution that opens an external UDP connection to get the local network IP address:
```
function getLanHostname(){
  $wanAddr = '8.8.8.8';
  $sock = socket_create(AF_INET, SOCK_DGRAM, SOL_UDP);
  socket_connect($sock, $wanAddr, 53);
  socket_getsockname($sock, $localAddr); // $name passed by reference
  return $localAddr ;
}

```
If this pull request's approach isn't suitable I'll change it to this - I didn't want to create an external network connection if possible.

https://github.com/emoncms/emoncms/issues/980

### DIFF
--- a/Lib/responsive-linked-tables.js
+++ b/Lib/responsive-linked-tables.js
@@ -141,8 +141,8 @@ function list_format_value(value) {
 function autowidth($container) {
     let widths = {}, // only store widest column values against each selector
     default_padding = 20;
-    // tbody
-    $container.find(".tbody [data-col]").each(function() {
+    // resize columns based on columns in tbody and thead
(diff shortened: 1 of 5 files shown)


emrysr:
Not the correct pull request... sorry

emrysr:
**please ignore!!!** 

mixed pull pull request with incorrect issue!

PR #999, opened 2018-08-23
gethostbyname() now uses a '.home' suffix to query LAN ips instead of…

### LINKED ISSUE #980: localhost displayed incorrectly on admin page

### LINKED ISSUE #995: Server IP address is not correct in Administration screen
The IP Address returned under the server information is not the true server address - I see a 127.0.1.1 address.  Is there a reason for the IP being discovered in this way?

By changing this line

https://github.com/emoncms/emoncms/blob/c33d4e419231ead9e20614463561bdf772551dc0/Modules/admin/admin_main_view.php#L48

to use the system variable `$_SERVER['SERVER_ADDR']` the true IP (in the sense the IP the HTTP server would use) would be returned.

emrysr:
… the localhost ip
fixes #995 
fixes #980 

### This corrects a closed pull request #998 

This was the simplest fix. adding a ".home" to the gethostbyname() parameter. This then queries the /etc/hosts file for the network name not the localhost name.

This works on my local network. requires further testing on other networks to see if the ".home" suffix is suitable.

I've found a more reliable solution that opens an external UDP connection to get the local network IP address:
```
function getLanHostname(){
  $wanAddr = '8.8.8.8';
  $sock = socket_create(AF_INET, SOCK_DGRAM, SOL_UDP);
  socket_connect($sock, $wanAddr, 53);
  socket_getsockname($sock, $localAddr); // $name passed by reference
  return $localAddr ;
}

```
If this pull request's approach isn't suitable I'll change it to this - I didn't want to create an external network connection if possible.

### DIFF
--- a/Modules/admin/admin_main_view.php
+++ b/Modules/admin/admin_main_view.php
@@ -40,12 +40,12 @@
         }
         $emoncms_modules .=  str_replace($emoncmsModulesPath."/", '', $emoncmsModuleFolder);
     }
-
+    $local_dns_tld = '.home'; // .local can sometimes not be compatable
     return array('date' => date('Y-m-d H:i:s T'),
                  'system' => $system,
                  'kernel' => $kernel,
                  'host' => $host,
-                 'ip' => gethostbyname($host),
+                 'ip' => gethostbyname($host.$local_dns_tld),
                  'uptime' => @exec('uptime'),
                  'http_server' => $_SERVER['SERVER_SOFTWARE'],
                  'php' => PHP_VERSION,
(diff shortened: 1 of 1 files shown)


borpin:
Why not just use the `$_SERVER['SERVER_ADDR']` variable as per #995?

Adding .home will not work if your network uses .local.  What if you are on an external IP address?

emrysr:
I've changed this to use the command `hostname -I `to get the IP address as `$_SERVER['SERVER_ADDR'] `returns my localhost ip address

here is the output of the different methods I tested to get the network IP address:
image


here's the code:
```

<?php
echo "<pre>";
print_r(array(
  'php_uname("n")'=> php_uname('n'),
  'gethostbyaddr()'=> gethostbyaddr($_SERVER['SERVER_ADDR']),
  'getHostByName()'=> getHostByName(php_uname('n')),
  'SERVER_ADDR'=> $_SERVER['SERVER_ADDR']
));
if ( $_SERVER["SERVER_ADDR"] == '127.0.0.1' ) {
  function getLanHostname(){
    $wanAddr = 'emoncms.org';
    $sock = socket_create(AF_INET, SOCK_DGRAM, SOL_UDP);
    socket_connect($sock, $wanAddr, 53);
    socket_getsockname($sock, $localAddr); // $name passed by reference
    return $localAddr ;
  }
  function getHostnameI(){
    exec("hostname -I",$ifconfig);
    return $ifconfig;
  }
  print_r(array(
    'php_uname("n")'=> php_uname('n'),
    'getHostByName() .local'=> getHostByName(php_uname('n').'.local'),
    'getHostByName() .home'=> getHostByName(php_uname('n').'.home'),
    'getLanHostname()' => getLanHostname(),
    'hostname -I' => getHostnameI()[0]
  ));
}
exit()
?>
```

emrysr:
I'd pulled from master before committing so there are other changes I've not done.

I've changed line 13 & 47 of Modules/admin/admin_main_view.php

borpin:
I get an output of:
```
Array
(
    [php_uname("n")] => DietPi-Emon
    [gethostbyaddr()] => 192.168.7.46
    [getHostByName()] => 127.0.1.1
    [SERVER_ADDR] => 192.168.7.46
)
```
I wonder if the HTTP_HOST index might be a better value?

borpin:
Hi @emrysr 

I think the only way is to try different methods until you find an IP address that is not text and not in the 127 range!  There seem to be a multitude of ways to setup a machine so no one method is guaranteed to return the right IP address.

emrysr:
This pull request now uses the php "SERVER_ADDR" value. This should be suitable for most cases.

We could add a "Network" section in the future to show detailed information on the separate localhost, LAN and WAN IP addresses. This would require adding an API endpoint called via javascript that runs the above `socket_create()` function and returns that value to the user as they request it (instead of creating an external network connection by default)

Issue #923, opened 2018-07-15
Introduce a hierarchical settings file structure


borpin:
Could a hierarchical user settings structure be introduced in a similar way to most other settings in UNIX.

Emoncms should first load the default settings file, then load a user settings file that simply over rules any default settings.

At present, each time a change is made to the settings file (new setting) the user has to recreate the setting file.  By making this hierarchical, a new setting file should not disturb the current behaviour but all ow the user to add in settings as required.

This could be done using the current 2 files (the default and the one the user creates).  I cannot see that there should be any issues with a migration to this system.

borpin:
Hi @TrystanLea 
I've done a bit more digging around this and it seems using the built-in PHP methods for parsing ini files is likely to be the way forward.

I found this in one of the comments when looking at the PHP docs http://php.net/manual/en/function.parse-ini-file.php;

> Working on a project for a client recently, I needed a way to set a default configuration INI file, but also wanted to allow the client to override the settings through the use of a custom INI file.
I thought array_merge or array_merge_recursive would do the trick for me, but it fails to override settings in the way that I wanted. I wrote my own function to do what I wanted. It's nothing spectacular, but thought I'd post it here in case it saved someone else some time.
```<?php
function ini_merge ($config_ini, $custom_ini) {
  foreach ($custom_ini AS $k => $v):
    if (is_array($v)):
      $config_ini[$k] = ini_merge($config_ini[$k], $custom_ini[$k]);
    else:
      $config_ini[$k] = $v;
    endif;
  endforeach;
  return $config_ini;
};
$CONFIG_INI = parse_ini_file('../config.ini', TRUE);
$CUSTOM_INI = parse_ini_file('ini/custom.ini', TRUE);
$INI = ini_merge($CONFIG_INI, $CUSTOM_INI);
?>
```
>This allowed me to put the default INI file above the web root with information that requires extra security (database connection info, etc.) and a writable INI file within the structure of the site without affecting the default settings of the default config.ini file.

What I do not really know is whether there is an easier way than simply then assigning the elements of the resulting array out into the current variables.

This could easily be backwards compatible with any `settings.php` file taking precedence but a warning given to the user.

PR #1010, opened 2018-08-29
Initial comit of hierarchical ini files to replace settings.php

borpin:
Relates to #923 

This commit converts the settings.php to a settings.ini structure.

WARNING - ini files in root folder are a BAD idea. Location is just for testing.

To test, copy default-settings.ini to a settings.ini.  Add in your specific settings to the ini file (delete any left unchanged to test concept).  Delete the contents of your settings.php and refresh.

Not finished yet but works.

* Could make it look for other ini files.
* Does not cover emonpi settings.
* location of settings files to be discussed (must not be in root folder).
* Is there a better way to assign each variable.
* do we need to maintain the Environmental variables?

### DIFF
--- a/default-settings.ini
+++ b/default-settings.ini
@@ -1,0 +1,179 @@
+; 1 #### Mysql database settings
+[database]
+server   = "localhost"
+database = "emoncms"
+username = "_DB_USER_"
+password = "_DB_PASSWORD_"
+port     = "3306"
+;   Skip database setup test - set to false once database has been setup.
+dbtest = true;
+
+
+; 2 #### Redis
+[redis]
+redis_enabled = false
+redis_server[host] = 'localhost'
+redis_server[port] = 6379
+redis_server[auth] = ''
+redis_server[prefix] = 'emoncms'
+
+
+; 3 #### MQTT
+[MQTT]
+;   The 'subscriber' topic format is rx/* - where * is the emoncms input node number.
+;   The 'publisher' topic format is user selectable from the 'Publish to MQTT' input process, for example power/solar
+;   Activate MQTT by changing to true
+mqtt_enabled = false
+mqtt_server[host] = 'localhost'
+mqtt_server[port] = 1883
+mqtt_server[user] = ''
+mqtt_server[password] = ''
+mqtt_server[basetopic] = 'emon'
+
+
+; 4 #### Engine settings
+[engine_settings]
+;feed_settings = array(
+;   Supported engines. Uncommented engines will not be available for user to create a new feed using it. Existing feeds with a hidden engine still work.
+;   Place a ',' as the first character on all uncommented engines lines but first.
+;   If using emoncms in low-write mode, ensure that PHPFIWA is disabled by removing the leading ; , from the PHPFIWA entry
+;engines_hidden'=>array(
+;  Engine::MYSQL     ;   0  Mysql traditional
+;  Engine::MYSQLMEMORY   ;  8  Mysql with MEMORY tables on RAM. All data is lost on shutdown
+;  Engine::PHPTIMESERIES ;  2
+;  ,Engine::PHPFINA  ;   5
+;  ,Engine::PHPFIWA  ;   6
+;  ,Engine::CASSANDRA;   10 Apache Cassandra
+
+;   Redis Low-write mode
+;   If enabled is true, requires redis enabled and feedwriter service running
+redisbuffer[enabled] = false
+;   Number of seconds to wait before write buffer to disk - user selectable option
+redisbuffer[sleep] = 600
+
+;   Max csv download size in MB
+csvdownloadlimit_mb = 25
+
+;   Engines working folder. Default is /var/lib/phpfiwa,phpfina,phptimeseries
+;   On windows or shared hosting you will likely need to specify a different data directory--
+;   Make sure that emoncms has write permission's to the datadirectory folders
+phpfiwa[datadir] = '/var/lib/phpfiwa/'
+phpfina[datadir] = '/var/lib/phpfina/'
+phptimeseries[datadir] = '/var/lib/phptimeseries/'
+cassandra[keyspace] = 'emoncms'
+    
+;   For use with emoncms module that require installation in home directory
+;homedir = "/home/username"
+
+;   Max number of allowed different inputs per user. For limiting garbage rf data
+max_node_id_limit = 32
+
+
+; 5 #### User Interface settings
+[interface]
+;   gettext  translations are found under each Module's locale directory
+default_language = 'en_GB'
+    
+;   Theme location (folder located under Theme/, and must have the same structure as the basic one)
+theme = "basic"
+;   Theme colour options: "standard", "blue", "sun"
+themecolor = "blue"
+
+;   Favicon filenme in Theme/$theme
+favicon = "favicon.png"
+
+;   Use full screen width
+fullwidth = true
+
+;   Main menu collapses on lower screen widths
+menucollapses = false
+
+;   Enable multi user emoncms.
+;   If set to false, emoncms will automatically remove the register form and
+;   ability to create further users after the first user has been created
+enable_multi_user = false
+
+;   Enable remember me feature
+enable_rememberme = true
+
+;   Allow user to reset his password
+enable_password_reset = false
+
+;   Email address to email proccessed input values
+default_emailto = 'emrys@localhost'
+
+;   (OPTIONAL) Email SMTP, used for password reset or other email functions
+smtp_email_settings[host] = "smtp.gmail.com"
+;   Possible ports 25, 465, 587
+smtp_email_settings[port] = "465"
+smtp_email_settings[from_email] = 'noreply@emoncms.org'
+smtp_email_settings[from_name] = 'EmonCMS'
+;   comment lines below that dont apply
+;   ssl, tls
+smtp_email_settings[encryption] = "ssl"
+smtp_email_settings[username] = "yourusername@gmail.com"
+smtp_email_settings[password]= "yourpassword"
+
+;   Default controller and action if none are specified and user is anonymous
+default_controller = "user"
+default_action = "login"
+
+;   Default controller and action if none are specified and user is logged in
+default_controller_auth = "feed"
+default_action_auth = "list"
+
+;   Public profile functionality
+;   Allows http:; yourdomain.com/[username]/[dash alias] or ?id=[dash id]
+;   Alternative to http:; yourdomain.com/dashboard/view?id=[dash id]
+;   Add optional '&embed=1' in the end to remove header and footer
+public_profile_enabled = true
+public_profile_controller = "dashboard"
+public_profile_action = "view"
+
+;   Default feed viewer: "vis/auto?feedid=" or "graph/" - requires module https:; github.com/emoncms/graph
+feedviewpath = "vis/auto?feedid="
+
+
+; 6 #### Other settings
+[other]
+;   Log file configuration
+log_enabled = true
+;   On windows or shared hosting you will likely need to specify a different logfile directory
+log_filename = '/var/log/emoncms.log'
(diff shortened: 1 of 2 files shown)


chaveiro:
Dont get me wrong its a good initiative. 
But aren't ini files kind of deprecated and the trend seems to be json?

borpin:
I picked ini files as there is a built-in mechanism to parse them.  Perhaps there is something similar for JSON but I didn't come across it.

The key argument against JSON is that it is less readable - I have never tried, but can you insert comments into a JSON structure?  If not I think that would remove it as a possible format.

You could say the same for YAML files being the next thing after JSON.

Personally I think it is about using the right tool for the job.  From what I have researched, for php, ini files seem to be the best solution (but I am willing to be convinced otherwise).

takkaria:
.ini files are OK but they make life difficult for if you want to use container systems like docker (which we are doing at Carbon Coop).  We would need some syntax to say "import from environment variable", e.g.

```ini
[redis]
enabled = false
host = {{REDIS_HOST}}
port = {{REDIS_PORT}}
auth = ''
prefix = 'emoncms'
```

then host would be set to `getenv('REDIS_HOST')`.

borpin:
@takkaria 
>.ini files are OK but they make life difficult for if you want to use container systems like docker (which we are doing at Carbon Coop). We would need some syntax to say "import from environment variable", e.g.

Would that syntax, if you put it in the settings.ini, get imported correctly for your purposes?  If so, you can easily use that syntax in your settings.ini.

takkaria:
I don't think PHP has support for that built in, no - it would require some processing on EmonCMS' side.

borpin:
Perhaps therefore, we need to leave a mechanism in for importing specific settings via a php based settings file (as it does at the moment).  That could be triggered by a setting :smile:.  It is proabably reasonable to assume that anyone installing it into a docker type environment probably know what they are doing.

This hierarchical system is designed to make it easier for 'standard' users to update when additional settings are added. Currently you need to create a new settings file.  That should become unnecessary.

takkaria:
Hmm, I we could simplify things a lot over the current setup by just adding support for `{{DOUBLE_BRACE}}` env vars in the .ini file.  Then you don't need the `allow_config_env_vars` setting or all that code in process_settings.php.  I am happy to write the code to do this if you want?

takkaria:
Also, could I suggest a simplification of the format as well?  Because once this lands it will be annoying to change it in future.  The redis & mqtt sections could be changed so the keys are just `enabled`, `server`, `port` etc rather than `redis_server[port]` or `mqtt_server[password]` etc.  Also would be good to split out the SMTP email section similarly.  The sub-array syntax `array[key]` isn't very user-friendly I think.  Also happy to submit a patch for this if you want

borpin:
Hi @takkaria yes happy for you to contribute.  I deliberately started with a simple replication of the existing settings (I did have to change 2).

As there is already a mechanism to pick up environmental variables, is there a need for the double braces?  I don't have any knowledge of this area.  If you feel it is the right way to go, crack on 😄

TrystanLea:
I've been discussing this issue with @emrysr here. I wonder if we can separate this proposal into two separate parts:

1. hierarchical settings
2. settings file format & parsing

**hierarchical settings**

We could implement hierarchical settings using php settings files as a first step without changing the location for settings files or file formats: I have put together a proposal for how this could be done here: https://github.com/emoncms/emoncms/compare/hierarchical_settings_php

There is a defaults file called defaults.php, then a user changes default.settings.php to settings.php and enters only the relevant settings they wish to change (defaults.php can be used as a reference for this if needed).

One benefit of this hierarchical approach already is the simplification of process_settings.php which no longer has to check for the existence and default state of many settings variables, cutting down one source of overhead.

**settings file format & parsing**

I am a little concerned that we might add another source of overhead by using a different file format, e.g composer packages for yaml or taml, parsing time? It may not be significant and outweighed by the benefits?

borpin:
>There is a defaults file called defaults.php, then a user changes default.settings.php to settings.php and enters only the relevant settings they wish to change (defaults.php can be used as a reference for this if needed).

I think that is basically the same as I have.  I had a feeling that php files could not be structured this way but I may be wrong.  That was the driver behind using a different format.  However, if you want to provide a mechanism for extracting current settings (for debugging), the ini format may make that simpler.

So I am clear, in this world Emoncms would first load the defaults.php file and then the settings.php file such that this would overwrite those defaults.

On system update, only the defaults.php (and default.settings.php) would be updated by a git pull and the default state would be no change for any user.  By then moving new settings to the settings.php file, those settings could be implemented by the user.

I think the defaults.php file should only be amended in the future by adding new settings to the bottom of the file below a version number.  If it is a modification to a set of settings (rather than a completely new setting), that set should move with suitable annotation.  This then will make it easier to identify what has changed and what needs to be transferred across.

I suggest that the defaults.settings.php file is initially set with all options commented out to reduce the risk of incorrect 'default' setting.

The aim from my perspective is to make it simpler to add in new settings/features without the risk of breaking a working system if the settings have not been updated, and also knowing what settings might need to be updated.

TrystanLea:
Apologies for the delayed reply.

> I think that is basically the same as I have

Great

> So I am clear, in this world Emoncms would first load the defaults.php file and then the settings.php file such that this would overwrite those defaults.

Yes that is correct

> On system update, only the defaults.php (and default.settings.php) would be updated by a git pull 

yes thats correct but that may result in a change visible to the user if the setting that is being changed has not been overwritten by the users settings file.

> I think the defaults.php file should only be amended in the future by adding new settings to the bottom of the file below a version number. If it is a modification to a set of settings (rather than a completely new setting), that set should move with suitable annotation. This then will make it easier to identify what has changed and what needs to be transferred across.

I appreciate the idea, but worry that the settings file could become confusing and unordered?

> I suggest that the defaults.settings.php file is initially set with all options commented out to reduce the risk of incorrect 'default' setting.

I think that will work, default.emonpi.settings will need to be set but default.settings could be all commented.

Great, should we proceed with merging this at the best next available point? perhaps we should schedule for after the current image release to allow time for testing?

borpin:
> Great, should we proceed with merging this at the best next available point? perhaps we should schedule for after the current image release to allow time for testing?

Introducing it, should not change anything anyone already has.  It will be more relevant to new users and as new features (requiring settings) as they are introduced.

> I appreciate the idea, but worry that the settings file could become confusing and unordered?

To avoid this, there would need to be clear instructions in the release notes on any new settings introduced.

borpin:
@TrystanLea PR #1094 is possibly a good test case.  Not familiar enough with PHP but if the default is the new setting array, will the existing settings overwrite and leave the new setting alone or will it complain of different size of array?

TrystanLea:
> will the existing settings overwrite and leave the new setting alone or will it complain of different size of array? 

it will overwrite. 

I have created a pull request for the php version for reference here: https://github.com/emoncms/emoncms/pull/1098

greeebs:
I'm watching this with interest and considering how this is going to impact the new python-based service-runner (specifically how I get any changes to those redis settings loaded by it at startup...)

takkaria:
I have a couple of suggested changes to this over at (takkaria/emoncms#settings-ini)[https://github.com/takkaria/emoncms/commits/settings-ini).

The [first commit](https://github.com/takkaria/emoncms/commit/0ed523ec5317440f90d640bb634a9746a5cac37b) turns on the section parser of `parse_ini_file()`, so that when you write e.g.

```
[database]
host = blah
```

it shows up in `$ini['database']['host']` rather than just in `$ini['host']`.  I then did a bit of puttings things that seemed like they should go together in new sections.  I think it's a bit neater.

The [second commit](https://github.com/emoncms/emoncms/commit/dd19d0c597864307919ddbebd07ba3b79e666142) adds support for `{{ENV_VAR}}` syntax to the ini files.  So when you write the following in the config file:

```
[database]
host = {{MYSQL_HOST}}
```

... the code tries to get the value of the environment variable `MYSQL_HOST` (producing an error if not set).  This supercedes the previous behaviour of having a load of hardcoded variables set with `allow_config_env_vars` I think.

borpin:
> turns on the section parser of parse_ini_file()

This disadvantage of using sections (I did consider it) is that the order of the settings becomes important.  If each item is unique, the order does not matter.  Whilst that may be clear and obvious to those used to it, it may not be obvious to everyone so could introduce errors.

The env vars bit is good.

I think @TrystanLea is going for sticking with a PHP format but use a hierarchy.

takkaria:
Ah, I think it's confusing to also include sections in the ini file if they don't have any practical effect.  Which is why I enabled them.

Fair enough re just using PHP.  I think I generally agree that having a config file in a format that's not in PHP makes sense from an end-user point of view, fewer things to break.

borpin:
Just reviewed #1098 whic is a PR from @TrystanLea to solve this issue using php files.

It seems to me that the difference between that (php) approach and this (ini), is that #1098 simply overwrites one set of settings with a different set rather than **merging**.

This does not achieve the aim of allowing a new setting to be added without breaking and existing installation.  If a new setting is introduced, the user must still include it into their settings file.

My aim is that, for instance, if there was a need to introduce a user setting for MQTT ClientID to replace the hardcoded value, this could easily be achieved with the ini format.  By including a default the same as the current value, any update would have zero impact.

borpin:
> Ah, I think it's confusing to also include sections in the ini file if they don't have any practical effect. Which is why I enabled them.

I included it for readability as much as anything else (plus I copied the format from somewhere else) - each section could simply be a comment though.

borpin:
I do wonder if there is a way of either using dynamic variables or the `extract` function to merge the 2 sets rather than allocating each 'manually'.  My PHP is not that advanced to know!

TrystanLea:
I've been looking at this again this morning. A key problem with my approach I think is that its harder for users to find what the available settings are. The default.settings.php files contain relatively few settings. Everything is in default.php. The user needs to know which settings to copy over and this is perhaps even more confusing than the current setup?

> This does not achieve the aim of allowing a new setting to be added without breaking and existing installation. If a new setting is introduced, the user must still include it into their settings file.

I think I see what you mean. E.g in default.emonpi.settings.php

    $redis_enabled = true;
    $redis_server['prefix'] = '';

if the user wrote 

    $redis_server = array( 'host'   => 'localhost',
                           'port'   => 6379,
                           'prefix' => 'emoncms');

that would overwrite:

    $redis_server = array( 'host'   => 'localhost',
                           'port'   => 6379,
                           'auth'   => '',
                           'dbnum'   => '',
                           'prefix' => 'emoncms');

and so auth and dbnum would be missing. I can see that that would be an issue!

TrystanLea:
noticing your modifications to process_settings.php perhaps moving to encapsulating the whole of settings in an object would be an improvement and would potentially allow for merging the php settings as you describe.

So rather than 

$redis_server = array( 'host'   => 'localhost',
                       'port'   => 6379,
                       'auth'   => '',
                       'dbnum'   => '',
                       'prefix' => 'emoncms');

we would have:

    $settings = array(
   
        ...

        "redis_server" => array( 'host'   => 'localhost',
                       'port'   => 6379,
                       'auth'   => '',
                       'dbnum'   => '',
                       'prefix' => 'emoncms');

borpin:
> noticing your modifications to process_settings.php perhaps moving to encapsulating the whole of settings in an object would be an improvement

Might do but lots of rework.  Part of what I was proposing would be such that it did not break existing setups.

Actually I was playing with dynamic generation of the settings objects using a combination of `parse_ini_file()` and `extract()`.  It was possible but not for arrays embedded in arrays.

If you were thinking of completely restructuring, I'd be inclined to go that direction.

takkaria:
It would be much better software engineering practice to keep settings in some kind of namespaced area, whether using actual PHP namespaces or some kind of accessor system.  Having everything be a global variable makes it much more difficult to differentiate where things are set.

TrystanLea:
Discussion & development continued here https://github.com/emoncms/emoncms/issues/1346

PR #1015, opened 2018-09-11
Removed warning in log file when creating phpfina feed

cagabi:
Hi,
Every now and again when checking the log file we found some warnings that seemed to have no reason and were worrying us:

`2018-09-11 12:44:29.099|WARN|PHPFina.php|get_meta() meta file does not exist '/var/lib/phpfina/130.meta'`

After having a look we noticed the warnings happened when we were creating the feeds. The warning was triggered by the the get_meta() method of the phpFINA engine that was used to check if the feed exists before creating it. I have changed it to just check the meta data file exists which is the first thing that get_meta() does


TrystanLea:
thanks @cagabi

Issue #1012, opened 2018-09-04
Wrong branch displayed in admin view


Paul-Reed:
This was introduced in git pull https://github.com/emoncms/emoncms/pull/994 by @borpin 

The git branch displayed in the admin panel does not reflect the actual branch being used, and could cause confusion....

![branch](https://user-images.githubusercontent.com/973580/45021030-6ab18780-b028-11e8-9d9c-9df27f52465f.PNG)

borpin:
Did you chnage branch?  Is it a cache problem?

The command used is 

```
"git -C " . substr($_SERVER['SCRIPT_FILENAME'], 0,strrpos($_SERVER['SCRIPT_FILENAME'], '/')) . " branch"
```

Is it possible the location variable used is incorrect?

Paul-Reed:
I've been on Master branch for at least a month, but have switched branches numerous times in the past.

It would be great if this could be fixed Brian, or if not, revoked, because it's currently providing misleading information, and could influence the support given to users through the forum.

Paul

borpin:
Hi @Paul-Reed, could you open the admin window in a private window to confirm if it is a cache problem or not please?

Paul-Reed:
It's not a cache problem, I've tried a incognito chrome browser and it shows the same.

Paul-Reed:
How do you want that line to look like? just take out the `@exec`?

borpin:
@Paul-Reed see my edit above - I know why, just not how to fix it.

borpin:
I think perhaps

```git branch --contains HEAD``` might be better than ```git rev-parse --abbrev-ref HEAD```.

I can see the logic of the first command and have no idea what the second is doing 😄.  Sorry need to go and will not get a chance to do a PR until next week.

Paul-Reed:
That now retrieves the master branch (which is correct).

You can always sync the remote branches yourself for testing your changes, I seem to recall it's `git fetch`.

Paul

borpin:
I didn't do the first PR correctly and only submitted it to my fork. Hopefully this is now a PR for emoncms/master.

glynhudson:
Great, nice work #1017 has fixed this issue 👍

PR #1017, opened 2018-09-13
Update command to get correct git branch

borpin:
Issue #1012 identified an issue with the command to retrieve the current branch in that it did not always get the right branch.

The git branch command retrieved all branches not just the current one.

2 solutions were identified git branch --contains HEAD and git rev-parse --abbrev-ref HEAD.

I have chosen the former.


TrystanLea:
thankyou @borpin

PR #1018, opened 2018-09-15
Remove null=true from database schema, fixing #1016

### LINKED ISSUE #1016: MySQL database setup for user registration.
I recently went through an install and the problem still exists.
https://github.com/emoncms/emoncms/issues/791

I also had to do the following:
ALTER TABLE users CHANGE preferences preferences TEXT CHARACTER SET utf8 COLLATE utf8_general_ci NULL DEFAULT NULL;

takkaria:
As a stopgap, this fixes #1016, which prevents a new instance of EmonCMS on a blank database being created.  It is caused by a bug in the database schema code but this is a good stopgap fix.

### DIFF
--- a/Modules/user/user_schema.php
+++ b/Modules/user/user_schema.php
@@ -19,11 +19,11 @@ $schema['users'] = array(
     'language' => array('type' => 'varchar(5)', 'default'=>'en_EN'),
     'bio' => array('type' => 'text', 'default'=>''),
 
-    'tags' => array('type' => 'text', 'default'=>NULL, 'Null'=>true),
+    'tags' => array('type' => 'text', 'default'=>NULL),
     'startingpage' => array('type'=>'varchar(64)', 'default'=>'feed/list'),
     'email_verified' => array('type' => 'int(11)', 'default'=>0),
     'verification_key' => array('type' => 'varchar(64)', 'default'=>''),
-    'preferences' => array('type' => 'varchar(255)', 'default'=>NULL, 'Null'=>true)
+    'preferences' => array('type' => 'varchar(255)', 'default'=>NULL)
 );
 
 $schema['rememberme'] = array(


TrystanLea:
Thanks @takkaria there seems to be an ongoing issue here that Im not getting to the bottom of.. as we've been adding and removing this multiple times, here's my last commit where I added it in to fix the same issue... https://github.com/emoncms/emoncms/commit/2c913bffed404e1893831e3a828b4e89af34bab0

Seems to be caused by differences in the installation environment somehow?

takkaria:
@TrystanLea so the problem is that the code in `dbschemasetup.php` is really inconsistent. I tried to explain what was going on in pull request #930 and also in [a comment](https://github.com/emoncms/emoncms/issues/791#issuecomment-406338727) on issue #791.

I imagine the reason this fix works for a new person but doesn't work for you is because you're not testing with a clean database which means that when you run the database schema update check, you are running a different code path to someone who is setting up from scratch (table creation vs table alteration).

I have just spent 20 minutes trying to write an explaination the bug and failing, because the code in the schema setup file is very difficult to follow.  Which is kind of the problem!  The logic for the NOT NULL constraints is different in the three different cases database definition alterations are made:

1. creating a table from scratch
2. creating a column on an existing table
3. altering the definition of an existing column

I already wrote an explanation of [why it fails in the first case](https://github.com/emoncms/emoncms/issues/791#issuecomment-406338727).  As far as I can tell, in the second case, in order to change a column to have a NOT NULL constraint, you'd have to have the 'Null' value in the database schema definition set to "NO".  And in the third case, to make it produce NOT NULL constraint, you'd have to have it set the schema definition 'Null' key to anything other than 'true'.  So it's quite inconsistent.

It's further complicated by the fact that you have to respecify the whole column spec when you run ALTER TABLE, not just the alterations.  At the moment cases 2 and 3 only attempt to respecify the bits of the column spec that have changed, which means that different EmonCMS instances with the same verison of EmonCMS are likely to have different database schemas depending on how many upgrades they've had.

I'd suggest taking a look at the pull request for #930, because I fixed all the issues I could find with this code.

TrystanLea:
Thankyou @takkaria I will prioritise looking at #930 thankyou for the reminder and for highlighting that there are inconsistencies that this sorts.

TrystanLea:
I've merged #930, can this pull request now be closed?

takkaria:
Yup!

Issue #985, opened 2018-08-21
Swipe action on mobile devices


glynhudson:
Currently, the Setup > Inputs / Feeds etc menu is small and difficult to use on mobile devices. Burger menus are already used in some Emoncms modules therefore cannot be used for the setup menu. 

One option could be to explore a swipe action in mobile devices to move between the Inputs / Feed pages, we could start by experimenting with: http://lyfeyaj.github.io/swipe/ 

First stage would implement a proof of concept the test how well it works before implementing fully.

emrysr:
I found when highlighting text in the new edit input that the page navigated to the feeds page!
This only happens during a "fast swipe", the same action as highlighting text in a textbox..

I should have disabled this feature for laptops and desktops (anything with a mouse pointer) so that text can still be highlighted without danger of incorrectly triggering the swipe action that navigates to another page.

@TrystanLea should I re-open this issue or create another?

Issue #990, opened 2018-08-22
Save user preference for expanded or contracted input/feed lists


TrystanLea:
There is a differing preference for expanded/contracted input/feeds lists, I wonder if we could save this preference in the new user preferences feature in the user module?

emrysr:
works well with issue #1021. as you swipe from one to the other the state of the accordion/dropdowns are remembered across page loads

pb66:
@TrystanLea and @emrysr 

Will this persist across logins or devices?

The preference to have all collapsed on first view (or not) is very different to remembering the current choice whilst swiping pages.

I think there still needs to be a default user setting, I would prefer that new pages always opened all collapsed, even if I did leave a panel or 2 open when I last visited.

PR #1021, opened 2018-09-18
added detection of swipe

emrysr:
fixes issue #985 

The "velocity" of the swipe is a little higher than default. This should stop accidental swipes - can be adjusted if required
> swipe faster if it seems not to work.

```
swipe = new Hammer.Swipe({ velocity: 1.1 }) // default 0.3
```
( [emrysr: emoncms/Theme/basic/theme.php L199](https://github.com/emrysr/emoncms/blob/feature/issue-985-swipe-between-feeds-and-inputs/Theme/basic/theme.php#L199) )

swipe-navigation
[screencast of swipe action navigating between pages]

Currently only goes between the feed list (/feed/list) and the input list (/input/view).

this will work a little slicker once issue #990 is done and the different views remember which groups have been expanded/contracted


TrystanLea:
nice work! hopefully adding the 20kb file to our already long list of js includes wont slow things down too much? merging for now, to evaluate over time

PR #1024, opened 2018-09-21
Feature/issue 1001 common units

emrysr:
I've attempted to do two different things with this pr

* standardize the units of measure across the different modules (allowing for further settings)
* have a simple system to edit config values (easy to read)
---
* ## Single place for units
the units of measure need to be available as an include in some cases and as a separate json object for others (ajax calls)
    * json version available as `/units.php`
    image
    * php version (array) available if you `include()` the file `/units.php` into your script.
    image

* ## Simple config
standardized config file format using [Toml](https://github.com/toml-lang/toml)
here's an example of the units section of the config file:
``` toml
[[units]]
short = "W"
long = "Watt"

[[units]]
short = "kWh"
long = "Kilowatt Hour"

[[units]]
short = "Wh"
long = "Watt-Hour"
```

This is then parsed as a PHP array and used as normal in PHP. The advantage over an INI or ENV file is that you can have hierarchy and data types in the config file - not just key=value pairs: eg:
1. arrays = ` [mysql.port] = '3306' `
1. comments = ` #this is a comment `
1. dates = `2018-09-20` is parsed as a date (might not be so useful yet??)
1. key/value pairs (just like an env file) = `max_timeout = 200`

The only drawback on this is that I've had to [install composer](https://www.digitalocean.com/community/tutorials/how-to-install-and-use-composer-on-ubuntu-14-04) to allow me to install the config file parser (which I probably would have to do if it was an ENV or INI file anyway)

so once you've fetched this branch you'll have to run `$ composer install` to install the dependencies.

I haven't fully finished this issue yet, as I'd need to implement the changes to read the config file in the other modules (didn't want to do the work until this pull request was accepted)

---

Using a pre agreed standard for config files would improve future integrations and (hopefully) simplify the config modification process

Let me know if you think this is worth finishing on the other modules.


emrysr:
other discussion regarding standardizing the config settings: #1010 
> Initial comit of hierarchical ini files to replace settings.php

emrysr:
.toml config file bad idea - gone with `Lib/units.php` as the file to include ( and edit ) that is the common location for the standard units of measure

PR #1040, opened 2018-09-23
Feature/issue 1038 move feed list buttons bottom

### LINKED ISSUE #1038: Feeds page buttons at both top and bottom of page
Should all the butttons be grouped together rather than split between the start and the end of the page?

Style wise the "create new virtual feed" button could be reduced to a "+" button with a more informative hover-over like the new "select all" etc. Likewise the "refresh feed size" could be reduced to refresh symbol. 

emrysr:
fix #1038 

There was just enough room for 2 more buttons in the mobile view: 
image

### DIFF
--- a/Lib/responsive-linked-tables.js
+++ b/Lib/responsive-linked-tables.js
@@ -13,7 +13,9 @@ function setExpandButtonState(state){
       .toggleClass('icon-resize-small', state)
       .toggleClass('icon-resize-full', !state)
     if(!$btn_expand.data('original-title')) $btn_expand.data('original-title', $btn_expand.attr('title'))
-    $btn_expand.attr('title', !state ? $btn_expand.data('alt-title') : $btn_expand.data('original-title'))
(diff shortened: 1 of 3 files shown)


emrysr:
As @pb66 's suggestion I've added a dropdown list of buttons for the mobile version:

image

not sure if this is as useful as the previous version, however there is now no wrap of buttons using smaller devices.

The desktop (wider screen) version still has the fixed toolbar at the top when scrolling down a longer list:
image

emrysr:
the desktop view now also has a separator between the buttons:
image

TrystanLea:
Sorry for the delay on this, I think we should keep as is. The dropdown menu makes the interface feel a bit more cumbersome. I think keeping the less used refresh and add virtual feed button at the bottom keeps things clean. Can I close the pull request so that we move forward as is for now?

TrystanLea:
PS: Thanks for your work putting this concept together @emrysr as we've discussed :)

PR #1046, opened 2018-09-24
Feature/issue 1033 timestamp in tooltip

### LINKED ISSUE #1033: Seconds are not displayed in the hover-over info box
In the hover-over info box on the feeds page, the PHPFINA feed start time is currently displayed as a date time string (see #1032) but it is missing the vital seconds bit and only displays whole seconds. can we have seconds please?  (assuming the date/time string to to stay)

emrysr:
fixes #1033 
The timestamp now shows in the feed list tooltips:
image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,5 @@ Modules/*/
 Theme/*
 Theme/*/
 !Theme/basic/
+
+/vendor/
--- a/Modules/feed/Views/feedlist_view_v2.php
+++ b/Modules/feed/Views/feedlist_view_v2.php
@@ -398,10 +398,15 @@ body{padding:0!important}
               
               for (var feed in nodes[node]) {
                   var feedid = nodes[node][feed].id;
-                  var row_title = ["Feed ID: "+feedid,
-                                    "Feed Interval: "+(nodes[node][feed].interval||'')+'s',
-                                    "Feed Start Time: "+format_time(nodes[node][feed].start_time,'LLLL')
-                  ].join("\n")
+                  var title_lines = ["Feed ID: "+feedid,
+                                    "Feed Interval: "+(nodes[node][feed].interval||'')+'s'
+                  ]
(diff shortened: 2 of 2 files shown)


Paul-Reed:
Wouldn't it have been better to display seconds as in 23:57:03 instead of an epoch?
I'm assuming the date time will be local time, whilst won't the epoch be UTC? which could cause confusion.

pb66:
> Wouldn't it have been better to display seconds as in 23:57:03 instead of an epoch?
I'm assuming the date time will be local time, whilst won't the epoch be UTC? which could cause confusion.

IMO the unix timestamp is of paramount importance as that literally IS what the start time meta field contains, that actual 10 digit value. All feed data is UTC, any local time representation is dependent on timezone of user (browser or user settings?) so it could change if settings change, where as the timestamp will not.

Plus if you are wanting to know the feed start time, it is most likely to use in a API or to debug something so the timestamp is possibly more useful.

In my comments in the forum thread I said we need the timestamp but both would be nice/useful, but IMO the timedat string with no TZ data is open to confusion, but I think both is best, but yes if the localtime is displayed then it really needs seconds to be useful.

Again IMO, the layout should maybe be 

> Feed start time: 1234567890
> (Friday, 13th December 2018, 23:59:59)

because the feed start time meta is the timestamp and then the longer string in the line below to reduce the overall  length of the line (width of the box) and in brackets would help suggest it is a representation of the preceding timestamp.


@emrysr, good work, I see you have made loads of changes and there are an assortment of screenshots, they all look good, hopefully @TrystanLea will merge soon so that we can assess them all together, so I won't reply to all the issues, so a big blanket thank you from me here. 👍

Paul-Reed:
> Again IMO, the layout should maybe be
> 
> Feed start time: 1234567890
> (Friday, 13th December 2018, 23:59:59)

Yes, that's a fair compromise!

Pul

PR #1047, opened 2018-09-24
Feature/issue 1029 feed tag in tooltip

### LINKED ISSUE #1029: Feed page hover-over info not clear as to which feed it belongs to
image

Could we include the feed tag:name at the top of the hover-over info box so it is more obvious which feed the info belongs to? Perhaps in bold so it works like a title eg

> **emontx : voltage**
> Feed ID: 77
> Feed Interval: 5s
> Feed Start Time: Monday, 5 Febuary 2018 08:12

emrysr:
fix #1029 
added <html> to the tooltip to allow for better styling of the information. now uses the bootstrap tooltip plugin to display the details

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,5 @@ Modules/*/
 Theme/*
 Theme/*/
 !Theme/basic/
+
+/vendor/
--- a/Modules/feed/Views/feedlist_view_v2.php
+++ b/Modules/feed/Views/feedlist_view_v2.php
@@ -328,7 +328,7 @@ body{padding:0!important}
   var feed_engines = ['MYSQL','TIMESTORE','PHPTIMESERIES','GRAPHITE','PHPTIMESTORE','PHPFINA','PHPFIWA','VIRTUAL','MEMORY','REDISBUFFER','CASSANDRA'];
 // auto refresh
   update();
-//   setInterval(update,5000);
+  setInterval(update,5000);
   var firstLoad = true;
   function update() 
   {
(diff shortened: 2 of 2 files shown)


TrystanLea:
Thanks @emrysr, is it possible to have the tooltip appear at the mouse location rather than the middle of the page? and is it possible for the tooltip to appear after a delay rather than always show?

emrysr:
@TrystanLea, I've rebuilt the tooltip fuctionality to better adapt the the mouse position.
Currently the delay timings are:
delay before = 500ms
delay after = 100ms

I can change them If you'd rather a longer delay.

TrystanLea:
Thanks @emrysr

emrysr:
If you like this feature I can re-work the other tooltip related issues to use the same code - that should help you with merging the branches

Paul-Reed:
@emrysr The tooltips are a great addition, but I'm finding them a bit intrusive because the hotspot extends the full width of the table, meaning that just about wherever the mouse pointer is parked, it displays the tooltip.

Would it perhaps be better if the tooltip hotspot area was confined to the 'name' column?

The tooltip information displayed is more in context to the feed 'name', and by extending it across the entire width gives the user expectation that the context would change as the pointer is moved across the other columns;
e.g. the 'size' column may display exact file size, data file location etc.

What do you think?

Paul-Reed:
....also, I assume that this error which has occurred a few times tonight is connected to this function -

uncaught

emrysr:
hi @Paul-Reed. I've fixed the error in the screenshot above... I wasn't able to re-create it in Firefox, however I assume that not correctly defining the variable in the global scope caused the other functions to throw the error. Please comment if it still becomes an error once this pr is merged.

Yes, the tooltip "target" could be restricted to a single column. I'll re-visit this again...

Once all the [recent issues (&pull requests) related to the feedlist](https://community.openenergymonitor.org/t/thoughts-on-the-new-input-and-feed-page-layouts/8583/11?u=emrys) are done, I'll have a better understanding of what to expect as the final layout before I add further changes.

Paul-Reed:
Thanks @emrysr 
I've been testing on google chrome, not Firefox, so yes, I'll let you know if it persists.

PR #1059, opened 2018-10-02
fixed the mqtt settings being saved to incorrect array

emrysr:
> # not fully tested !
> ## not ready for pulling into master!! 

@TrystanLea  I assume that these settings are not being used correctly or this issue would have been raised much sooner?...

### DIFF
--- a/process_settings.php
+++ b/process_settings.php
@@ -52,11 +52,11 @@ if(file_exists(dirname(__FILE__)."/settings.php"))
         if (!isset($mqtt_server)) $mqtt_server = array();
         if (isset($_ENV["EMONCMS_MQTT_ENABLED"]))  $mqtt_enabled = $_ENV["EMONCMS_MQTT_ENABLED"] === 'true';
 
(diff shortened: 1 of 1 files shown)


emrysr:
the array `$mqtt_server` is also created in these files so the incorrect settings would be overwritten
- /settings.php
- /emonpi.settings.php

Paul-Reed:
It would be good to increment the $config_file_version in settings.php whenever an important change is introduced in settings.
This would force users to update their settings file.

borpin:
>It would be good to increment the $config_file_version in settings.php whenever an important change is introduced in settings.
This would force users to update their settings file.

Even better would be a hierarchical system that would **not** require the whole setting file to be regenerated every time a change was made. #923 

I think the concern around having to do that is a block on users upgrading.

Paul-Reed:
Yes. Have you seen https://github.com/emoncms/emoncms/pull/1010#issuecomment-426236385

PR #1061, opened 2018-10-05
added CORS complient preflight headers

emrysr:
while testing the `Authorization: Bearer APIKEY` as per the emoncms `/feed/api` in the browser I got issues with the [CORS preflight request](https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request) not returning the expected headers

**used during development the remote calls to the api**
before this is pulled in, we'll need to discuss the pros and cons of the `allow-origin` header being set to `*` (all). 
``` 
'Access-Control-Allow-Origin: *'
```
might be better to be set to `openenergymonitor.org` if that is where the requests will be originating from?

These are the HTTP headers I'm returning when a HTTP `OPTIONS` request  is made to the API. 
_(my changes don't effect GET or POST requests)_
```
            header('Access-Control-Allow-Origin: *');
            header('Access-Control-Allow-Headers: Authorization');
            header('Access-Control-Allow-Methods: GET');
```

this will not be required until the branch i'm working on now is pulled in and requests are made to a different domain

### DIFF
--- a/route.php
+++ b/route.php
@@ -132,6 +132,13 @@ class Route
         } elseif(in_array($requestMethod, array('POST', 'DELETE', 'PUT'))) {
             $this->method = $requestMethod;
         }
+        if($requestMethod === 'OPTIONS') {
+            // "CORS PREFLIGHT REQUESTS" EXPECT THESE HEADERS. no content required
+            header('Access-Control-Allow-Origin: *');
+            header('Access-Control-Allow-Headers: Authorization');
+            header('Access-Control-Allow-Methods: GET');
+            exit();
+        }
     }
 
     /**
--- a/route.php
+++ b/route.php
@@ -131,8 +131,7 @@ class Route
             $this->method = 'PUT';
         } elseif(in_array($requestMethod, array('POST', 'DELETE', 'PUT'))) {
             $this->method = $requestMethod;
-        }
-span class="pl-s1">         elseif($requestMethod === 'OPTIONS') {
+        } elseif($requestMethod === 'OPTIONS') {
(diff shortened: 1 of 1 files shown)


emrysr:
@borpin yes the api call works with the authorization header using curl (as per the docs), however while attempting to use that header in a browser (JS XMLHttpRequest) they now send a OPTIONS request first (as per the CORS starndard) to see what methods are allowed for the desired request... not sure why? must be a reason - I need to read up on the implications of using this method to request data.

still testing if this is a required feature or not. I might implement a different request to the local server that handles external requests with PHP instead of crossing domains in the browser.

TrystanLea:
Can this be an 'else' here?: https://github.com/emoncms/emoncms/pull/1061/files#diff-7aa76a900040c1bef43af8c5945d17f1R135

route.php
```diff
@@ -132,6 +132,13 @@ public function decode($q, $documentRoot, $requestMethod)
         } elseif(in_array($requestMethod, array('POST', 'DELETE', 'PUT'))) {
             $this->method = $requestMethod;
         }
+        if($requestMethod === 'OPTIONS') {
```
emrysr: @TrystanLea - just trying the new "insert a suggestion" feature.
https://github.com/emoncms/emoncms/pull/1061

Yep I can change this to an else - I'll try this github feature to add the changes

emrysr:
```suggestion
        elseif($requestMethod === 'OPTIONS') {
```

emrysr: added html to my edits of a php file!!
```
span class="pl-s1"> 
```

use this new github feature with caution! (still beta)

PR #1072, opened 2018-10-18
Improved Continuous Integration and Test suite

cagabi:
As part of the Group module and MyHomeEnergyPlanner we have started writing behavioural tests. On top it we have implemented Continuous Integration for MHEP. I have modified it to fit emonCMS.

This is a much better built than the one there was before which basically did nothing (1+1=2).

With this new .travis.yml we:
 - set up the server twice (for PHP 5.6 and PHP 7.2), 
 - do the whole emonCMS installation in both cases 
 - set up test suite (based on Jasmine, Selenium and WebDriverIO)
 - run some basic tests: 
     - first we create an admin user and update the database, 
     - then we just login.

These  tests are the bare minimum, I know, but it's a start. They are very easy to write once you familiarise with Jasmine and WebDriverIO


glynhudson:
Nice work @cagabi 👍

glynhudson:
Hi @cagabi ,

The Travis builds seem to be passing for Chrome but failing on Firefox:

https://travis-ci.org/emoncms/emoncms/builds/445424453?utm_source=github_status&utm_medium=notification

Did it pass when you were testing this PR? 

https://travis-ci.org/emoncms/emoncms/jobs/446269925

cagabi:
I have done a new PR that fixes this problem see #1099

What has happened: Travis was using the latest Firefox and the problem was that firefox 63.0 has just been released. There are some problems with this version and the gecko web driver (the bit of software that works as a bridge between our tests and firefox). I have changed the travis built to use firefox 60.0 instead of latest, at some point I will need to change it again to use the latest version of firefox.

PR #1073, opened 2018-10-19
removed the dependance on hammer js on views that need it

### LINKED ISSUE #1055: Mobile view doesn't scroll up or down
Using current emoncms & device module 'master' git versions, and viewing on Android mobile device.
Both the inputs & feeds pages will not scroll up or down, so you cannot see all of the pages.
It appears that vertical scrolling is disabled.

emrysr:
fix horizontal scrolling issue on admin page
fixes #1055


glynhudson:
Thanks that's fixed it

PR #1082, opened 2018-10-23
 added arrow to "configure your devices" alert for first timers

emrysr:
While I [was working on the device module](https://github.com/emoncms/device/pull/30) I noticed the `"Configure your device here"` message could direct the user better with an arrow pointing at the correct button:

image

the message disappears once a device has been initialized and is saving to a feed

### DIFF
--- a/Modules/input/Views/device_view.php
+++ b/Modules/input/Views/device_view.php
@@ -182,7 +182,7 @@ function noProcessNotification(devices){
         }
     }
     if(processList.length<1 && Object.keys(devices).length > 0){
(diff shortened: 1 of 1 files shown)


glynhudson:
That's a really good idea, I also found the white spanner to configure the device to not be very obvious. I found I was drawn to the more prominent black spanners below. 

How about going a step further and making the configure whole device spanner icon a black spanner and making the individual spanners below white. This would focus a users attention on the Configure All spanner which is what most users will want to use for the easiest setup. 

We could consider changing the top spanner to a wizard icon since the device module config can be thought of as a setup wizard e.g https://fontawesome.com/icons/magic?style=solid&from=io

Issue #1066, opened 2018-10-15
Difficulties renaming feeds.


pb66:
Using latest emoncms 9.8.31, indexedinputs branch "device-module integration beta" off.

I had created a feed with a slight typo, instaed of "GrossActiveEnergy" I had typed "grossActiveEnergy", when I tried to correct via the edit button on the feeds page, I kept getting a message telling me feed name and tag combinations must be unique. It seems emoncms is comparing the revised feed name with the existing list of feed names  including it's pre-edit version of itself, and that comparison is case insensitive so  "G" == "g".

I managed to get around this by editing the name to "XrossActiveEnergy" and then edit again to "GrossActiveEnergy".

Easy to recreate, here is the result of attempting to change back to "grossActiveEnergy" from "GrossActiveEnergy"

![image](https://user-images.githubusercontent.com/5606042/46965074-bb91b280-d0a1-11e8-9de0-4911f4bf8f58.png)

glynhudson:
>Using latest emoncms 9.8.31, indexedinputs branch "device-module integration beta" off.

I'm assuming you mean 9.9.3?

PR #1087, opened 2018-10-26
fix issue with renaming feeds not being case sensitive

emrysr:
fix issue #1066 
- feed->exists_tag_name() is now case sensitive
- clicking save now only effects the selected entries


emrysr:
@TrystanLea - do you think a case sensitive check for the tag name is also required?
This pull request ensures that the unique check in the database is case sensitive for both feed names and tags. Let me know if tag name check should be put back to being case insensitive.

glynhudson:
Spoke to Trystan, agree that tags should also be case sensitive. Thanks for the fix :+1:

PR #1094, opened 2018-10-28
redis: support db selection via dbnum config

karlp:
Prefixes for keys are handy, but redis supports multiple separate dbs entirely,
via "select".  Expose that as a configuration setting.

Signed-off-by: Karl Palsson <karlp@tweak.net.au>

--

I didn't find any real place to add documentation for this, there's nothing for the auth option either?

### DIFF
--- a/default.settings.php
+++ b/default.settings.php
@@ -15,6 +15,7 @@
     $redis_server = array( 'host'   => 'localhost',
                            'port'   => 6379,
                            'auth'   => '',
+                           'dbnum'   => '',
                            'prefix' => 'emoncms');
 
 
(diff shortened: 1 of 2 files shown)


karlp:
travis failures for this don't appear related? is that a known current issue with your travis setup?

TrystanLea:
Thanks @karlp  looks great!

borpin:
Can I point out that setting a precedent for checking for an empty setting (because it has not been incorporated from the default into the users main `settings.php`) file is a bad thing

https://github.com/emoncms/emoncms/blob/1da0828d790ddfef7c1ad9852125d5a85c7af0a9/index.php#L42

#1010 should be investigated further to avoid the need for this.

PR #1099, opened 2018-10-30
WP - Travis built: Firefox version for downgraded from latest to 60.0

cagabi:
fixes #1072 

Travis was using the latest Firefox and the problem was that firefox 63.0 has just been released. There are some problems with this version and the gecko web driver (the bit of software that works as a bridge between our tests and firefox). I have changed the travis built to use firefox 60.0, at some point I will need to change it again to use the latest version of firefox.


cagabi:
Wait, my commit was passing the travis check https://travis-ci.com/carboncoop/emoncms/builds/89673208
But now it shows as not passing.
Still having a look

cagabi:
all checks passed now

glynhudson:
Great, thanks.

Issue #1105, opened 2018-11-02
MQTT input does not handle certain characters in topic names


borpin:
A topic of `emon/xiaomi/4c:65:a8:d9:59:1a/battery` results in a new input being created with every data item via MQTT.

ref https://community.openenergymonitor.org/t/delete-multiple-input-nodes-100s/9034

borpin:
Merge of #1156 solves the specific case of colons, but a check is required to see if there are other characters that could cause issues.

TrystanLea:
Here's the result with your example above and the latest topic mapping: 

![image](https://user-images.githubusercontent.com/503186/77153342-7f7e3080-6a91-11ea-8d66-33b8185f05a3.png)

Then with: 

    emon/xiaomi/4c,65*a8/d9^59;1a/battery

![image](https://user-images.githubusercontent.com/503186/77153696-22cf4580-6a92-11ea-8187-da8ad09a2e1e.png)

borpin:
Ah Ok great. Cheers.

PR #1106, opened 2018-11-02
Quick fix to colons in MQTT Topic

borpin:
Quick fix to remove colons from topic.  Suggest it is expanded to include all characters inputs should not use.

Issue #1105


borpin:
> Not sure that they should be removed, as the colon forms part of the unique topic name.

Whilst true, emoncms strips them out when creating the input node / name.  This then creates the multitude of inputs I am trying to delete 😄

borpin:
Mucked up my Git fork. I will resubmit.

PR #1117, opened 2018-11-16
added more detail to the window.onerror event

### LINKED ISSUE #1113: javascript error caught with alert would be more useful with a line number

emrysr:
fix #1113 

[screenshot of new alert contents]
image


chaveiro:
This is nice, how about add a link to report to developers (or github) with some debug context?

PR #1125, opened 2018-11-19
No error or output shown when json_encode() returns an error.

fcauwe:
Bugfix for: "No error or output shown when json_encode() returns an error."
This error causes a lot of searching when debugging, for example when there is an invalid character in a string.


TrystanLea:
Thanks @fcauwe looks great!

PR #1134, opened 2018-12-01
show status of running services on admin page

TrystanLea:
image

### DIFF
--- a/Modules/admin/admin_main_view.php
+++ b/Modules/admin/admin_main_view.php
@@ -9,7 +9,9 @@
     $db = $result->fetch_array();
 
     @list($system, $host, $kernel) = preg_split('/[\s,]+/', php_uname('a'), 5);
+    @exec('ps ax | grep emonhub.py | grep -v grep', $emonhubproc);
     @exec('ps ax | grep feedwriter.php | grep -v grep', $feedwriterproc);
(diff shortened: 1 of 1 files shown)


borpin:
Would have been nice to have some discussion.  I'd have suggested that you need the output to tell you more for debugging purposes especially with SSH disabled.

Is it only those services?  service_runner?

TrystanLea:
Sure, this is not complete yet, I was thinking the same regarding status output etc

Paul-Reed:
Does it report the emonhub status for both versions of emonhub?
i.e. the emonpi version & pb's version as used in self builds.

TrystanLea:
It tests for emonhub.py so should pick up both, https://github.com/emoncms/emoncms/commit/8540e01f3c402f4dc2f178f69c867774eb0ae260#diff-adef21b62b8a6ff7eecf2e8ebf361ab6R15

Paul-Reed:
I've been trying the PR tonight, and found the following;
**Emonhub installed as per build guide -**
Correctly detects if running or not, but Start/Stop/Restart/Kill has no effect.
**Mqtt_input installed as per build guide -**
Start/Stop/Restart/Kill has no effect.

glynhudson:
Awesome! Nice work. That looks like a really useful addition. I'll test on Monday

Issue #1135, opened 2018-12-02
emoncms does not work behind reverse proxy


imro2:
Emoncms breaks when running behind reverse proxy. In my case I am using traefik. The problem is in the way the base url ($path) is constructed in core.php. It uses wrong header HTTP_X_FORWARDED_SERVER which determines the host name of the proxy https://httpd.apache.org/docs/2.4/mod/mod_proxy.html#x-headers. Changing it to HTTP_X_FORWARDED_HOST fixes the problem.

PR #1136, opened 2018-12-05
Corrected reverse proxy support

imro2:
This addresses issue #1135 where putting emoncms behind reverse proxy breaks the website. It is due to incorrect header (HTTP_X_FORWARDED_SERVER) being used to identify the real host URL. The correct header is HTTP_X_FORWARDED_HOST.

### DIFF
--- a/core.php
+++ b/core.php
@@ -29,8 +29,8 @@ function get_application_path()
         $proto = "https";
     }
 
-    if( isset( $_SERVER['HTTP_X_FORWARDED_SERVER'] ))
-        $path = dirname("$proto://" . server('HTTP_X_FORWARDED_SERVER') . server('SCRIPT_NAME')) . "/";
+    if( isset( $_SERVER['HTTP_X_FORWARDED_HOST'] ))
+        $path = dirname("$proto://" . server('HTTP_X_FORWARDED_HOST') . server('SCRIPT_NAME')) . "/";
     else
         $path = dirname("$proto://" . server('HTTP_HOST') . server('SCRIPT_NAME')) . "/";
 


TrystanLea:
Hello @imro2 thanks for this, is this header change compatible with different versions of php and environments?

imro2:
@TrystanLea, I don't think this is php specific. The availability of any specific variable is server depended according to PHP documentation:
http://php.net/manual/en/reserved.variables.server.php
> The entries in this array are created by the web server. There is no guarantee that every web server will provide any of these; servers may omit some, or provide others not listed here. That said, a large number of these variables are accounted for in the » [CGI/1.1 specification](http://www.faqs.org/rfcs/rfc3875), so you should be able to expect those. 

From the linked [RFC3875](http://www.faqs.org/rfcs/rfc3875) 4.1.18:
> Meta-variables with names beginning with "HTTP_" contain values read from the client request header fields, if the protocol used is HTTP. The HTTP header field name is converted to upper case, has all occurrences of "-" replaced with "_" and has "HTTP_" prepended to give the meta-variable name. 

HTTP_X_FORWARDED_SERVER and HTTP_X_FORWARDED_HOST hold values for X-Forwarded-Server and X-Forwarded-Host respectively.

The only documentation I was able to find regarding X-Forwarded-Server is from apache mod_proxy https://httpd.apache.org/docs/2.4/mod/mod_proxy.html#x-headers. In there it states that:

> X-Forwarded-Host
>     The original host requested by the client in the Host HTTP request header.
> X-Forwarded-Server
>     The hostname of the proxy server.

There is plenty of documentation regarding X-Forwarded-Host. For example https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Host:

> The X-Forwarded-Host (XFH) header is a de-facto standard header for identifying the original host requested by the client in the Host HTTP request header.

I don't have the ability to test different environments. I guess, if you want to be extra cautious, you could check with @thunderace, who seems to have originally added the HTTP_X_FORWARDED_SERVER in commit 3c6baf1b161ceb0c46cad928e83c0708b0f73ca6

TrystanLea:
thanks @imro2, @thunderace would you be able to comment?

TrystanLea:
could we check for both perhaps?

timpur:
I can confirm that it should be `X-Forwarded-Host`

https://docs.traefik.io/v1.3/

image


From my testing (Reduced some info):
```
Host: whoami.****.****.local
****
X-Forwarded-For: ****
X-Forwarded-Host: whoami.****.****.local
X-Forwarded-Port: 443
X-Forwarded-Proto: https
X-Forwarded-Server: 5a43264ba2be
X-Real-Ip: ****
```

Kinda need this ... any chance this will go in soon ?

glynhudson:
@emrysr any chance you could test this. Happy to merge if behaviour is as expected. Thanks for your help

glynhudson:
Thanks for the help with testing this, sorry for the delay merging

PR #1140, opened 2018-12-18
fix padding spikes where input feed has null value

TrystanLea:
This is to fix the issue highlighted in the forum thread here: https://community.openenergymonitor.org/t/getting-spikes-in-emoncms-graph/9494/26

@chaveiro for some reason the use of static $lastvalue_static_cache did not store the last value as expected. Changing to a private variable array seems to work much better. Can you see any issue with this approach?

### DIFF
--- a/Modules/feed/engine/PHPFina.php
+++ b/Modules/feed/engine/PHPFina.php
@@ -10,6 +10,7 @@ class PHPFina implements engine_methods
     private $dir = "/var/lib/phpfina/";
     private $log;
     private $writebuffer = array();
+    private $lastvalue_cache = array();
     private $maxpadding = 3153600; // 1 year @ 10s
 
     /**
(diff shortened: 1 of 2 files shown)


chaveiro:
I've not run the code but think what made the trick was what that you added L718 to save last value , it should work the same with the static variable ? https://github.com/emoncms/emoncms/pull/1140/files#diff-5b2bae4bf00f1938393368515a475c88L718

TrystanLea:
Thanks @chaveiro, I tried adding L718 before switching to the private variable but for some reason it did not work. When I logged the content of $lastvalue_static_cache it only seemed to store a small number of feeds (3-4), whereas the private variable on my system quickly built up to 20 feeds+

Is there any reason not to use a private variable in this context?

PR #1143, opened 2018-12-18
Feature/gzip

SeanDS:
This adds support to decompress bulk data received over HTTP from a remote `emonhub` client. If the `c` URL parameter is specified, it will attempt to decompress the `data` payload using `zlib`. It is intended to be merged alongside the [corresponding pull request](https://github.com/openenergymonitor/emonhub/pull/70) for `emonhub`.


TrystanLea:
Thanks @SeanDS nice simple pull request for the additional benefit!

PR #1146, opened 2018-12-29
Use override conf instead of editing system file.

bwduncan:
/etc/systemd/system/mysqld.service is a symlink to /lib/systemd/system/mariadb.service which is replaced by package updates. Use systemctl edit to create an appropriate /etc/systemd/system/mariadb.service.d/override.conf instead.


bwduncan:
Thanks @borpin 

Yes you're right, according to the man page:

> After the units have been edited, systemd configuration is reloaded (in a way that is equivalent to daemon-reload).

In any case I think these instructions are now really only used in reference from [the SD card build instructions ](https://github.com/openenergymonitor/emonpi/blob/master/docs/SD-card-build.md) since the root fs isn't readonly anymore, so anyone building a new SD card will certainly be rebooting.

Issue #1141, opened 2018-12-18
remove scale and delete options from editrealtime


TrystanLea:
remove scale and delete options from editrealtime, features do not currently have end point API's and so it would be good to hide these options until fixed to avoid confusion.

emrysr:
@TrystanLea - you mentioned that you had also worked on this. I have un-assigned myself and assigned you to this one. Can this now be close?

PR #1147, opened 2019-01-03
datafeed edition for PHPFina engine

alexandrecuer:
This PR makes it functional to edit a datarange of a PHPFINA feed via the editrealtime visualisation :
- You can multiply the values in the window by a float or a fraction
- You can replace the values in the window by NAN if they are erronous due to some communication problems
- You can convert all the values in the window to absolute values

cf the following issue :
https://github.com/emoncms/emoncms/issues/1141
and this PR :
https://github.com/emoncms/emoncms/pull/1142 

working only for the PHPFINA engine

is it what you were looking for when designing the editrealtime visualisation ? or yu want the process to inject first the data in Redis, the feewriter service being responsible for transferring everything into the feeds ?

Alex

### DIFF
--- a/Modules/feed/engine/PHPFina.php
+++ b/Modules/feed/engine/PHPFina.php
@@ -247,6 +247,104 @@ class PHPFina implements engine_methods
     }
 
     /**
+     * scale a portion of a feed
+     * added by Alexandre CUER - january 2019 
+     
+     * @param integer $feedid The id of the feed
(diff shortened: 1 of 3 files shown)


TrystanLea:
Thanks @alexandrecuer this looks great. Im happy with the direct edit approach for this kind of operation as it will usually be applied for data on the disk already. 
Nice idea to allow setting as NAN and abs.

PR #1152, opened 2019-01-14
documentation update for get_data() method in all Engine

alexandrecuer:
@TrystanLea 
This is just a documentation PR, following my previous PR, in order to be fully clean
cf  https://github.com/emoncms/emoncms/pull/1148

It removes documentation duplicates for the get_data() method...hope this is OK for you....

I've kept the initial formulation ``@param integer $limitinterval Limit datapoints returned to this value (used by some engines)`` in the shared_helper.php
I've introduced the new formulation in the PHPTimeSeries.php
Everywhere a var is not implemented, I made a specific mention

compatible with phpdocumentor standards...

Alex


TrystanLea:
Thanks @alexandrecuer looks great

PR #1153, opened 2019-01-15
hidden uninstalled services

### LINKED ISSUE #1150: Background colours for service status list on admin page
I've merged pull request: https://github.com/emoncms/emoncms/pull/1134

@glynhudson @borpin @Paul-Reed the services all show up as red if not running, which is fine for emonpi/emonbase but may be confusing for installs that dont use these services as they are all optional.

I could remove the background colouring if $allow_emonpi_admin is true perhaps? do you have any preferences?

I removed the start,stop,restart and kill buttons for now as they are not yet implemented and I have not yet added any access to logs, which would be great to add in future.

emrysr:
showing green if  installed and running
showing red if installed and not running

fix #1150 

image

thanks @borpin for the `systemctl` command example in issue #1150

### DIFF
--- a/Modules/admin/admin_main_view.php
+++ b/Modules/admin/admin_main_view.php
@@ -3,16 +3,42 @@
   // Retrieve server information
   $system = system_information();
 
+  /**
+   * get running status of service
+   *
+   * @param string $name
+   * @return mixed true == running | false == stopped | null == not installed
+   */
+  function getServiceStatus($name) {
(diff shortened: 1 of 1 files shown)


borpin:
No bother. I'd use the actual text returned as that tell you more when debugging.  Why isn't it running? Has it failed for instance?

Also the "Service is" text is actually redundant.

I'd display it with the text returned for these elenments - 

**LoadState | ActiveState | SubState**

PR #1156, opened 2019-01-15
Replace colons in MQTT Topics

borpin:
Quick fix to replace colons in MQTT topic names as EmonCMS cannot handle them in node names.

This should be reviewed and any other characters that are legal for MQTT topics, that EmonCMS cannot handle in node names, also replaced.

Ref #1106 that I accidentally closed
Issue #1105


TrystanLea:
Thanks @borpin

Issue #1069, opened 2018-10-16
Use of user "pi" in service-runner


pb66:
A good point was raised in the ["Added systemd based service-runner in python"](https://github.com/emoncms/emoncms/pull/1025#issuecomment-429472440) PR shortly before it was merged, I've opened this issue so the point isn't lost with the closed PR thread.

Use of user "pi" in the service unit was carried over from the old init script and it is specific to using a raspberry pi with the default user only.

The generic service-runner should not really be run as user "pi".

greeebs:
IMHO running it as root by default is also the wrong answer. Remember that this script takes whatever is on the unauthenticated "service-runner" redis key and then _runs it_. 

Hmm, I just checked the update script in the emonpi repo (which is what started the need for service-runner in the first place):
https://github.com/openenergymonitor/emonpi/blob/master/service-runner-update.sh
sudo is all over the place in that script so maybe root is the right answer for an emonpi?

What about other self installs though?

What are the implications on EmonCMS if a specific user is created for service-runner? What permissions are actually required for it to do its thing - are they just file permissions? If so, should we create a emonAdmin group and make sure all of the relevant files and folders are `chgrp emonAdmin` and `chmod g+w`, then add the service-runner user to that group?

TrystanLea:
Noting this to review as part of generic target OS installation compatibility, to come back to after emonSD build script release.

borpin:
Could you add a label for those Issues you have 'Noted' please @TrystanLea?

TrystanLea:
Pretty sure this issue was fixed, the user can be set using:

    sudo systemctl edit service-runner.service

See:https://github.com/emoncms/emoncms/blob/master/scripts/services/install-service-runner-update.md

PR #1165, opened 2019-01-18
Tidy feedwriter, service-runner and mqtt_input services

borpin:
Changes made for sake of consistency (mostly).

Service-runner
* remove user Pi (#1069 refers). Document and comments updated.
* Restart on failure (same as feedwriter)
* documentation link updated
* Update instructions to uninstall/disable old service-runner

feedwriter
* had a 'wants' entry of `redis.service` corrected to `redis-server.service`
* documentation updated.

### DIFF
--- a/scripts/services/service-runner/service-runner.service
+++ b/scripts/services/service-runner/service-runner.service
@@ -47,9 +47,10 @@ Documentation=https://github.com/emoncms/emoncms/blob/master/scripts/services/in
 [Service]
 Type=idle
 ExecStart=/usr/bin/python /var/www/emoncms/scripts/services/service-runner/service-runner.py
+User=pi
 
 # Restart script if stopped on a failure. Will not restart if not configured correctly
-Restart=on-failure
+Restart=always
 # Wait 30s before restart
 RestartSec=30s
 
--- a/scripts/services/service-runner/service-runner.service
+++ b/scripts/services/service-runner/service-runner.service
@@ -1,9 +1,5 @@
 # Systemd unit file for mqtt input script
 
-# ***** NOTE: RUNS AS USER "pi" BY DEFAULT *****
-#  If running on a non-Raspbian environment, change "User=pi" in the [Service] section
-#  to the user of your choice (user must exist and should be the "emoncms" admin account)
-
 # INSTALL:
 # sudo ln -s /var/www/emoncms/scripts/services/service-runner/service-runner.service /lib/systemd/system
 
@@ -41,7 +37,7 @@ Description=Emoncms service-runner Input Script
 Wants=redis-server.service
 After=redis-server.service
 StartLimitIntervalSec=5
-#Documentation=https://github.com/emoncms/emoncms/blob/master/docs/service-runner.md
+Documentation=https://github.com/emoncms/emoncms/blob/master/scripts/services/install-service-runner-update.md
 
 # Uncomment this line to use a dedicated log file for StdOut and StdErr.
 # NOTE: only works in systemd v236+
@@ -51,11 +47,10 @@ StartLimitIntervalSec=5
 [Service]
 Type=idle
 ExecStart=/usr/bin/python /var/www/emoncms/scripts/services/service-runner/service-runner.py
-User=pi
 
-# Restart script if stopped
-Restart=always
-# Wait 60s before restart
+# Restart script if stopped on a failure. Will not restart if not configured correctly
+Restart=on-failure
+# Wait 30s before restart
 RestartSec=30s
 
 # Tag things in the log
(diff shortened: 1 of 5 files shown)


borpin:
Ref #1069 for question on the 'user' in service-runner.

TrystanLea:
Thanks @borpin @greeebs 

Looking through the pull request I cant see the line where you remove the user pi, I can see the removal of the comments though?

I was building another service yesterday and noticed that the script did not restart after causing a deliberate exit with "die;" with the service setting set to on-failure. I would have expected it to restart? Is always safer? 

We want this to restart in all cases, but not create duplicates of itself..?

borpin:
>Looking through the pull request I cant see the line where you remove the user pi, I can see the removal of the comments though?

Old file - line 54.

> I was building another service yesterday and noticed that the script did not restart after causing a deliberate exit with "die;"

That is the expected behaviour as you explicitly told it to stop.  If you tell it to die, do you really want it to just start up again?

greeebs:
>>I was building another service yesterday and noticed that the script did not restart after causing a deliberate exit with "die;"

>That is the expected behaviour as you explicitly told it to stop. If you tell it to die, do you really want it to just start up again?

I think @TrystanLea means the service is terminating due to an exception or other "failure" triggered by the "die;" call. Most people would call that "a failure".
It probably depends on the exit code seen by systemd. Usually "Normal exit" is identified by an exit code of 0 and non-normal is anything else.

According to [the php `die;` function page](http://php.net/manual/en/function.die.php) `die(); is equivalent to `exit();` and returns 0 - so not recognised as a failure.

See the second user contributed note in that link for an alternative method of "crashing".

greeebs:
Yep, that is exactly what I'm talking about - and a clean exit code is "0".

So the behaviour of "on-failure" is probably appropriate for feedwriter because the backup module needs to stop it. I question if that is the case for service-runner though. Is there any situation where you would want to shut it down manually (as opposed to restarting it)? The original service-runner ran indefinitely but was checked every minute (via cron) and restarted if it had died.

borpin:
>Is there any situation where you would want to shut it down manually

More to the point, if someone stops it (in the future) they would expect it to stay stopped which would be the correct behaviour I'd suggest.

greeebs:
> [edit] Just to be clear, I'm learning about systemd as are most folk, so I'm not being dogmatic about it. It is just it seems that often the directives are being misused by all sorts of different projects simply by copying what has been done elsewhere and not fully appreciating the impacts (I include myself in this group!).

I understand the intent behind your questioning and I'd just like to comment that I spent a LOT of time thinking about the service file for service-runner and deliberately chose each of the settings after due consideration. I hasten to add that those settings may or may not be appropriate for other services that are part of the EmonCMS back-end.

As I said previously, I'm all for consistency, but I think a service file should meet the specific needs of the particular service first.

greeebs:
> More to the point, if someone stops it (in the future) they would expect it to stay stopped which would be the correct behaviour I'd suggest.

Until they post in the forums asking why their backup's aren't working and why the update wont run.
You are assuming that people know what they're doing ;)
I could argue that in general, people don't actually know what they're doing and should be saved from themselves.

borpin:
> I spent a LOT of time thinking about the service file for service-runner and deliberately chose each of the settings after due consideration.

Great. And therefore perhaps always is the right directive.  It could still be stopped by issuing a `mask` command.

borpin:
> It would be useful to dev and debugging to be able to suspend service-runner to insp the redis queue

You can still do this by issuing a `systemctl mask` command (works for any services).  Even disabling a service does not prevent it from being started if it is 'wanted'. Enabling a service sets it up at boot time (I believe).

borpin:
OK so I have put the user and restart=always back to where they were.

I have added in instructions on how to change the user in a non destructive way i.e. without making a change to the GitHub controlled file using `systemctl edit`.

scripts/services/service-runner/service-runner.service
```diff
@@ -53,9 +49,9 @@ Type=idle
 ExecStart=/usr/bin/python /var/www/emoncms/scripts/services/service-runner/service-runner.py
 User=pi
 
-# Restart script if stopped
+# Restart script if stopped on a failure. Will not restart if not configured correctly
```
greeebs: Change this comment back again after the discussion about using "always"

greeebs:
I'm happy enough with that - you need to change the comment about "restart on failure".

I think it would be good to put the information you have above about stopping it if you really need to into the Readme too. Probably above the install instructions as a separate paragraph just after the numbered list?

borpin:
> I think it would be good to put the information you have above about stopping it if you really need to into the Readme too.

Of course I realised that because no other service has this as a required/wants etc, it would never be restarted except by a reboot or manual restart.  Feedwriter is different as it is a dependency on another service.

Issue #1167, opened 2019-01-20
Redis prefix in service-runner


borpin:
https://github.com/emoncms/emoncms/blob/e6da270246c7981210e668c81d448d9f052f5a20/scripts/services/service-runner/service-runner.py#L45

As discussed [here](https://community.openenergymonitor.org/t/development-roadmap-2019-robust-backup-process/9747/33), it was found that `service-runner` was not picking up the command from the backup module as `service-runner` was simply looking for `service-runner` and my settings had a redis prefix of `emoncms`.

When the python code was changed to look for `emoncms:service-runner` the backup module worked as expected.

borpin:
Just ran it, 6109!  Lots of..
```
   6088) "emoncms:input:5847"
   6089) "emoncms:input:lastvalue:603"
   6090) "emoncms:input:lastvalue:5208"
   6091) "emoncms:input:lastvalue:5218"
   6092) "emoncms:input:lastvalue:4874"
   6093) "emoncms:input:lastvalue:2602"
   6094) "emoncms:input:lastvalue:3647"
   6095) "emoncms:input:lastvalue:3396"
   6096) "emoncms:input:lastvalue:4586"
   6097) "emoncms:input:lastvalue:5706"
   6098) "emoncms:input:lastvalue:3133"
   6099) "emoncms:input:lastvalue:5687"
   6100) "emoncms:input:lastvalue:3473"
   6101) "emoncms:input:lastvalue:3862"
   6102) "emoncms:input:lastvalue:2005"
   6103) "emoncms:input:lastvalue:4949"
   6104) "emoncms:input:lastvalue:1432"
   6105) "emoncms:input:lastvalue:2230"
   6106) "emoncms:input:lastvalue:2540"
   6107) "emoncms:input:lastvalue:1847"
   6108) "emoncms:input:lastvalue:3713"
   6109) "emoncms:input:lastvalue:148"
```
so I wonder if something is not quite right.

borpin:
Could emoncms fire in a setup key that service-runner could scan for once (perhaps) and then service-runner will know what the prefix to use is?

greeebs:
> Could emoncms fire in a setup key that service-runner could scan for once (perhaps) and then service-runner will know what the prefix to use is
I think @pb66 had a better idea - to not use the prefix when sending stuff to service-runner.
@TrystanLea, @glynhudson what are your thoughts on this? Is there much work to remove the instance prefix when pushing values into redis with the `service-runner` name?

[...]

PR #1196, opened 2019-02-07
Update to service-runner.py for redis prefix

borpin:
Resolve issue whereby the different default redis prefix used in emonpi default settings and default settings mean that service runner was unable to get the right item off the queue.

As discussed in #1167 

This is only a temporary fix as any prefix could actually be used.

### DIFF
--- a/scripts/services/service-runner/service-runner.py
+++ b/scripts/services/service-runner/service-runner.py
@@ -42,9 +42,13 @@ server = connect_redis()
 while True:
   try:
     # Check for the existence of a redis 'service-runner' key
-    if server.exists('service-runner'):
+    if server.exists('service-runner') or server.exists('emoncms:service-runner'):
       # We've got one, now to turn it into a cmdline
-      flag = server.lpop('service-runner')
+      if server.exists('service-runner'):
+        flag = server.lpop('service-runner')
+      else:
+        flag = server.lpop('emoncms:service-runner')
+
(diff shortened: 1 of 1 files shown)


greeebs:
Thanks Brian, I think you've made the if statement too complex though...
You don't really need line 45... one check for each prefix in lines 47 to 50 would be more efficient (one less check into redis)
```python
   if server.exists('service-runner'):
     flag = server.lpop('service-runner')
   elif server.exists('emoncms:service-runner'):
     flag = server.lpop('emoncms:service-runner')
```

TrystanLea:
Thanks guys, Ive amended the code to your suggestion @greeebs https://github.com/emoncms/emoncms/commit/884c34fd320a44122e30ba81b4b93a0c5c4e76b1

borpin:
***URGENT***

This has changed the whole flow. https://github.com/emoncms/emoncms/commit/884c34fd320a44122e30ba81b4b93a0c5c4e76b1

Previously, the first `if` checks for a flag and if it exists does some stuff,

Now the 'some stuff' (line 50 on) is executed every time!

TrystanLea:
Thanks @borpin your right, I've amended further https://github.com/emoncms/emoncms/commit/69ad755fe6f1ff35ecc6dd673a1804162fd72d3d, testing now

TrystanLea:
That works fine here

PR #1203, opened 2019-02-13
Limit size of Mount point text

borpin:
**Needs Testing**

Discussion https://community.openenergymonitor.org/t/long-mount-points-disrupt-server-information/10089

Long mount points disrupt server information on admin page.  This PR limits the display to 30 characters plus ellipse.  There may be something clever to display the full path when it is moused over.


TrystanLea:
Looks ok to me, thanks @borpin

PR #1219, opened 2019-03-02
Add absolute value process

nithia:
Process to take the absolute value of an input before logging to a feed.


TrystanLea:
Thanks @nithia looks great!

PR #1222, opened 2019-03-19
Feature/issue 1199/remove url from js error messages

### LINKED ISSUE #1199: Reveal less sensitive detail in error messages
Could we possibly use relative urls rather than full urls revealing the domain name?

It would negate any need to redact details like below before posting on here or the forum. This reveal, may put users off doing screenshots, limiting the info we have for debugging.

image

emrysr:
fix #1199 

I'm able to remove the full url and only show the relative path within the information shown in the alert, however, the alert title and button is controlled by the browser.

delwedd

### DIFF
--- a/Theme/basic/theme.php
+++ b/Theme/basic/theme.php
@@ -129,7 +129,7 @@ if (!in_array($themecolor, ["blue", "sun", "standard"])) {
                     'EmonCMS Error',
                     '-------------',
                     'Message: ' + msg,
-                    'URL: ' + source,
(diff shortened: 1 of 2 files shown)


pb66:
That's cool, thanks @emrysr.

It is easy to just do a screen grab omitting to top of the message box, but harder to obfuscate the path in the text and if not done accurately, users could remove valuable info ie the rest of the path. 

Here's a screenshot taken from issue #1199, with a relative path it would be perfectly safe to post. 

image

PR #1230, opened 2019-04-02
Update Options: All, Emoncms, EmonHub, Firmware, Database

TrystanLea:
Increased flexibility for the update process. 
Update Options: All, Emoncms, EmonHub, Firmware, Database

### DIFF
--- a/Modules/admin/admin_controller.php
+++ b/Modules/admin/admin_controller.php
@@ -14,7 +14,7 @@ defined('EMONCMS_EXEC') or die('Restricted access');
 
 function admin_controller()
 {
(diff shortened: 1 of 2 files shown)


borpin:
@TrystanLea how is the variable `$admin_show_update` set?

pb66:
Since `$admin_show_update` has no default in process_settings.php this remains unset and non-emonpi/base/sd users lose the ability to update the db See https://community.openenergymonitor.org/t/did-manual-update-to-9-9-9-beta-emoncms-no-database-update-like-earlier/10680?u=pb66.
Can we please revert the changes to the update db button so that it remains separate and permanent and not dependent on enabling these extended update buttons? It doesn't use the same code or the service-runner so there seems little point in disturbing something so essential.

PR #1232, opened 2019-04-07
Error evaluating engine number

borpin:
There appears to be an error in evaluating the engine number such that creating virtual feeds fails. https://community.openenergymonitor.org/t/problems-creating-virtual-feed-after-update/10601

My PHP is not good enough to know why, but this change seems to work and it makes more sense to use the helper, get_engine I suggest.


TrystanLea:
Thanks @borpin

PR #1237, opened 2019-04-14
Added hook on_logout

cagabi:
Hi,
As part of the module we have made to use [Auth0](https://github.com/carboncoop/emonCMS_auth0) for single login between all the tools we use in CarbonCoop, we need to be able to logout from Auth0 when the user logs out from emonCMS.
This pull request adds calls to a "logout_hook"


cagabi:
Thanks

PR #1241, opened 2019-04-26
Support for user_capabilities module - invasive changes - do NOT merge

lvogt:
Hi everybody,

this adds some simple support for the user capabilities module. However, the changes are very significant as the default retracts pretty much all access to input, feed and device modules. (Obviously, only test this in a non-crictical environment...)

Maybe someone else finds some use for this, too.

I already liked to my fork in an issue at the capabilities module, but I expect visibility with this PR is better..
https://github.com/takkaria/emoncms-user-capabilities/issues/5


TrystanLea:
Thanks @lvogt for contributing this and apologies for the slow reply, I will continue to keep this open for now until we get a chance in future to review properly.

PR #1257, opened 2019-05-04
Restore download update log button

### LINKED ISSUE #1251: Download update log button missing
After latest updates to the update page the "download update log" button is missing. This button is very useful for users to download update log for support requests etc. 

Selection_153

glynhudson:
Attempt to fix #1251, successfully restored download log button but logger window is not full screen
Discuss with @emrysr and @TrystanLea if log window would be best displayed on a full page display that would be more adaptive to mobile and be more useful to  view more log info the current log boxes are quite small and don't shrink well on mobile. 

Screenshot 2019-05-04 at 19 16 12

### DIFF
--- a/Modules/admin/admin_main_view.php
+++ b/Modules/admin/admin_main_view.php
@@ -212,7 +212,7 @@ table tr td.subinfo { border-color:transparent;}
         </td>
     </tr>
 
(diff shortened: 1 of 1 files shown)


glynhudson:
included in https://github.com/emoncms/emoncms/pull/1259

PR #1265, opened 2019-05-13
defined string constant to check for when no route found

emrysr:
checked for the new constant at output
checked for the new constant when receiving controller response
added x-emoncms-message header on json errors
added not authenticated message to error log
used the new constant as the feed controller's default output

this doesn't relate to an issue. I'd noticed the string `#UNDEFINED#` being used in many places. just made one version of it.


glynhudson:
That's great. Thanks :+1:

PR #1266, opened 2019-05-14
Feature/issue 1245 tab nav

### LINKED ISSUE #1245: Sidebar tab navigation shotcuts
It would be useful to be able to navigate the new sidebar using tab navigation. 

In my option tabs should first traverse the left side bar, bookmarks, bookmark star, then the system cog menu 

In Emoncms V9 it was possible to tab though the top bar then the side bar. I found this quite useful

emrysr:
fix #1245 

the hidden sidebar elements are now not accessible via the keyboard, therefore the user can "get to" all the relevant menu items via the keyboard.

the "top navbar menus" are the first in the "tab order" - so it takes 6 presses of the `tab` key before the sidebar menu items are accessed.

[screenshot with item tabindex overlayed]
delwedd


glynhudson:
Awesome, layout issue fixed and tab-nav working perfect :+1: 

Thanks

PR #1267, opened 2019-05-14
user/gettimezones doesn't require authentication

emrysr:
related to https://github.com/emoncms/graph/issues/83

user module endpoint `user/gettimezones` doesn't expose any personal information so doesn't require authentication before returning a value. removed the check for the `$session['read']` variable.


chaveiro:
Is there any usecase where the user/gettimezones API is invoked by a not logged in user?

emrysr:
@chaveiro - this error was shown when a user's session times out and they load a graph page.
screenshots here: https://github.com/emoncms/graph/issues/83

PR #1276, opened 2019-05-17
sanitized array inputs via POST

### LINKED ISSUE #1275: bookmarks not saving
clicking the bookmark icon submits the request to `user/preferences.json` and gets this error back:
```
<br />
<b>Warning</b>:  rawurldecode() expects parameter 1 to be string,
array given in <b>/var/www/emoncms/core.php</b> on line <b>88</b><br
/>
```

emrysr:
submitted bookmarks as string
set_preferences now sanitizes and decodes ajax inputs
added more logger messages (info)
user/bookmarks stopped the default dashboard from mixing in

fix #1275

### DIFF
--- a/Lib/misc/sidebar.js
+++ b/Lib/misc/sidebar.js
@@ -205,9 +205,9 @@ $(function(){
             }
             // save new user preferences
             $.post(path+'user/preferences.json', {
-                preferences: {
-                    bookmarks: JSON.stringify(bookmarks)
-                }
+                preferences: JSON.stringify({
+                    bookmarks: bookmarks
+                })
             }, function(data) {
                 var $menu = $('#sidebar_bookmarks');
                 var $template = $('#bookmark_link');
--- a/Lib/misc/sidebar.js
+++ b/Lib/misc/sidebar.js
@@ -171,6 +171,7 @@ $(function(){
         var remove = $icon.is('.star');
         var currentTitle = $('h2').first().text();
         if(currentTitle.length==0) currentTitle = $('h3').first().text();
+        if(currentTitle.length==0) currentTitle = document.title;
         if(getQueryStringValue("name")) {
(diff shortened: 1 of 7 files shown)


glynhudson:
This has fixed the issue of not being able to bookmark a page, however it seems to have caused a couple of other issues:

1. Error and broken CSS when viewing default dashboard via bookmark link: `Fatal error: Uncaught Error: [] operator not supported for strings in /var/www/emoncms/Theme/basic/sidebar_view.php:245 Stack trace: #0 /var/www/emoncms/core.php(71): include() #1 /var/www/emoncms/index.php(360): view('Theme/basic/sid...', Array) #2 {main} thrown in /var/www/emoncms/Theme/basic/sidebar_view.php on line 245`

Selection_187

Same issue when viewing the default dashboard directly via the dashboard page. If the dashboard is made 'non-default' then it works fine. 

2. When bookmarking a 2nd (non-default) dashboard the name is not populated in the dashboard list but entry is visible and clickable

Selection_186

Selection_188


2.  Bookmarking Feed page turn off Device Module beta setting and clears all other bookmarks

Device module beta should be turned on by default and maybe the beta toggle taken away. Will discuss this with @TrystanLea

glynhudson:
Nice work, issue 1.) and 2.) above have been fixed. I'm still however experiencing issue 3.) where all the bookmarks will disappear when the Feed page is bookmarked, see this screen cast video when I managed to re-create the issue twice in a row: https://drive.google.com/file/d/1I-cAJL4v6Di6PHjda9FHwUZrp_pr43Ql/view

I will merge this PR since the major issues 1.) and 2.) have been fixed. 

3.) is a minor issue since I can't think of any sane reason why anyone would want to bookmark the Feed page since it's easily accessible via the same side menu. Maybe it should not be allowed to bookmark the Feed and Input page if this makes it easier to fix.

PR #1279, opened 2019-05-20
Fix/bookmark sidebar empty values

emrysr:
a mixture of different fixes related to the css styles, bookmarks and navigation functions. see the commit messages below for a breakdown.


glynhudson:
Dashboard bookmarks have the name "Emoncms - dashboard view <name>" is possible to use just the <name> of the dashboard as the bookmark name? 

Screenshot 2019-05-22 at 12 11 38

PR #1280, opened 2019-05-20
Feature/issue 1274 show inputs if empty feeds

### LINKED ISSUE #1274: If no Feeds are setup Input page should be displayed by default
Currently, the Feed page is displayed by default after login. When setting up a new system the Feed page will be empty therefore it would be more user friendly to display the Input page

emrysr:
fix #1274 - for testing and discussion not for merge until confirmed.


glynhudson:
See discussion https://github.com/emoncms/emoncms/issues/1274

PR #1302, opened 2019-06-04
re-ordered responsive columns to keep latest value

### LINKED ISSUE #1300: Input and Feed values are hidden on iphone4
Due to the small screen size of an iphone4 the value is missing. This has caused an issue for a user setting up a system.

The value is the most important thing to display. On the Input list we could hide the last updated time since we now have the colour band. On the Feed list the feedname could be shortened if required to display the feed value. In the Feed screenshot below there looks like there is sufficient space to display the feed values. 

IMG_0020
Screenshot 2019-06-04 at 02 15 48

emrysr:
added tooltips for input devices
added translation placeholders for js values
modified user-locale.js to use global variable `_user`
fix #1300

latest input value is now the last to be hidden on smaller devices:
delwedd

there is also now a tooltip showing all values... (don't think mobile users can see this??)
delwedd

### DIFF
--- a/Lib/user_locale.js
+++ b/Lib/user_locale.js
@@ -25,4 +25,4 @@
     }
     document.getElementsByTagName("script")[0].parentNode.appendChild(script);
 
-})(typeof user !== 'undefined' ? user: null, typeof _locale_loaded !== 'undefined' ? _locale_loaded: null);
+})(typeof _user !== 'undefined' ? _user: null, typeof _locale_loaded !== 'undefined' ? _locale_loaded: null);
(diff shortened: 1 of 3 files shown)


emrysr:
If the tooltip is required on a device without a pointing device we could display a button/link on smaller screen sizes to show all the relevant information in the tooltip..

[example mobile friendly toolitip target]
delwedd

please create a new issue if this is a good idea to implement

glynhudson:
Interesting idea, adding mobile tool tip target. I think that could be useful otherwise feed ID is impossible to access on mobile view. 

Did you manage to fix the issue with the Input values being hidden on devices with small screens? 

There is currently a merge conflict.

emrysr:
@glynhudson - the screenshots above show the latest value column and not the last updated time

fixed the conflict

PR #1322, opened 2019-06-19
checked for missing REQUEST_SCHEME

### LINKED ISSUE #1321: Error after update master
I am seeing this error after updating all repos to master on my DietPi VM install
```
Notice: Undefined index: REQUEST_SCHEME in /var/www/emoncms/Lib/misc/nav_functions.php on line 126
```

It appears on every page, logged in or logged out.

image

@TrystanLea

emrysr:
fix #1321 

thanks @borpin for spotting this... apparently `$_SERVER['REQUEST_SCHEME']` is not always available on all web servers (by default). I've changed it to check for the missing value.

### DIFF
--- a/Lib/misc/nav_functions.php
+++ b/Lib/misc/nav_functions.php
@@ -119,14 +119,25 @@ function path_is_current($path) {
 /**
  * return the current route path
  *
- * @return void
+ * @return string
  */
 function current_route() {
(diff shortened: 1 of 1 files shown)


borpin:
That'll be because I'm using Lighttpd as the web server.

Issue #1323, opened 2019-06-20
separate device module dependence in input view


emrysr:
currently the input module displays items from the device module. display inputs if no device module installed.

PR #1324, opened 2019-06-20
added check from device module in controller

emrysr:
only loading device js if module installed
loaded input list and device list separately
processed inputs and devices
@todo: draw_devices() with inputs and devices

related to #1323 - separate device module dependance in input module

@TrystanLea - Do you think I'm going the right way with this??
(not for merging... code review)


TrystanLea:
Yes great @emrysr this looks like the right approach

emrysr:
@trystan - could you now test this. not yet tested on a new install of emoncms.

TrystanLea:
Im getting quite a few issues running this with the device module disabled, perhaps we catch up on it next week? I also wonder if it would be better to order this after the vue implementation, as its likely to all change again?

PR #1329, opened 2019-06-24
Updated pi-model.json with new rpi models

gablau:
Updated pi-model.json with new rpi models, of course detect the new Raspberry Pi 4 too!
If you need it here is the generator: [https://github.com/gablau/rpi-revision-codes-utility](https://github.com/gablau/rpi-revision-codes-utility)


TrystanLea:
Thanks @gablau

PR #1333, opened 2019-06-28
Unbreak travis tests

reedy:
It seems the individual tests are failing on 5.6 and Chrome (that's a different issue), but I've mostly got your travis tests working so they run...

And they all pass on 7.2, and FF passes on 5.6

See https://travis-ci.org/reedy/emoncms/builds/551952830

image

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,19 +40,16 @@ before_install:
     ## configure apache virtual hosts 
     - sudo cp -f Lib/travis-CI/travis-ci-apache /etc/apache2/sites-available/000-default.conf
     - sudo sed -e "s?%TRAVIS_BUILD_DIR%?$(pwd)?g" --in-place /etc/apache2/sites-available/000-default.conf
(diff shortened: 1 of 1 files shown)


reedy:
Looks like your browser tests are kinda flaky...

Still. Better they're actually running rather than dying because of test setup issues

reedy:
Passes in this PR but failed in my fork? Helpful :)

https://travis-ci.org/emoncms/emoncms/builds/551957855
https://travis-ci.org/reedy/emoncms/builds/551956996

glynhudson:
Looks like an improvement, let's monitor how it gets on. Considering removing the browser tests. Seems trouble then it's worth

reedy:
Browser tests can be pretty flaky... IF someone is interested in working on them, great. But if they're just rotting, they're not great to have around and just end up annoying people when they flap :)

glynhudson:
Indeed, they keep failing randomly so most devs have stopped taking notice of the tests which is obviously not great!

PR #1345, opened 2019-07-03
Settings ini

TrystanLea:


borpin:
Like the update for ENV vars. I still think a pair of ini files is the way forward for settings.

I would like a means of modifying the settings from the web interface. Could enable dynamic changes to debug level for instance.

A migration tool, that takes the existing settings.php in use and converting it to a settings.ini would also be a good route for migration.

Issue #1286, opened 2019-05-22
feed and input view collapse state not saved


emrysr:
the collapse state of the feed and input list should be saved so that the layout is preserved between page loads.

my version doesn't do this anymore..? not sure when this crept in?

glynhudson:
Good spot, I can confirm I experience the same. State does not save. It would be good to restore this

PR #1360, opened 2019-07-15
Vue.js input list

emrysr:
fully ported inputs list to vuejs to remove the delay using jquery DOM manipulations.

layout is now calculated from data received by API

also removed the input module's dependency on the device module. now works well even when device module not installed.

[new vuejs input list]
delwedd

[selecting inputs to edit/delete]
delwedd

[new vuejs delete dialogue]
delwedd

[new vuejs edit dialogue]
delwedd

[existing node edit dialogue]
delwedd


borpin:
@emrysr are you able to push this into a branch on the emoncms repo to make testing easier?

emrysr:
> @emrysr are you able to push this into a branch on the emoncms repo to make testing easier?

hi @borpin, thanks for taking a look at this. Sorry but I don't have write access to the emoncms repo. I just supply pull requests to be merged - this keeps the repo less messy with all my branches and changes. 

These commands should allow you to create a testing branch with my code changes:
```
git checkout -b emrysr-view-input-list master
git pull https://github.com/emrysr/emoncms.git view-input-list
```

I'd appreciate any feedback on the speed/usability improvements gained from using the [vuejs](https://vuejs.org/) library over the [jQuery methods currently used](https://github.com/emoncms/emoncms/blob/master/Modules/input/Views/input_view.js#L266). I'd like to use more vuejs in other modules (as and when we get to them) to standardise the UI templates/components across the project

emrysr:
@TrystanLea - when devices are deleted all the associated inputs are now also deleted. Also noticed that the column resize functions don't count the device name/description columns only the input rows - this means when an empty device is show the columns are all badly resized (not sure if this case happens a lot?)

delwedd

TrystanLea:
Thanks @emrysr deleting the inputs appear to be working, after clicking on delete there is a duplicate dialog/modal asking if its ok to delete, is it possible to keep the original modal rather than the more basic alert style modal? and just have the single confirmation?

image

emrysr:
@TrystanLea - the double confirmation dialogue. input now deleted once you click `Delete Permanently`

TrystanLea:
Thanks Emrys, the deletion appears to proceed before the "Delete permanently" button has been pressed? so the confirmation doesn’t work as expected?

I also notice that the process list items do not link though in the same way as they used to. e.g clicking on "log" should open that corresponding feed in a graph view.

The modals also show briefly at when you load the page?

TrystanLea:
Thanks Emrys, the deletion appears to proceed before the "Delete permanently" button has been pressed? so the confirmation doesn’t work as expected?

I also notice that the process list items do not link though in the same way as they used to. e.g clicking on "log" should open that corresponding feed in a graph view.

The modals also show briefly at when you load the page?

emrysr:
@TrystanLea - some code was also changed on the device module. here's the pull request
https://github.com/emoncms/device/pull/44

emrysr:
added collapse state to this vuejs version. related #1286

glynhudson:
Just tested this, few things I noticed:

- when all the inputs are deleted for a particular device the device should also be removed from the input list 
- When loading the input page the edit modal flashes up in the background for a split-second before being overwritten 
- Separate issue: Feed list should also be updated for consistency
- I didn't notice much speed improvement, my Input list always loaded quite fast even though there are lots of inputs. But the UI improvements are nice 
- Java script version needs to be  bumped to avoid error when upgrading, needed to use private browsing more to test without errors

emrysr:
@glynhudson - I've worked to correct the issues raised above.

-  stopped template showing on screen before javascript is ready
-  checked for & removed empty devices once inputs are deleted
-  updated view once modal hidden

TrystanLea:
Performance test with old

image

Performance test with new:

image

Width of node description too short, not dynamic?

image

TrystanLea:
@emrysr any chance that you could look at the node description length issue above?

emrysr:
delwedd
[adjusted the device columns width offset to accommodate long descriptions]

PR #1362, opened 2019-07-16
included theme icons as a .svg file

emrysr:
added new `.svg` file with all the theme's `<svg>` icons.
removed the `.php` file that output the `<svg>`

the icons file could be loaded as an external resource by the browser and then cached instead of just being included into every page. other systems could make use of the same `svg` icons as they are in a separate file. could be a good addition to have the module icons do the same? *(eg graphs/icons.svg)*


TrystanLea:
Thanks @emrysr

PR #1363, opened 2019-07-16
username now autofocused

emrysr:
this will enable a user to login quickly with their keyboard as the login form will be in focus on page load.

delwedd


TrystanLea:
Looks good thanks @emrysr

PR #1365, opened 2019-07-17
added check for 'public' && 'admin_only' in  makeListLink()

emrysr:
add 'public' property to login button in user menu
all '*_menu.php' items are now set as 'logged_in_only' unless 'public' is true


TrystanLea:
Thanks @emrysr, looks like its working, seeing the menu title for Login is good.

The menu font appears smaller?

image

emrysr:
thanks @TrystanLea - menu font size is fixed:
delwedd

PR #1369, opened 2019-07-19
added datatype to feedlist tooltips

### LINKED ISSUE #1349: Feeds table lost information
Important details used for adminitration of feeds disapeared on V10.
If the less usefull elements disapeared only when the screen size is small i would understand but that does not seem the case.

See the image bellow. 
Left side is v10, right side  v9.

The problems are signaled by the arrows.
1 - Yellow: Here did the virtual feeds processlist  preview went ?
2 - Red: How the check datatype?
3 - Orange: How to bulk export an entire feed TAG ?

image

emrysr:
fix #1349 

delwedd


TrystanLea:
Diolch!

PR #1377, opened 2019-08-01
clarified the default language selected

### LINKED ISSUE #1243: Default language not used by default but as last resort.
https://github.com/emoncms/emoncms/blob/e3e2cf05a94d5cad0c17996a7d992c0596a127b6/locale.php#L82

Should a default language be the language used by default (i.e. always unless overridden within the program) or used as a language of last resort?

Ref community question https://community.openenergymonitor.org/t/help-to-change-default-user-interface/10776 it seems that when deciding on language, first choice is User, second is the php language settting and finally the 'default' value?

If this is by design I suggest the `settings.php` file has a comment added to this effect with some instructions on how to change the 'first choice' by default.

My preference is to take user define, `default_language` then determine supported php languages with a hard coded fall back to en-GB (in case php has no supported languages listed) and a comment in the settings file to this effect.

emrysr:
fix #1243 

this is now how the language is selected for a user with this pull requeset...

*ORDER OF PREFERENCE FOR EMONCMS LANGUAGE SELECTION:*
1. non logged in users use the browser's language via `$_SERVER('HTTP_ACCEPT_LANGUAGE')`
2. logged in users use their saved language preference from mysql `emoncms.users.language`
3. logged in users without language saved uses `$default_language` from `settings.php`
4. else fallback is set to `en_GB`

If this isn't suitable we can discuss a preferred order of preference for language selection

### DIFF
--- a/locale.php
+++ b/locale.php
@@ -68,7 +68,6 @@ function set_lang($language)
     // DEFAULT - from settings.php (if not in file use 'en_GB')
     $fallback_language = !empty($default_language) ? $default_language : 'en_GB';
 
-    $firstChoice = !empty($language[0]) ? filter_var($language[0], FILTER_SANITIZE_STRING) : $fallback_language;
     $supported_languages = array(
         'cy' => 'cy_GB',
         'da' => 'da_DK',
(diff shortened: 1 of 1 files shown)


emrysr:
https://community.openenergymonitor.org/t/help-to-change-default-user-interface/10776/11

PR #1383, opened 2019-08-12
Settings ini

TrystanLea:


borpin:
How is the file `settings.emonpi.ini` ever used?

PR #1388, opened 2019-08-15
moves apps to center when sidebar collapsed

emrysr:


chaveiro:
How about resize the contents (apps) to fill the available width left over by the closed menu?

PR #1414, opened 2019-09-17
Open github in a new window

Twincarb:
Opens outgoing link in a new browser window with the rel="noopener" tag


TrystanLea:
Thanks @Twincarb

PR #1415, opened 2019-09-17
Update SharedLinuxHostingInstall.md

Twincarb:
Changed para 5 to one step
Para 6 now a list of items that need to be changed which is correct as of the current Master branch.


TrystanLea:
Thanks @Twincarb

PR #1419, opened 2019-09-19
Consistent PSR2 formatting

Twincarb:
Formatted the files to PSR2 which makes them easier to read and maintain.
If this is accepted I will work through the other files in small batches.

### DIFF
--- a/Modules/vis/visualisations/common/api.js
+++ b/Modules/vis/visualisations/common/api.js
@@ -1,9 +1,9 @@
 // Get feed list
 function get_feed_list(apikey){
   var list = [];
-  $.ajax({                    
+  $.ajax({
     type: "GET",
-    url: path+"feed/list.json?apikey="+apikey,       
+    url: path+"feed/list.json?apikey="+apikey,
     dataType: 'json',
     async: false,
     success: function(dt) { list = dt; }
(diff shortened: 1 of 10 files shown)


chaveiro:
I dont approve this kind of generic changes that do not add value at this phase of the project.
On the contrary it makes an headheak to merge and understand code changes from this point after due to major identation, spaces and line changes.

borpin:
I suspect this may meet with some resistance, but I think it is a good thing. Poorly written and laid out code is far more prone to bugs and errors and whilst there may be some short term pain, the longer term gain makes it worth while IME.

TrystanLea:
Thanks for the thought @Twincarb 

Im afraid I like a bit of flexibility in the way the code is formatted e.g:

    if ($apikey===false) { echo "User not found"; die; }

I know its probably not a popular view.. 

I think at the moment with everything else going on, its not a suitable time, although it always feels like we are running behind..

Twincarb:
@TrystanLea not a problem, I just need to remember to disable the automatic PSR2 converter on any PR's I submit (not a problem on my windows box, just on the linux one!)

TrystanLea:
Thanks @Twincarb

PR #1384, opened 2019-08-14
ISC Konstanz contribution

adminde:
Hi guys,

I wanted to commit some hopefully useful things that come in a bundle with other pull requests I'm creating right now, hence I summarized some things in a [forum thread](https://community.openenergymonitor.org/t/isc-konstanz-summer-contributions/)

This pull-request consist of mainly small changes like fixed indents and systax for the new `input_view.php/.js` and `feedlist_view.php`. Generally, I tried to improve the javascript code consistency a little to use more camelCase, as conventions encourage. I additionally changed the collapse-indicators to be chevron icons, as I simply preferred those visually.
Also, I improved the themes' style sheet functionalities a little. E.g. to make less use of `!important` (still a lot though) and include a few things such as the color of top affix background color in the new views.

As we added a total of 3 new menu entries, I increased the menu order values of most modules, as I arrogantly didn't like the input menu `order => 0` :)

Further, the sidebar now allows to be differently colored, depending on the theme used:

emoncms-app-standard

emoncms-app-blue|690x327

emoncms-app-sun|690x325 

At last, I updated the MysqlTimeseries to include some new features, as we use those feeds as a shared resource e.g. for forecasts and EMS a lot.
To improve this, we simply added the possibility to optionally use a separate database for the mysql feeds, configurable via the `settings.php`:

    $feed_settings = array(
        'mysql'=>array(
            'prefix' => 'feed_',
            'generic' => true,
            'database' => 'emondata',
            'username' => 'emondata',
            'password' => '<password>'
        )
    );

If nothing is configured, the engine will behave as before.
Additionally, a different prefix can be configured. Alternatively, by setting *generic* and optionally *prefix* **false**, the feed table name may be configured in the inputs process configuration dialog.

PR #1420, opened 2019-09-19
Configurable SQL engine table names

adminde:
Hi again,

I just removed the MysqlTimeseries engine changes in #1384 to breakout the feature to this PR here.

Basically this PR adds the possibility to optionally use a separate database for the mysql feeds, configurable via the settings.php:

~~~
$feed_settings = array(
    'mysql'=>array(
        'prefix' => 'feed_',
        'generic' => true,
        'database' => 'emondata',
        'username' => 'emondata',
        'password' => '<password>'
    )
);
~~~

If nothing is configured, the engine will behave as before.
Additionally, a different prefix can be configured. Alternatively, by setting _generic_ and optionally _prefix_ **false**, the feed table name may be configured in the inputs process configuration dialog.


TrystanLea:
Thanks @adminde testing here I get the following error trying to create a feed, 

image

I get the same error trying to create any feed engine..

adminde:
Ah sorry I really shoul'd have documented this better.

One of the most important features of this PR (for us) was the possibility to name the tables more dynamically, as we use this database partially as a mediator between services. I implemented this with an additional table name input field, which potentionally wasn't shown until now due to the settings.ini changes.
This table name gets passed to the backend functions in the *options* JSON. But since I need this table name more persistently, **I controversially just added an *options* column in the feeds table**, similar to the processlist.
Sorry for not mentioning this... I kinda forgot it... shame

As everything works for me, I imagine this could be caused by the database not being updated.

TrystanLea:
Thanks @adminde and again my apologise for the delay.
Is there a way to keep this feature as only modifiable via the settings file, without the options field and ability to set table name via process ui? Or I wonder maybe this is a feature best kept in your fork. It's not something we are going to be able to maintain in the emoncms core.. we are not really officially supporting mysql for feed engines as it is.
I'm sorry to not have a more positive response on this but do appreciate all the work you've put in and attempt to contribute it back to the project

TrystanLea:
Did a mix up there and accidentally merged while trying to merge specific changes from emonmuc repository manually, I've now reverted the merge

adminde:
No worries, I commonly started by apologizing about arbitrary changes or something like this for a reason.
And I agree and was already strongly suspecting this 😉
I originally hesitated to demand additional directories like `/var/opt/emoncms/mysqltimeseries` for meta data files, like they already exist for the phpfina/-wa engines, but I guess this would be a less invasive way of doing so.

I will do this for the device and app modules etc. as well in the coming weeks and attempt a less invasive way of bringing our changes to a mergable state. I hope you don't mind the open pull requests until then.

And happy new year btw.

TrystanLea:
Thanks @adminde, 

> I will do this for the device and app modules etc. as well in the coming weeks and attempt a less invasive way of bringing our changes to a mergable state. I hope you don't mind the open pull requests until then.

Great that sounds good! the smaller and less wide-ranging the pull request the higher the likelihood of my ability to merge them, otherwise I just get lost and don’t really know what I'm merging, it also makes it possible for us to test in stages, e.g for the app module: 1. feed selection, 2. css changes 3. feed.js changes 

Thanks again for your work on this :)

PR #1421, opened 2019-09-20
corrected reference to config module's version of vue.js

### LINKED ISSUE #1416: Hardcoded path to `emoncms`
https://github.com/emoncms/emoncms/blob/fa459ed5c01508c7494ab0684b1a58a0c9eac373/Lib/gridjs/grid-demo.php#L111-L116

While digging around for something else, I note this has hardcoded `emoncms` pathe and still points to `config/vue.js`.

Not sure it solves the problem as I'm not sure how this code is used.

@emrysr

emrysr:
fix #1416 

this is not system critical.. just a demo of how the vue.js version of grid.js works

[screenshot of grid js layout of dashboard api data]
delwedd

### DIFF
--- a/Lib/gridjs/grid-demo.php
+++ b/Lib/gridjs/grid-demo.php
@@ -106,9 +106,8 @@
     <!-- END MAIN TEMPLATE CONTAINER ----------------------------------------------------------------  -->
 
 
-
-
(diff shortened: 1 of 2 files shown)


emrysr:
@borpin - on second thought, I'll keep this demo in place for possible future developments. Updated the path to not be reliant on the `config` module.

PR #1427, opened 2019-09-24
documentation for the vis module and attempt to improve codacy grade for multigraph.js

alexandrecuer:
@TrystanLea : a proposition of a simple readme for the vis module..to be continued

A few formal changes in multigraph.php and multigraph.js, so that both files could reach a correct codacy grade....multigraph.js could go from the E grade to C...
The main advantage of complying with codacy rules is that global javascript variables are better identified, which is positive for code readability, which becomes difficult as js files multiply

In multigraph.php, use of the new path var recently introduced/reworked by @emrysr in emoncms.js, and use of the urlParams var of the vis_helper which is not systematically used (is there a reason ? lack of time ?)
this allows multigraph.php to avoid the use of <?php echo blabla ?> within its script section....

This is just a proposition...if yu dont want such mods, no problem

Best

Alex


TrystanLea:
Thanks @alexandrecuer looks good to me. One small question, why did you add in parseInt(i,10)?
https://github.com/emoncms/emoncms/pull/1427/files#diff-599eef820f2dcc17ac39365a5efc4861R155

alexandrecuer:
the use of parseInt(i,10) is related to solving security problems related to square bracket notation :

https://github.com/nodesecurity/eslint-plugin-security/blob/master/docs/the-dangers-of-square-bracket-notation.md

the codacy rule is no user input in property name fields, which they traduce in no unprocessed variable

actually there is no pb on that point but to fool codacy, this is the only way...

I did a similar thing above in the code, introducing the var currentFeed, but only once which is less visible : 
```
var currentFeed=multigraphFeedlist[parseInt(z,10)]; 
```
I've seen you removed the codacy security issues on the repo, so yu dont have this kind of issues
IMO you should keep the security issues, 
the grade will still be C but you will have to remove the /Lib directory (settings> ignored files), which makes sense..

by the way, did you see that the development of flot has resumed since the beginning of the year ? 
https://github.com/flot/flot

on another level, the good thing is to be able to use concole.error and console.warn....
```
/*eslint no-console: ["error", { allow: ["warn", "error"] }] */
```

alexandrecuer:
@TrystanLea : do yu want me to only commit the readme part and to remove all the js implementation....just tell me so that I can manage my branches accordingly :-)

best

Alex

TrystanLea:
Sorry for the delay merging this @alexandrecuer and thankyou again for your work on this

Issue #1424, opened 2019-09-22
Error on Inputs page


borpin:
This error just appeared in the browser window - the tab was in the background at the time.

![image](https://user-images.githubusercontent.com/1322749/65395267-72a5ac00-dd90-11e9-9c43-a7349a3628d8.png)

It might be this line of code `devices[nodeid].inputs.push(inputs[z]);`

In console

![image](https://user-images.githubusercontent.com/1322749/65395282-aa145880-dd90-11e9-8bca-0b7712fc7b3c.png)

alexandrecuer:
I also noticed that the input/view route does not work with microsoft Edge (but who use Edge nowadays?)
![image](https://user-images.githubusercontent.com/24553739/65442206-0b3d3a00-de2c-11e9-82e2-4d31d08f4e10.png)

emrysr:
@borpin - I can't reproduce the error you reported... any additional information might allow me to reproduce it.

I've created a pull request #1428 that simply checks the variable type before running the function, this is always good practice - however this exception shouldn't occur as the variable is always set to be the expected type (array)

emrysr:
@alexandrecuer - thanks for reporting this issue, however I dont think this error is related to the original issue. I'll create a new github issue to track this problem

borpin:
I am getting random errors like this ATM. On a different error, @TrystanLea suggested one of the reasons was REDIS getting out of sync. Could that be a factor here?

borpin:
Errors seem to have gone so closing for now. Cheers.

PR #1428, opened 2019-09-25
checked for type array to avoid error

emrysr:
related to #1424 

@borpin - I can't reproduce the error you reported... added an if statement to check that the variable is an array before the `push()` function is used

please let me know if the issue persists.


borpin:
Could a log message be raised? If it is a situation that should not happen, then I suggest it should be logged.

I'm not sure what the 'fix' is actually doing. Is it overwriting the cookie as the existing cookie is corrupted?  What impact will a user see?

PR #1435, opened 2019-10-01
Fix default settings on emonpi

TrystanLea:


borpin:
@TrystanLea Did you mean to merge to stable?

PR #1437, opened 2019-10-02
Update default settings and example setting

borpin:
Minor modifications to default settings

Add some items to example settings file.

### DIFF
--- a/default-settings.ini
+++ b/default-settings.ini
@@ -4,6 +4,8 @@
 ; 1. copy example.settings.ini and rename to settings.ini
 ; 2. edit entries in settings.ini as required
 ; 3. copy settings from below into settings.ini as required
+;    The items in [] are sections. Any new setting must be 
+;    within the right section in the settings.ini file
 ; ------------------------------------------------------------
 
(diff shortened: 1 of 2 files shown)


TrystanLea:
Thanks @borpin

TrystanLea:
I've changed the default buffer time to 300s, it reduces write load a fair bit. The alternative would be to set it to longer here: https://github.com/openenergymonitor/EmonScripts/blob/master/defaults/emoncms/emonpi.settings.ini? and keep the default at 60?

borpin:
> I've changed the default buffer time to 300s, it reduces write load a fair bit.

There has been discussion about this on the forum. I think it was different in different places but most recently I thought it has been 60. Might be worth checking with others.

I think a default of 60 and raise it for the emonpi might be the best route.

TrystanLea:
Ok done:

https://github.com/emoncms/emoncms/commit/34d976db7f09b74ad17d86f234bc37edb3621834

and 

https://github.com/openenergymonitor/EmonScripts/commit/57cf3f3c683570cb5b472424333307705ad10536

PR #1438, opened 2019-10-02
Update admin_model.php with demandshaper service

borpin:
Add `demandshaper` service to the services array


TrystanLea:
Thanks @borpin

PR #1440, opened 2019-10-02
Add 'none' to firmware type in admin

borpin:
For systems without any RFM or serial interface, a 'none' firmware type is needed to make updates cleaner.

Ideally there would be a means of setting the firmware type.


TrystanLea:
Thanks @borpin

PR #1461, opened 2019-10-24
merge master to stable

TrystanLea:
- localization thanks to @thib66 
- documentation archive
- example settings in php format
- update of default php settings
- UDP broadcast setting


borpin:
Tag isn't annotated :cry: .....

TrystanLea:
It should be: https://github.com/emoncms/emoncms/releases/tag/10.1.10?

PR #1466, opened 2019-10-25
Add admin page for viewing the current Settings object

borpin:
Extend the admin view capabilities with a separate page for viewing the current Settings object.

TODO
* Introduce copy txt/md buttons at top - not sure the best way to do this.
* In admin view, float button across to right - tried, but my CSS is not up to it 😢

### DIFF
--- a/Modules/admin/admin_main_view.php
+++ b/Modules/admin/admin_main_view.php
@@ -275,6 +275,9 @@ listItem;
         </dl>
 
         <h4 class="text-info text-uppercase border-top pt-2 mt-0 px-1"><?php echo _('Emoncms'); ?></h4>
+        <div>
+            <a href="<?php echo $path; ?>admin/settings" class="btn btn-info"><?php echo _('View Emoncms Settings'); ?></a>
+        </div>
         <dl class="row">
             <?php echo row(_('Version'),$emoncms_version); ?>
             <?php echo row(_('Modules'), $emoncms_modules); ?>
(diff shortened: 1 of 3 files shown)


TrystanLea:
Thanks @borpin would you like this kept open until further development on css etc is added or merged in the interim? 

Still not 100% comfortable about printing out the settings file, maybe obfuscating all passwords would be a good idea first?

borpin:
> Thanks @borpin would you like this kept open until further development on css etc is added or merged in the interim?

I'd have suggested merging in the interim.



> Still not 100% comfortable about printing out the settings file, maybe obfuscating all passwords would be a good idea first?

Not considered this. Strike me that this is inside the admin account, and could be accessed off the machine directly (i.e. not hashed and salted) so I don't think that is an issue.

The API key is available, so actually I think this falls into the same sphere.

Modules/admin/admin_main_view.php
```diff
@@ -275,6 +275,9 @@ function row($title, $value, $title_css = '', $value_css='') {
         </dl>
 
         <h4 class="text-info text-uppercase border-top pt-2 mt-0 px-1"><?php echo _('Emoncms'); ?></h4>
+        <div>
+            <a href="<?php echo $path; ?>admin/settings" class="btn btn-info"><?php echo _('View Emoncms Settings'); ?></a>
+        </div>
```
Twincarb: To address the comment of how to align right the code below will work (although it would be better to give the div a class and then add the style into admin_styles.css)
Change the first div tag too:
`<div style="float:right;">`
After the closing div tag add:
`<div class="clearfix"></div>`

Issue #1402, opened 2019-09-08
Add notice to banner of new stable version available


borpin:
Could a notice be added to the top banner that informs a user when there is a new stable version of emoncms available.  This can be done via manipulating the strings returned from the local repo and
```
https://raw.githubusercontent.com/emoncms/emoncms/stable/version.txt
```
Consideration will need to be made to Emoncms versions in use prior to the `version.txt` being available and perhaps a 'last safe' version to update from. If earlier than that version perhaps a general text message should be displayed instead.

TrystanLea:
This is under development here closing duplicate issue https://github.com/emoncms/emoncms/pull/1468

PR #1468, opened 2019-10-26
Update notification 10 2019

thib66:
Add an update notification icon in the main menu when the version found in local version.txt and the version placed in https://raw.githubusercontent.com/emoncms/emoncms/stable/version.txt are different.
image
The name of the last stable version of EmonCMS is indicated in the tooltip.
Should fix the issue #1402.
Let me know if everything is OK or if you need another improvements.


borpin:
Great idea. #1402

TrystanLea:
Nice idea, would it be possible to move this to the Admin > Update page so that it does not run on every page load? There will be users who dont always have internet access. Might also be good to add in catching of timeouts

thib66:
Maybe you want to move the notification here :
image
Let me know if is Ok for you
regards,

borpin:
BTW @TrystanLea, I noted elsewhere that process settings seems to be called multiple times on a page load.

thib66:
Hello @TrystanLea @borpin ,
script has been improved in order to check the version update only at the connexion of the user in order to limit the call to the distant version.txt. Use of http_request as requested. I guess that this function manages timeouts.
Looking forward to your feedback,
Regards,

thib66:
image
when clicking on "update Emoncms" button we could set $_SESSION['checkUpdate'] to false in order to force the update check and change the status of the update notification

thib66:
Hello,
@TrystanLea @borpin , may I have a feedback from you?
Thanks in advance,

borpin:
> may I have a feedback from you?

Yes that looks to be a good solution to me.

glynhudson:
Nice idea, I've just tested. It seems to appear on every page, not just the admin page. I agree with @TrystanLea this should only appear on the admin page. 

Selection_288

Also, listing all the updates in the tooltip does not work too well since this will overflow off the page. How about the tool tip just says "update available" then clicking the icon would take the user to the update page which would list which module updates are available? 

Selection_290

Currently, clicking the icon does not seem to have any effect.

thib66:
Hello @ glynhudson,
ok for displaying the icon only for the admin page.
You mention that a click on the icon would take the user to the update page.
Could you tell me what is the url to reach the update page? 
Maybe there is a confusion with the admin page which already lets the user an update of the versions.
In this case I do not see the interest to create an update page which will duplicate the existing admin page.
Looking forward to your feedback,

thib66:
Hello @glynhudson ,
can I have a feedback from you?
Regards,
Aymeric

TrystanLea:
Hello @thib66 sorry for the delay. I dont think a link is needed on every page, it generates the same update check overhead anyway - which is what Im keen to avoid. Just the entry as you suggested in your screenshot above on the admin page in the update section - so that the user knows whether there is any point hitting update.

Thanks a lot

thib66:
image
icon moved to administration page.
@glynhudson looking forward to your feedback,

thib66:
Use of snakbar property in order to display the list of updates in an elegant way (no longer overflows off the page).
image
image

@TrystanLea @glynhudson looking forward to your feedback

thib66:
Hello @TrystanLea @borpin @glynhudson ,
Is there something that prevents from putting PR # 1468 from Master branch?
I would like to delete my repository.
Thanks a lot for your answer.
Regards,

TrystanLea:
Hello @thib66 sorry for the delay on this. I've been thinking about this in relation to another idea as well to show a full list of modules and their current and available versions e.g https://github.com/emoncms/emoncms/pull/1601 

As I look through the current admin view code I really think it needs rebuilding to use javascript based templating rather than php so that we can then use ajax and provide API end points for the different server information requests and actions such as getting the available update versions.

It would be great to do the same with your implementation here, move it to a admin_model.php function and then provide access via an end point described in the admin_controller.php *e.g /admin/checkforupdates*, which is then loaded via ajax request from the client side asynchronously. But it may be better to wait until we refactor the rest of this admin page before we do that...

Sorry to keep this on hold for now.

PR #1484, opened 2019-12-14
Virtual null values

emrysr:
Check query response when describing table in db_schema_update_column().
Returned NULL value for get_timevalue() virtual feeds with no lastvalue
added comments
Hidden E_DEPRECATED PHP warnings as they'll break any json responses

delwedd

### DIFF
--- a/Lib/dbschemasetup.php
+++ b/Lib/dbschemasetup.php
@@ -184,14 +184,6 @@ function db_schema_update_column($mysqli, $table, $field, $spec, &$operations)
     $diff_datatype = db_schema_diff_datatype($spec, $current, $field);
     $add_key = db_schema_diff_key($spec, $current);
 
-
-    // if($table === 'feeds' && $field === 'id') {
-    //     // var_dump($add_key);
-    //     var_dump($diff_datatype);
-    //     // var_dump(array($table, $field, $spec));
-    //     exit('dave');
-    // }
-
     // If we do, generate a new field spec and update table
     if ($diff_datatype || $add_key) {
         $field_spec = db_schema_make_field($mysqli, $field, $spec, $add_key);
--- a/Lib/dbschemasetup.php
+++ b/Lib/dbschemasetup.php
@@ -35,7 +35,7 @@ defined('EMONCMS_EXEC') or die('Restricted access');
 // the data type, and so not respecifying all of them when specifying
 // the field will result in them being lost.
 //
-function db_schema_diff_datatype($spec, $current, $field)
+function db_schema_diff_datatype($spec, $current)
 {
     $changed = false;
 
--- a/Lib/dbschemasetup.php
+++ b/Lib/dbschemasetup.php
@@ -177,7 +177,7 @@ function db_schema_add_column($mysqli, $table, $field, $spec, &$operations)
 //
 function db_schema_update_column($mysqli, $table, $field, $spec, &$operations)
 {
-    $result = $mysqli->query("DESCRIBE `$table` `$field`");
+    $result = $mysqli->query("DESCRIBE $table `$field`");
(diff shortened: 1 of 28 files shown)


TrystanLea:
@chaveiro could you review this proposed change to virtual feeds?

Modules/feed/feed_model.php
```diff
@@ -516,7 +516,12 @@ public function get_timevalue($id)
         if ($engine == Engine::VIRTUALFEED) { //if virtual get it now
             $this->log->info("get_timevalue() calling VIRTUAL lastvalue $id");
             $lastvirtual = $this->EngineClass(Engine::VIRTUALFEED)->lastvalue($id);
-            return array('time'=>$lastvirtual['time'], 'value'=>$lastvirtual['value']);
+            if ($lastvirtual) {
+                $timevalue = array('time'=>$lastvirtual['time'], 'value'=>$lastvirtual['value']);
+            } else {
+                $timevalue = array('time'=>null, 'value'=>null);
```
chaveiro: What problem is this fixing?

Lib/dbschemasetup.php
```diff
@@ -172,13 +177,21 @@ function db_schema_add_column($mysqli, $table, $field, $spec, &$operations)
...
-    $result = $mysqli->query("DESCRIBE $table `$field`");
+    $result = $mysqli->query("DESCRIBE `$table` `$field`");
     $current = $result->fetch_array();
 
     // Check if we 1. need to change type 2. need to add a primary key
     $diff_datatype = db_schema_diff_datatype($spec, $current, $field);
     $add_key = db_schema_diff_key($spec, $current);
 
+
+    // if($table === 'feeds' && $field === 'id') {
```
chaveiro: Please try to keep code clean

process_settings.php
```diff
@@ -63,8 +63,10 @@
 
 // Set display errors
 if (isset($settings["display_errors"]) && ($settings["display_errors"])) {
-    error_reporting(E_ALL);
+    // @todo: add setting to allow for deprecated warnings
     ini_set('display_errors', 'on');
+    // show all errors except for E_DEPRECATED warnings
+    error_reporting(E_ALL & ~E_DEPRECATED);
```
chaveiro: Why omit the error if we could fix the code?

Modules/feed/feed_model.php
emrysr: has hit errors when $lastvirtual is null and has no ['time'] or ['value']. so I checked for that and returned null values for the `time` and `value`.

process_settings.php
emrysr: The deprecated messages are warnings (nothing wrong with the code), however they'll print out and break any API responses that require a specific output format.

emrysr: If you'd like to keep all errors and warning messages showing I'll change it back.

Modules/feed/feed_model.php
chaveiro: **Time** must not be null but **value** can be null.
This fix should go to where the original error came from, the engine class lastvalue() function.

process_settings.php
chaveiro: They are warning that on a next version of PHP the code will not work as the functions are deprecated.
Why wait until it brokes if you can fix it now?

chaveiro:
Comments added to the previous review.

emrysr:
this pull request should be ok now. all the confusion taken out.

TrystanLea:
This issue was a bit of a strange issue, we observed two different behaviours on two different systems, if the virtual feed last value function returned false on one system, the feed_model would ignore the missing time and value keys and populate them as null anyway, on the other system this same section would throw the expected undefined error...

PR #1495, opened 2020-01-06
added try/catch block arround file_exists() code

### LINKED ISSUE #1494: Include file_exists check in try catch block
Re discussion https://community.openenergymonitor.org/t/need-help-with-first-emoncms-setup-on-synology-diskstation/12439/22

The check for file exists may raise an exception.  The check should be included in a try catch block.

https://github.com/emoncms/emoncms/blob/1a587503d00780214640f70c7cc4e17bdf43ab27/Modules/admin/admin_model.php#L318

This should be included across all system info checks as we cannot assume any particular hardware/software setup.

emrysr:
outputted errors to log

fix #1494 

@tehnoinstyle / @borpin - I was unable to recreate the error that causes the Exception, however I have added the check based on the messages reported in the issue.

please let me know if this doesn't fix the errors you see.

### DIFF
--- a/Modules/admin/admin_model.php
+++ b/Modules/admin/admin_model.php
@@ -315,8 +315,13 @@ class Admin {
         if(strtoupper(substr(PHP_OS, 0, 3)) === 'WIN') {
             $v = "n/a";
         } else {
-            if (file_exists('/usr/sbin/mosquitto')) {
-                $v = exec('/usr/sbin/mosquitto -h | grep -oP \'(?<=mosquitto\sversion\s)[0-9.]+(?=\s*)\'');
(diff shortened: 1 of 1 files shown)


borpin:
Might be worth setting `$v` to `N/A` as it does for WIN OS.

Also, I do not think it is an error as such, it is simply a different setup. Perhaps downgrade to INFO?

emrysr:
thanks for the feedback @borpin - I've made the suggested changes in : https://github.com/emoncms/emoncms/pull/1495/commits/b6803e4de3535213a994df2ecef08e532d02a61d

PR #1513, opened 2020-01-24
Include the received JSON string in error message

borpin:
As title.

### DIFF
--- a/Modules/input/input_methods.php
+++ b/Modules/input/input_methods.php
@@ -132,7 +132,7 @@ class InputMethods
                 }
                 $inputs = $jsondata;
             } else {
-                return "Input in not a valid JSON object";
+                return "Input in not a valid JSON object : ".$datain;
             }
         } else {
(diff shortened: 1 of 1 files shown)


TrystanLea:
Thanks @borpin what is the reason for this change? The user already has access to the JSON string sent what does returning it achieve?

I think its also best practice to sanitise user input before printing it back, not sure what harm it would do in this case but a little weary of causing an unexpected issue.

borpin:
It is a debugging help. All you get when it fails is 'false' - sending back what was received would help to debug issues. Alternative is to make the log message include more information. Again usually just says 'failed'.

TrystanLea:
How about writing it to the emoncms log instead?

borpin:
That would be helpful.

TrystanLea:
I've added in a log of the invalid json https://github.com/emoncms/emoncms/commit/611fd01be39052b6aa5688ce9215aae0d428f795

PR #1524, opened 2020-03-08
FIX: JS error on deleting empty feed

hutorny:
When user is deleting an empty feed, browser shows JS error.
This pull request suggest a fix for that error


TrystanLea:
thanks @hutorny

PR #1525, opened 2020-03-15
beautify list of php modules

gablau:
I tried to make the list of modules more beautiful, if the module version is the same as the php version then it doesn't display it.

Screenshot_2020-03-15 Emoncms - admin view


TrystanLea:
Cheers @gablau looks great!

PR #1527, opened 2020-03-16
Updating italian language

gablau:
Current status:
- admin     100%
- eventp     58%
- feed       53%
- input      52%
- process    60%
- schedule   68%
- user      100%
- vis        75%
- theme     100%


gablau:
@emrysr Thanks to you for this Open source project

> the change in the `directoryLocaleScan()` function in `locale.php` is an good improvement as it looks cleaner, however it has been as it is for years without issue. both work fine for this case.

Yes, it worked well, but if the function found .pot files in the "locale" folder then it didn't work properly.
That's why I had to fix it

emrysr:
> @emrysr Thanks to you for this Open source project

I can't take credit for all the contributions from all the core developers - I work for openenergymonitor.org and so am able to work on this during the day as part of my job.

re: `directoryLocaleScan()` - well spotted - good fix :100: 

The translations will open the software out to so many more users - thanks again!

PR #1533, opened 2020-03-23
Configurable SQL engine table names

adminde:
Hi guys,

sorry for the loong delay, but thanks to armageddon I had some time to start working through all my problematic pullrequests.

This is the PR #1420 again, just without the database schema changes.
This PR adds the possibility again, to optionally use a separate database for the mysql feeds, configurable via the settings.php:

~~~
$feed_settings = array(
    'mysql'=>array(
        'datadir' => '/var/opt/emoncms/mysql/',
        'prefix' => 'feed_',
        'generic' => true,
        'database' => 'emondata',
        'username' => 'emondata',
        'password' => '<password>'
    )
);
~~~

If nothing is configured, the engine will behave as before.
Additionally, a different prefix can be configured. Alternatively, by setting _generic_ and optionally _prefix_ **false**, the feed table name may be configured in the inputs process configuration dialog.


TrystanLea:
Thanks @adminde much appreciated! @chaveiro would you be able to help review this? as I dont use MySqlTimesSeries Im not best placed to review. Ideally we could identify a number of users using MySqlTimeSeries that could help test, otherwise Im merging blind really

Modules/feed/engine/MysqlTimeSeries.php
```diff
@@ -742,60 +836,138 @@ public function post_bulk_save()
...
+            }
+        }
+        if (!empty($field)) {
+            return $table[$field];
+        }
+        return $table;
+    }
 
-// #### \/ Bellow are engine private methods    
+    protected function get_table_name($feedid)
```
hutorny: I would also use this method if it is made public

Modules/feed/engine/MysqlTimeSeries.php
```diff
@@ -742,60 +836,138 @@ public function post_bulk_save()
...
+        $result = $mysqli->query("SELECT datatype FROM feeds WHERE `id` = '$feedid'");
+        $row = $result->fetch_array();
+        return $row["datatype"];
+    }
+
+    private function get_table($feedid, $field=null)
+    {
+        if ($this->generic) {
+            $table = array(
+                "name" => ($this->prefix ? $this->prefix : "").trim($feedid),
```
hutorny: I'd suggest a call to `get_table_name` instead of computing table name in place

TrystanLea:
Hello @adminde I've done a basic test here, creating a feed, posting a few datapoints and it appears to work fine. I noticed in the code the creation of I think a non-sql meta file, I wonder if you could provide a bit more detail on what the intention is here?

TrystanLea:
@chaveiro @alexandrecuer @isfh-ohrdes any comment?

TrystanLea:
Thanks for your input @alexandrecuer @chaveiro.

@adminde I noticed in the code the creation of I think a non-sql meta file, I wonder if you could provide a bit more detail on what the intention is here?

adminde:
Hei guys,

and thanks @hutorny! I included both suggestions in the latest commit.

@TrystanLea, to be honest.. I am not sure what a "sql" meta file would be in comparison :)
I tried to get away from possible breaking points if the database schema does not get updated, as well as its invasiveness.
The general intention was to rely on redis if possible. The meta file (and also the possibility of a test if the directory exists or not), was a less invasive, backwars compatible way to make sure, that the engine behaves completely "normal", if the user makes no changes to the settings or the database schema.
I thought it wouldn't hurt, if the meta file can be edited by hand, if the admin for whatever reason decides to change a table name.

Very open to suggestions though

TrystanLea:
Hello @adminde I've tested your latest version here and it looks to be working fine. I am happy to merge this now. When would be the best time for you for me to merge this, it would be great to be able to keep in contact if there are any issues for users using the mysql engine?

adminde:
Hi @TrystanLea,
the whole week or generally from now on is perfectly fine for me regarding this

TrystanLea:
Thanks @adminde sorry for the delay again, this is now merged in :)

PR #1537, opened 2020-03-28
FIX: sidebar divider made unclickable and with href=#

hutorny:
Side bar divider is shown as a clickable empty menu item navigating to `+` causing page reload. Such behaviour may confuse the user. This pull request makes the divider visually distinct and non-clickable. Also its href is changed to `#` so that even it is clickable it would not reload the page.


TrystanLea:
Thanks @hutorny

PR #1538, opened 2020-03-28
FIX: column H with the process list accomodate all remaining space

hutorny:
Column H containing list of processes is of a fixed width. When an input has many processes, the content is wrapped and hidden. It looks confusing. This pull request removes width constraint on that column making it stretching on all remaining space

### DIFF
--- a/Modules/input/Views/input_view.js
+++ b/Modules/input/Views/input_view.js
@@ -974,7 +974,7 @@ function draw_devices() {
     app.col.G = ((max_description_length * 8) + 70); // additional padding to accomodate description length
     app.col.D = ((max_value_length * 8) + 17);
     app.col.E = ((max_time_length * 8) + 20) + 20; // additional padding to accomodate the 'weeks/days/hours/minutes/s' suffix
-    app.col.H = 200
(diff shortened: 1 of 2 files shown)


TrystanLea:
Thanks @hutorny 

It's still not quite right, you end up with an overflow over the configure buttons on small screens e.g:

image

This provides another approach https://github.com/emoncms/emoncms/commit/d1336b39bfb41f644534d474c968c6a85c34d0d1 but could still do with expanding the column width on larger screens

image

hutorny:
I see. Perhaps a style with overflow hidden would help to deal with a too wide column. I'll investigate.

hutorny:
Updated PR suggest solution with using css `calc` for calculating remaining space. Also it changes style of pull right buttons on this page to ensure they are not pushed away if column width is not calculated to fit pixel-precisely.

TrystanLea:
Thanks @hutorny looks like that improves it but still not quite right, though we should have got this right the first time, happy to merge for now and revisit in future

image

TrystanLea:
Recent updates should have fixed or at least improved this issue and a further review is ongoing.

PR #1539, opened 2020-03-28
FIX: PHP error caused by disabled REDIS

hutorny:
In some cases, when REDIS is disabled, cleaning a process list may endup with a PHP error. This pull request suggest a fix for that.


TrystanLea:
Thanks @hutorny

PR #1540, opened 2020-03-28
ADD: support for schemas with compound key

hutorny:
Currently `dbschemasetup` supports only schemas with a single-field primary key. In some cases this may impose unnecessary limitations on design of a custom module. This pull request suggests an approach on handling compound keys with minimal changes to `dbschemasetup`.
A schema just indicates more than one column as a key and `db_schema_make_table` detects this case and generates the key as the `PRIMARY KEY` clause
For example, the following schema :
```
$schema['coolstuff'] = array(
    'feedid' => array('type' => 'int(11)', 'Null'=>'NO', 'Key'=>'PRI'),
    'time' => array('type' => 'int(11)', 'Null'=>'NO', 'Key'=>'PRI'),
    'data' => array('type' => 'float', 'Null'=>'NO'),
);
```
is translated to the following SQL:
```
CREATE TABLE `coolstuff` (`feedid` int(11) NOT NULL, `time` int(11) NOT NULL, `data` float NOT NULL, PRIMARY KEY (`feedid`,`time`)) ENGINE=MYISAM;
```

### DIFF
--- a/Lib/dbschemasetup.php
+++ b/Lib/dbschemasetup.php
@@ -133,6 +133,20 @@ function db_schema_make_index($table, $field) {
 }
 
 //
+// Make a compound key if the schema has more than one key
+//
+function db_schema_make_compound_key($schema)
+{
+    $fields = array();
+    foreach ($schema as $field => $spec) {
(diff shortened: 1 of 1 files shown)


TrystanLea:
Thanks @hutorny

PR #1543, opened 2020-03-30
Update service-runner to python3.

bwduncan:
Port service-runner to python3.

- Change indentation to 4 spaces to conform with the rest of the
project.

- Utilise redis' blpop method to block until there is something to do,
instead of spinning with a short sleep.

- There was a lot of code fighting against itself, like signal handlers
which didn't do any cleanup, other than printing a message, which
complicated the main loop. Since the program is running under a service
supervisor (presumably systemd) this is unnecessary, so remove it.

- Utilise the print(..., flush=True) syntax.

- Avoid subprocess.call(..., shell=True). The command lines are so
simple that we can (and already do) parse them ourselves.

- Avoid wrapping too much code in a try/except. There is only one call
to redis and therefore only one place where the connection can fail, so
only handle those errors there.

- Move everything into a main function, to keep linters happy.

Tested on emonpi 2 with raspbian buster.

### DIFF
--- a/scripts/services/install-service-runner-update.md
+++ b/scripts/services/install-service-runner-update.md
@@ -14,7 +14,7 @@ The process is as follows:
 **Check the old version has been uninstalled**
 Install the service using the following commands (if redis is already installed skip that command):
 ```bash
-sudo pip install redis
+sudo pip3 install redis
 sudo ln -s /var/www/emoncms/scripts/services/service-runner/service-runner.service /lib/systemd/system
 sudo systemctl enable service-runner.service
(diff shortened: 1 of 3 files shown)


borpin:
Great work but I think this should go into a dev branch for wider testing before hitting `master`. Some folk use `master` and this has the potential to break stuff badly.

In addition, it will require the installation of python3 on all systems plus some other update tweaks I suspect.

TrystanLea:
Thanks a lot @bwduncan, will merge as part of coordinated shift to python3 to address @borpin's points

PR #1544, opened 2020-04-01
FIX: handling of missing feeds

hutorny:
When user deletes a feed and then open a previously saved graph, the `graph` module shows improper message and stalls with a broken graph with no ability to fix it. This PR fixes PHP error (improper collection of feed ids) and and suggest a solution for broken graphs by returning an array of missing feed ids. This will let the client (`graph` module) to remove bad feeds from the graph and keep good ones.


TrystanLea:
Thanks a lot @hutorny

PR #1545, opened 2020-04-01
MOD: timezone as a linked field

hutorny:
Currently the timezone is retrieved in two steps - (1) get user id, (2) get timezone
This pull request suggest linking timezone from the user's table and accessing it with `get_field($feedid,'timezone')`.
Since get_field function uses REDIS cache when available, this approach should improve performance.
To handle timezone changes, the feeds cache is invalidated on updating the timezone field.

### DIFF
--- a/Modules/feed/feed_model.php
+++ b/Modules/feed/feed_model.php
@@ -497,9 +497,10 @@ class Feed
                 $val = $lastvalue[$field];
             }
             else if ($this->redis) {
+                if (! $this->redis->hExists("feed:$id",$field)) $this->load_feed_to_redis($id);
                 $val = $this->redis->hget("feed:$id",$field);
             } else {
-                $result = $this->mysqli->query("SELECT * FROM feeds WHERE `id` = '$id'");
+                $result = $this->mysqli->query("SELECT feeds.*, users.timezone FROM feeds INNER JOIN users ON feeds.userid = users.id  WHERE feeds.id = $id");
                 $row = $result->fetch_array();
                 if (isset($row[$field])) $val = $row[$field];
             }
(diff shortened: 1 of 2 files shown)


hutorny:
The user may change timezone in his profile and this change will have no effect on the timezone stored in REDIS. This may confuse the user - the old timezone will be applied instead of the new one.

chaveiro:
Ok, perfect

borpin:
It will still be better linking the Feed to it's own timezone - i.e. the TZ of the data source. This is not necessarily the same as the timezone where the data is being recorded.

To link it to the user is simply wrong.

Yes multigraph needs to be handled differently, but inherently, when collecting Solar for instance, when local midday is for the data being collected is more relevant than the TZ of the person looking at the data.

hutorny:
@borpin , I agree with you reasoning, but currently there is no means for storing and managing feeds' time zones. This PR could be an intermediate step toward to feed specific time zones - all consumers of TZ migrate to `get_field($feedid, 'timezone')` so that when TZ is implemented as a real field no changes needed on their side.

borpin:
@hutorny 

> I agree with you reasoning

I'm glad someone does.

I think it would be a good addition.

hutorny:
While using this approach I discovered some scenarios that bypass the timezone caching. Updated PR handles such cases

borpin:
I'll link this issue https://github.com/emoncms/graph/pull/51. 

As far as I can tell the implemented 'change TZ' functionality does not currently work. If I change the TZ and reload, nothing happens.

It is unclear if the *Start* and *End* DTG are 'local' times or not.

hutorny:
Here is a scenario I am using to verify TZ feature:
1. Load a graph, note time of some data points on the graph
2. Open My Account (user/view)
3. Change time zone and save it
4. Go back to the graph, and reload it
5. Compare time of the data points noted on step one

borpin:
> Here is a scenario I am using to verify TZ feature:

But what about new data on that Feed once you change the TZ?  Unless you include a timestamp with the Feed data, and depending on which way you moved the TZ, either Feed data will be overwritten (if you go back), or there will be a gap in the data.

Are you testing this with multiple feed sources (HTTP with time, HTTP without time, MQTT value, MQTT full JSON)? How time is handled at the input point varies.

hutorny:
This MR does not change the way how TZ is used. It only suggests a new way of getting TZ in the PHP code.

TrystanLea:
> As far as I can tell the implemented 'change TZ' functionality does not currently work. If I change the TZ and reload, nothing happens

You are quite right, I have removed the feature and also fixed another timezone related issue on the graph module that resulted in fixed interval rather than day, week, month, year modes being used on the graph module, see recent commits here https://github.com/emoncms/graph/commits/master 
I've updated stable as well and created a release tag v2.0.9

PR #1547, opened 2020-04-01
Extra " causes invlid html

Twincarb:
Removed 3 x occourances of " are causing invalid html. Firefox produes "="" chrome just ignores it.


TrystanLea:
Thanks @Twincarb

PR #1549, opened 2020-04-03
Update logic to check for settings.ini first

Twincarb:
Updating/Reversing the logic for which settings file to check for, if I am correct we are now using the settings.ini file so it should be checked for before settings.php

removed comments that are repeating the code

edited the error title to be specific on what is missing.

### DIFF
--- a/process_settings.php
+++ b/process_settings.php
@@ -17,27 +17,21 @@ require_once('Lib/enum.php');
 // Load settings.php
 $settings_error = false;
 
-if(file_exists(dirname(__FILE__)."/settings.php")) {
+if (file_exists(dirname(__FILE__)."/settings.ini")) {
+    $CONFIG_INI = parse_ini_file("default-settings.ini", true, INI_SCANNER_TYPED);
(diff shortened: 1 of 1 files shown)


borpin:
The issue here is that a new system (which will have a `settings.ini`, that imports a backup, will then use the wrong settings file as the import of a backup brings across what might be a `settings.php` file from the system backed up.

What we need is a means of transferring the `settings.php` to a `settings.ini` file to complete the migration. Ideally only the settings that are different to the default should be included in the `settings.ini` file.

Even better would be a means of editing settings from the UI.

For now it should stay as-is.

Twincarb:
@borpin that does explain why it is the way around that it is, I havent looked to see what is saved in the backup, its possible that the restore script could be made to handle the conversion to settings.ini But that is outside of this PR.

I agree with having the ability of editting of settings in the UI would be handy... 

image

TrystanLea:
Did you publish that settings UI somewhere @Twincarb ?

PR #1551, opened 2020-04-04
Clear warning when hosted on php7.4

Twincarb:
The following warning is generated when php = 7.4
<b>Warning:</b>
Use of undefined constant LC_MESSAGES - assumed 'LC_MESSAGES' (this will throw an Error in a future version of PHP) in C:\xampp\htdocs\emoncms\locale.php on line <b>108</b>

### DIFF
--- a/locale.php
+++ b/locale.php
@@ -105,7 +105,7 @@ function set_lang($language)
 function set_lang_by_user($lang)
 {
     $locale = $lang.'.UTF8';
-    define(LC_MESSAGES, $locale);
+    define('LC_MESSAGES', $locale);
     putenv("LC_ALL=$locale");
     setlocale(LC_ALL, $locale);
 }


TrystanLea:
This broke things for me, I've reverted back

Twincarb:
@TrystanLea what version php are you using and what are the errors you are seeing?

reedy:
I'd be interest to know what broke too... `define()` is supposed to take a string

https://www.php.net/manual/en/function.define.php

Twincarb:
@reedy the error displayed is:
**Notice**:  Constant LC_MESSAGES already defined in **/var/www/emoncms/locale.php** on line **108**

I had a look around to get the language to change from english but couldn't, it would be good for someone else to confirm they can change language, I did look on the forum and there haven't been recent complaints of the language not working. It appeared that only en_GB is loaded for me regardless of the lengths I went to.

reedy:
I guess if it's set once during a session, it can't be unset/changed (without hacks that we shouldn't even entertain)

Based on the error message, `set_lang_by_user` is being called more than once, presumably erroneously.

You could wrap it in if a `if (!defined('LC_MESSAGES') { define('LC_MESSAGES', $locale); }` but this feels more like it's masking an error than actually fixing it... Plus, if things are depending on `LC_MESSAGES` potentially changing.. Well, that's wrong and can't ever actually happen being a constant

Certainly, it shouldn't be setting language once (to en) and then overriding it to anything else during the runtime, as that is/was never going to work...

So potentially, at a higher level, not going back into the `set_lang_by_user` function might be a better way forward

PR #1552, opened 2020-04-07
Update user table with extra column

Twincarb:
Specific column added to hyperlink to the users profile page.


TrystanLea:
Looks good!

PR #1554, opened 2020-04-08
service runner pulling ENV vars for docker-compose (v2)

emrysr:
these changes allow the service runner to take ENV vars for the redis hostname and port number.

this will enable the v2 branch of docker-compose repo to connect to the correct redis server (instead of the default 'localhost' and '6379'

will fallback to above defaults if no ENV vars set.

merge this only for testing... currently v2 branch uses my repos

### DIFF
--- a/version.json
+++ b/version.json
@@ -1,6 +1,6 @@
 {
     "name"         : "Emoncms Core",
-    "version"      : "11.4.0",
+    "version"      : "11.4.1",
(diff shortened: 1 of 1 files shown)


borpin:
The problem I see with this, is the specifying of the default in the code. This then breaks the link to the emoncms settings.

Can the documentation / readme be updated as part of the PR please.

PR #1557, opened 2020-04-10
add <form> to user login and profile edit

Twincarb:
The gods of google Console complain that the user form isn't enclosed in a <form> element, this PR fixes that for both the login_block.php and profile.php files.
Additionally a <br> was removed and the equivalent padding was added to the existing inline <style>
autocomplete"" have been added and filled in with reccommended values for assistance to password managers.

I am not sure why github has formatted the way it has, line indents were only adjusted in the code areas that were changed.
Can attempt to push again if needed.
Testing insturctions, check console in Chrome before and after applying the patch

### DIFF
--- a/Modules/user/login_block.php
+++ b/Modules/user/login_block.php
@@ -16,112 +16,114 @@ global $path, $settings;
 
 ?>
 <style>
-  .main {
-    max-width: 320px;
-    margin: 0 auto;
-    padding: 10px;
-  }
-  
+    .main {
+        max-width: 320px;
+        margin: 0 auto;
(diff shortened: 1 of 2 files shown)


Twincarb:
@TrystanLea I am not sure about this one now, there is an issue appearing on Chrome when using lastpass... It's fine on Firefox mind

TrystanLea:
Apologies for the delay, due to merge conflicts with the passage of time, I've hopefully applied the changes with the same intention here https://github.com/emoncms/emoncms/commit/e5e22fdf7617a6b10243e626505319b3303c347c

PR #1560, opened 2020-04-12
spelling error

Twincarb:
error in the css postion isn't valid.
To see ignored attribute goto feeds create a new virtual feed then select delete. In the modal that pops up head to modal-footer > feeds-to-delete


TrystanLea:
Thanks!

Issue #1567, opened 2020-04-22
Upgrade to a modern version of jQuery


reedy:
jQuery 1.11.3 is ancient. It's probably high time for an upgrade.

jQuery Migrate plugin might be helpful

https://jquery.com/upgrade-guide/3.0/
https://jquery.com/upgrade-guide/3.5/

Issue #1568, opened 2020-04-22
Upgrade jquery.flot


reedy:
Following on from https://github.com/emoncms/emoncms/issues/1567

https://github.com/emoncms/emoncms/blob/master/Lib/flot/NEWS.md says 0.8.3, and 4.2.0 is the latest release - https://github.com/flot/flot/releases

reedy:
It's unclear if 511e987438d94a9eb4a9be1873f118c13567ae0a made some sort of fork, or it was a partial upgrade of the files...

reedy:
https://github.com/emoncms/emoncms/blob/master/Lib/flot/date.format.js doesn't even seem to actually be a flot file, just dumped in the Lib/flot dir and merged into a combined file?

reedy:
The same seems to be the case for https://github.com/emoncms/emoncms/blob/master/Lib/flot/excanvas.js

date.format is merged into a combined file for flot... excanvas is not?

TrystanLea:
I think date.format.js ended up in the flot dir early on because it was included in some of the flot examples that the emoncms flot graphing was based on. We'd need to ammend the location for all optional modules as well if we change.

reedy:
Yeah. It'd make sense to have a place for "vendor" (ie 3rd party) code and "in house" code.

Generally, I wouldn't expect "in house" code to live in a folder called lib, when it mostly contains "vendor" code

PR #1569, opened 2020-04-24
Upgrade jQuery from 1.11.3 to 3.6.0

reedy:
Ping https://github.com/emoncms/emoncms/issues/1567

Definitely not saying this is anywhere near ready to merge :)

It might be more beneficial to do an incremental upgrade, stepping through a few versions. Let's see what happens


reedy:
Clicking "Setup" to hide/show the sidebar

EmonCMS Error
-------------
Message: Error: Syntax error, unrecognized expression: #
Route: Lib/jquery-3.5.0.min.js
Line: 2
Column: 13606

Which is obviously not the most useful in a minified file ;)

Other errors

Synchronous XMLHttpRequest on the main thread is deprecated because of its detrimental effects to the end user's experience. For more help http://xhr.spec.whatwg.org/

alexandrecuer:
@reedy: did you upgrade flot as well ?
http://github.com/flot/flot

reedy:
> @reedy: did you upgrade flot as well ?
> [http://github.com/flot/flot](https://github.com/flot/flot)

If you look at the commit content, no ;)

https://github.com/emoncms/emoncms/issues/1568 was filed for that specifically

reedy:
Also, there's many many other libraries that are on ooold versions :)

alexandrecuer:
Sorry :-)

reedy:
> @reedy: did you upgrade flot as well ?
> [http://github.com/flot/flot](https://github.com/flot/flot)

It seems this is going to be difficult. It's not easy to tell what is the original library, what's a fork, what's an addition to it...

3rd party changes (even by emoncms) shouldn't be in there, and especially not extra files and such...

alexandrecuer:
Do yu think the actual old "version" of flot can work with an updated version of jquery ?

reedy:
> Do yu think the actual old "version" of flot can work with an updated version of jquery ?

The newest can run on jQuery 1.2.6 (1.3.2 required)...

https://github.com/flot/flot#installation

Same as it was in https://github.com/flot/flot/blob/v0.8.3/README.md

It doesn't seem impossible :)

TrystanLea:
Thanks for testing @reedy I get the same error:

image

I see you changed to the non minified version, the line related to the "Sizzle.error" function and it looks like the most likely place that is then called from might be the selectors..

TrystanLea:
The old version of flot seems to be working, but clicking on any buttons in the dashboards gives the same error as above:

image

TrystanLea:
Same on the graph module, but otherwise navigation of feed selector and graph appears to be working

alexandrecuer:
I've tried to upgrade bootstrap as well :
bootstrap 5.1.0
jquery 3.6.0

feed view : 
image

input view is all black

no more error message in graph, but the calendar is no more working

image

alexandrecuer:
there is very less errors with :
bootstrap 5.1.0 with the old 2.3.2 bootstrap css
jquery 3.6.0

input view
image

feed view : the accordion on each feed cannot collapse 
image

graph : the feeds in view/statistics at the bottom cannot show / show csv button working fine
image

alexandrecuer:
This line in theme.php is not needed : 

https://github.com/emoncms/emoncms/blob/e31776090d66067e15d701f154969ce8596e5e86/Theme/theme.php#L36

since bootstrap 3, you dont need anymore the responsive css, everything is in the bootstrap.css

TrystanLea:
Thanks for your testing of the different bootstrap versions @alexandrecuer interesting that bootstrap 5.1.0 with the old css works. Hope to get on to this next week

alexandrecuer:
> Thanks for your testing of the different bootstrap versions @alexandrecuer interesting that bootstrap 5.1.0 with the old css works. Hope to get on to this next week

I made a mistake, it is with bootstrap 3.0.0 and not 5.1.0, with the old 2.3.2 css

bootstrap 3.0.0 was the first stable release that succeeded bootstrap 2.3.2

alexandrecuer:
when injecting the 3.0.0 css, things get tougher :
input module :
image

feed module : 
image

graph module :
image


So I guess we will have to follow the migration guide

https://getbootstrap.com/docs/3.4/migration/

TrystanLea:
Ok great thanks, that does sound like the way to go unless we drop bootstrap altogether as I discussed here https://github.com/emoncms/emoncms/issues/1661

I think at this point we need to take the action that is faster to implement, not sure if that's switching to bootstrap 3 or pulling out only the css and js that we use from bootstrap 2 into our own custom theming approach..

alexandrecuer:
yes my feeling is that going to bootstrap 3 and after to 4 and 5 will be a lot of work for a little profit. 

If the use of bootstrap is superficial, yes it could be nice to remove, the less dependancies you will have, the better it is for maintenance. I am not familiar enough with the view files to be able to appreciate how much it is used in the code. For me, understanding feed_view and input_view requires a certain amount of time.

By the way, which are the dependancies of menu_v3 ? None ? Jquery only ?

best

Alex

alexandrecuer:
Tried using **jquery 2.2.4** with keeping the old **bootstrap 2.3.2** : everything looks fine

https://github.com/jquery/jquery/releases/tag/2.2.4

Once going to jquery 3, the same problem reappear : cf graph module, when choosing a feed  `Syntax error, unrecognized expression: #`

alexandrecuer:
my tests are messy but I want to track them. It is difficult to find where this problem of `unrecognized expression: #` comes from

The # error disappear with :
**jquery 3.0.0**
**bootstrap-5.1.0.js with the old 2.3.2 css**
[frontend.zip](https://github.com/emoncms/emoncms/files/6947991/frontend.zip)

cf this screenshot of the graph module, working fine except the `feeds in view / show statistics`
image

But the modals in feeds module are broken 
feed_modal_bug

In the input module, one modal is working
image
**But the main modal, which configures the processing to the feeds, is broken**

alexandrecuer:
**jquery 3.6.0**
**bootstrap-3.3.7.js with the old 2.3.2 css**
everything seems fine, only the accordion in the feed view does not work nicely, as it always wants to stay opened. The accordion in the input module works fine. (Both should be managed same way)
[bootstrap-3.3.7-dist.zip](https://github.com/emoncms/emoncms/files/6949086/bootstrap-3.3.7-dist.zip)
[jquery-3.6.0.zip](https://github.com/emoncms/emoncms/files/6949088/jquery-3.6.0.zip)

TrystanLea:
Investigating the unrecognized expression: # error.

First it seems that the error catching function in emoncms.js that prints the errors as an alert (https://github.com/emoncms/emoncms/blob/master/Lib/emoncms.js#L78) obscures the actual line numbers relating to the error. By commenting this out and looking in the browser console I get a more helpful:

image

Line 746: https://github.com/emoncms/emoncms/blob/master/Lib/bootstrap/js/bootstrap.js#L746 
commenting that out gets rid of the error but breaks the dropdown menu

Searching for a related issue bought up this:
https://stackoverflow.com/questions/35539374/bootstrap-dropdown-jquery-uncaught-error-syntax-error-unrecognized-expression/49517080

which works! brilliant!

TrystanLea:
I've pushed this to the master branch https://github.com/emoncms/emoncms/commit/564f9f126a099161177a618525310ba5fa809831

TrystanLea:
The last time I looked for this fix I got stuck strange that it was actually such a simple fix in the end. I think we only use the dropdown part of the bootstrap js so it may be possible to only include the source for that component. I will double check

alexandrecuer:
Perfect :-)

PR #1573, opened 2020-04-24
Move date.format.js from flot folder

reedy:
Move it out of the flot/jquery.flot.merged.js file


TrystanLea:
Should I still merge this? happy to do so if you think its better to do so?

reedy:
PHP files still need updating

`<script language="javascript" type="text/javascript" src="<?php echo $path; ?>Lib/flot/date.format.min.js"></script>`

reedy:
Test failures should be fixed by https://github.com/emoncms/emoncms/pull/1928

PR #1574, opened 2020-04-24
Use minified jquery.flot.canvas.js in jquery.flot.merged.js

reedy:
Add newlines so comments from the next module aren't after the minified js of the first


reedy:
I don't think any of the actual issues in the Codacy review are new, it's just because I'm touching the file...

Issue #1336, opened 2019-06-28
Ensuring consistent code style


reedy:
I don't mind how code is styled (things like tabs vs spaces - #131), but it's nice if it's standardised, and also enforced by CI to prevent regressions and keep things from regressing

Is anyone against using PHPCS on this repo, and gradually enabling rules to define a coding standard, to make things consistent?

Something that does slightly concern me is the lack of braces being used on ifs... And hopefully we're all aware of the [problems](https://embeddedgurus.com/barr-code/2014/03/apples-gotofail-ssl-security-bug-was-easily-preventable/) this can cause elsewhere...

It'll be a bit disruptive to update the various files (that aren't 3rd party) to consistently use braces, but I think the benefit in the long run is worth it. Most rebasing should be relatively trivial

TrystanLea:
Thanks @reedy Im in the process of working through bringing two forks of emoncms together, the code base for emoncms.org and the main repo here, Id like to avoid large styling changes until that is done. 

Im happy to standardise on including braces, I've generally adhered to this but I know others favour missing them out where not strictly needed. Happy to receive pull requests with this change specifically.

reedy:
That's fair enough, I'm in no rush to do so, but I know from other experience it'll be worth it in the long run for readability and alike. Certainly other stuff that can be done, such as working on the subproject repos ala https://github.com/emoncms/wifi/pull/24 as well as cleanup here in the form of #1333, #1334 and #1335 to fix up various other issues... Stuff like this will probably keep me busy for a while, getting used to the codebases

I'm happy to submit other non code style type fixes for the time being as I'm going through the code to see what's what.

Like I say, a longer term goal of having things enforced by CI, and also automatically fixable too (`phpcbf` using phpcs rules)... Spaces and braces generally increase readability, but things like position of braces is YMMV and personal opinion, so are a bit more contentious, but at least being consistent is a good thing

Any idea when you might be done with that with the fork merges?

TrystanLea:
Great thanks @reedy how does PHPCS work, do you run it as part of the github travis testing? If its just providing an advisory rather than applying styling rules globally then thats fine. I prefer doing it that way so that I can see the changes being suggested/applied in a gradual step by step way.

reedy:
It will be runnable locally too, and for some rules, it'll have an "autofix" functionality too. You can define a set a rules for a repo, and then set exclusions (or not set the rules to begin with) on files etc.

[...]

PR #1585, opened 2020-04-25
Add PHPCS via composer

reedy:
Ping https://github.com/emoncms/emoncms/issues/1336


reedy:
This is the bare minimum to get it running. It'll no doubt fail, so we put rule amendments ontop :)

reedy:
Filed #1586 as travis isn't running here now...

But it is running in my fork

See https://travis-ci.org/github/reedy/emoncms/jobs/679410625 for example failures. It gave up because of there being so many, but we can narrow it; turning off rules, or scanning for certain files etc

reedy:
Ok, https://travis-ci.org/github/reedy/emoncms/jobs/679413070 is more representative

TrystanLea:
Great, maybe we should fix some of these before merging this in :)

reedy:
> Great, maybe we should fix some of these before merging this in :)

I would probably concur. And/or as a seperate commit in the same tree

I imagine we don't want to run it on the Lib directory either... But there's code in there (like with JS) that seems to be non 3rd party...

Twincarb:
Is this the right time to move to "php-parallel-lint/php-parallel-lint": "1.2.*" as the currently used one is marked as abandoned?

reedy:
> Is this the right time to move to "php-parallel-lint/php-parallel-lint": "1.2.*" as the currently used one is marked as abandoned?

It's completely unrelated, so should be a seperate patch IMHO :)

Noting, while the warning is (intentionally) scary, I don't think there's been any major bugs fixed, so not massively urgent.

In other projects we've been waiting for them to fix up the sub libraries to update namings, tag releases and bring it all together before updating - For example https://github.com/php-parallel-lint/PHP-Console-Highlighter/issues/1 and https://github.com/php-parallel-lint/PHP-Console-Color/issues/1

Though, this probably isn't a problem here... I'll make a patch in a bit :)

reedy:
> Though, this probably isn't a problem here... I'll make a patch in a bit :)

Done for all repos

reedy:
Mostly just needs a rebase and the package updating to `3.13.2`...

reedy:
Yeah, needs a config file to at least suppress things we're not going to fix (yet)...

reedy:
Why have you merged it when the tests are failing? 😅

reedy:
Followup coming

PR #1590, opened 2020-04-26
Add additional requirements

Twincarb:
Adds the requires to match those of the code content. redis not added as its not a required.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,4 +15,11 @@
         ]
     },
     "license": "AGPL-3.0-or-later"
+    "require": {
+    "ext-mysqli": "*",
+    "ext-gettext": "*",
+    "ext-json": "*",
+    "ext-curl": "*",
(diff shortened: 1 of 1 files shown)


reedy:
For reference in future when reviewing JSON... If GitHub shows a red/burgandy highlight, it's not valid
Screenshot 2020-04-29 at 19 40 04

reedy:
>Adds the requires to match those of the code content. redis not added as its not a required.

You can use suggests for that https://getcomposer.org/doc/04-schema.md#suggest

Twincarb:
Cheers @reedy I hovered over it but Github doesnt give a hint! If I had pushed from the desktop it would have had the comma, but I only copied and pasted before removing redis. 
Everydays a school day :)

reedy:
Tools like https://jsonlint.com/ can be useful too :)

PR #1592, opened 2020-04-26
How to add additional languages to the Pi

Twincarb:
I am more used to using the command line that a gui so was scratching my head as to why I coulndn't install extra languages for testing. It's possible that this will be helpful to someone else.
I am not certain if the Pi needs rebooting after the languages are installed as I didn't check before doing a reboot. If its needed it can be added in.


TrystanLea:
Thanks @Twincarb much appreciated

PR #1599, opened 2020-05-06
Timezone support csv export

TrystanLea:
- Provides timezone support in PHPFina and PHPTimeSeries csv export API's
- Provides option to request averages for both engines
- Implements single get_data_combined method used by all data request types to reduce code duplication


Modules/feed/feed_model.php
```diff
@@ -668,37 +664,25 @@ public function get_average_DMY($feedid,$start,$end,$mode)
         if (!$this->exist($feedid)) return array('success'=>false, 'message'=>'Feed does not exist');
         
         $engine = $this->get_engine($feedid);
-        if ($engine!=Engine::PHPFINA && $engine != Engine::MYSQL ) return array('success'=>false, 'message'=>"This request is only supported by PHPFina AND MySQLTimeseries");
-
-        // Call to engine get_data
-        $userid = $this->get_field($feedid,"userid");
-        $timezone = $this->get_user_timezone($userid);
+        if ($engine!=Engine::PHPFINA && $engine != Engine::PHPTIMESERIES && $engine != Engine::MYSQL ) return array('success'=>false, 'message'=>"This request is only supported by PHPFina AND MySQLTimeseries");
```
chaveiro, line 667: Why not let the engine class decides if it supports or not this?

borpin:
I've not used the CSV export directly, but does it allow you to pull off the *actual* data stored?

chaveiro:
Hi Trystan, do you think it's possible to ensure that the timeformat also works with mysql engine?

borpin:
Coincidentally, this thread popped up on the Grafana community. 

https://community.grafana.com/t/mysql-datasource-timezone-issue/29384/5

Never knew you could convert an epoch to date Time on command line. Every day a learning day.

TrystanLea:
> but does it allow you to pull off the actual data stored?

Yes already for PHPFina, still need to add that in for PHPTimeSeries, there are a couple of things left to do that I should have noted above. The introduction of averaging on PHPTimeSeries is a bit slow at the moment ~1.0-1.5s compiling daily data over say half a year, I need to see if I can optimise that a little bit more.

> Hi Trystan, do you think it's possible to ensure that the timeformat also works with mysql engine?

I'm not familiar with the get_data function on the mysql engine any more so I would need yours or @adminde's help with that :)

TrystanLea:
I've been looking at the question of exporting original data with PHPTimeSeries @borpin, here are a couple of examples with different export settings:

Original data export

```
1589360575 | 300
1589360584 | 100
1589360595 | 150
1589360605 | 250
1589360612 | 550
1589360620 | 520
1589360635 | 120
1589360659 | 120
1589360665 | 20
1589360675 | 22
```

10s fixed interval:

```
1589360560 | null
1589360570 | 300
1589360580 | 300
1589360590 | 100
1589360600 | 150
1589360610 | 250
1589360620 | 520
1589360630 | null
1589360640 | 120
1589360650 | null
1589360660 | 120
1589360670 | 20
1589360680 | null
```

30s fixed interval:

```
1589360520 | null
1589360550 | 300
1589360580 | 300
1589360610 | 250
1589360640 | 120
1589360670 | 20
1589360700 | null
```

Does it make sense to provide 30s interval at original timestamps??

The following is the result of the nearest datapoint (within one 30s interval) to the allocated time interval in column 1, the actual timestamp is given in column 2 for comparison:

```
1589360550 | 1589360570 | 300
1589360580 | 1589360584 | 100
1589360610 | 1589360612 | 550
1589360640 | 1589360635 | 120
1589360670 | 1589360665 | 20 or 22
1589360700 | 1589360675 | 22
```

TrystanLea:
If PHPTimeSeries is being used to record changes of state on an infrequent basis then the any periodic interval approach to fetching the data may fail. In this case doing an original data export with a limit on the maximum number of datapoints that you can export may be the best approach

TrystanLea:
Comparison across feeds, on the fly calculations such as those in the my solar pv battery app and stacked graphs etc all require the data to be returned at the same intervals across the feeds selected, hence the focus on supporting fixed interval data in emoncms. 

Recording state changes using PHPTimeSeries probably needs a ground up approach to visualisations etc as the system isnt really designed to cater for that..

borpin:
Could the export not get the feed interval from the meta data and use that?

I thought timeseries data was by definition periodic, with an offset from the start (unless I have misunderstood something).

The reason for it is to see where a variation in the input timing might be causing gaps in the Feed data.

Lib/enum.php
```diff
@@ -26,7 +26,7 @@ class Engine
     const GRAPHITE = 3;      // Not included in core
     const PHPTIMESTORE = 4;  // Depreciated
     const PHPFINA = 5;
-    const PHPFIWA = 6;
+    const PHPFIWA = 6;       // Depreciated
```
borpin: "Deprecated" https://www.google.co.uk/search?q=Deprecated

PR #1600, opened 2020-05-06
new input process: kwh per half hour

TrystanLea:
A new input process in testing to produce kWh per half hour data, similar to that used for billing by smart meters and tariffs such as Octopus Agile.


washcroft:
Looking to use this but it isn't behaving as expected, i.e. no real time view of the current half hour.

Any particular reason you're using insert_data for the previous half hour rather than update_data for the current half hour - just like how the kwh per day process works?

rjsc:
May I suggest a kwh per configured time block?
In Portugal, the smart meters are configured to do net metering within 15 min blocks.
If the total is negative (exported energy) in a block it counts a zero. Any energy you export, you must take back from the grid in the same 15 minute block or else you loose it.
In other countries the period may be different.

chaveiro:
Done in https://github.com/emoncms/emoncms/pull/1922

TrystanLea:
replaced with @chaveiro 's implementation, now available in the master branch

PR #1601, opened 2020-05-06
Show full table of emoncms core and module details

TrystanLea:
Following ongoing discussion on two emoncms administration features that would be useful: the ability to switch between stable and master branches from the admin UI and the ability to see when updates are available PR https://github.com/emoncms/emoncms/pull/1468. 

One option could be to show a table of emoncms components installed with the option to select the branch and update inline with each component. Im not 100% sure whether this fine grained approach is a good idea as module updates and core updates can often be intertwined, but showing an extended table of module information could still be useful.

Ideally the list would include components like EmonScripts, emonhub etc rather than emoncms and modules folder only..

A concept idea for now, for further discussion, not to be merged.

A key issue with the implementation here is that it doesnt work so well with the copy to clipboard markdown format. Im not sure how that part works, so would be happy for input on this.

image


borpin:
Looks useful. The lack of an entry in Git Describe column indicates no annotated tags?

borpin:
Perhaps for a secondary page, but it would be useful when working on issues, to know if a users tree is clean so an update has worked (or why it hasn't).  There are a few examples out there, this one for instance from BASH https://unix.stackexchange.com/questions/155046/determine-if-git-working-directory-is-clean-from-a-script has various commands to check the status.

I think it would be worth pulling in the version file for the branch off GitHub which would show if there was an update or not.

Would need **all** version files, for every branch to be correct to be reliable of course 😀.

chaveiro:
Looks very good and a welcome feature for many years!

1 - Does it also show available (official) modules for installation that are not currently installed ?

2 - You can add support for non official modules by let user enter in a text box that module git URL. And so it' would become known and supported on that installation. What do you think?

TrystanLea:
Thanks, yes missing tags and yes would be great to show if there are local changes that will block updates.

It doesnt show modules not installed, or provide a button click install, its a long long way from an installation manage it could be :)

2 & 3 perhaps yes, I was thinking it could integrate into the update section rather than the server information section and perhaps just leave the server information section as is, for a compact form that users can copy and paste for support..

TrystanLea:
Im not sure if providing per module branch switching and updates would create more problems than its worth.. 1. Its a lot of work to implement. 2. Branches may need switching in unison to work given requirements across modules and core and also changes to modules that require services to be restarted e.g emoncms_mqtt, demandshaper

Maybe its better to just provide this more detailed information, but only allow interface switching between stable and master branch..

chaveiro:
Thinking on this a beter approuch is to provide tag switching and use the tag as the stable release version .

borpin:
> Thinking on a beter approuch for switching, i think provide tag switching is ideal

Possibly, but Master does not have a tag so that would always need to be available. In addition, where specific testing may be happening in a dedicated Branch, being able to switch to that would be useful.

TrystanLea:
> Providing a list of available modules is a mater of reading a json file that should be maintained in core repo. Not very difficult and only for the official ones.

Yes I think some kind of central way of pulling in all the different module versions etc would probably be needed to keep things running smoothly, rather than making loads of requests per module, make a single request to a fast API end point that returns all the information needed may be better..

This should also all happen on the client side on the admin / update UI page asynchronously so as not to slow down the page load..

chaveiro:
> Possibly, but Master does not have a tag so that would always need to be available.

Tags purpose is that developers only tag a master once it's stable and important to be made an official versioned/tagged release ready for production and allow regular users to update. 

For developing it's another usecase and using git manualy already does the trick.

TrystanLea:
I think the admin main view page really needs rebuilding using a client side javascript approach to move forward with this and I need to get a better understanding of the html to markdown conversion. I will discuss with Emrys and come back to this one.

TrystanLea:
What are your views on the formatting of the markdown output? E.g the old approach in post 5 here https://community.openenergymonitor.org/t/log-emonpi-remotely-doesn-t-work-for-me-again/12569/5 

and the new approach in post 3 here: https://community.openenergymonitor.org/t/emon-mqtt-data-incomplete/10964/3

Was that all discussed somewhere and I missed it?

borpin:
Newer version is easier as the older one scrolls off the right and doesn't wrap.

Twincarb:
Would the idea of the admin page being more of a dashboard be a useful idea. If the normal control is from a tablet device clicking on "Setup" would open the dashboard which holds basic information only as well as items for links to the different areas. At the moment even on a desktop the page is long.
Potentially being bs4 cards to logical areas and the current states of the server only showing services (you then see if something has stopped) and view showing the disk space and memory useage.
The copy to clipboard, could be replaced with a single button to "collect system information" which collects all the information required to allow it to be posted to the forum. example from Joomla:
[https://github.com/ForumPostAssistant/FPA](url)
Everything else contained in its own view, if the system notifies me there are updates then I go to the updates page, here I can see the different options for updating as well as the current states of the "packages/modules/cms" as is showing here.
A seperate page dedicated to the modules which would then be a stepping stone to having the ability to show what other modules are then available. This list being stored on a central server which means anyone writing a custom module hosted on github would be able to submit a request to have their module added to the list. (this is similar to @chaveiro  suggestion)

TrystanLea:
This has been replaced with a new implementation in the latest version.

PR #1623, opened 2020-08-20
Fix URL Encoding of Input API example with time to close #1621

### LINKED ISSUE #1621: Input API example with the time in JSON doesn't work
I have been trying to use the second example on the Input API Help page to provide the time along with the data values in JSON:

https://myserver.example.com/emoncms/input/post?node=emontx&fulljson={"power1":100,"power2":200,"power3":300,"time":"2020-08-17T18:37:12+01:00"}

This doesn't appear to work on my server. I end up with 4 data values under the "emontx" node: power1, power2 and power3 as expected but also a "time" data value with the value "2020-08-17T18:37:12+01:00".

If I provide a time parameter it works as expected. For example:
https://myserver.example.com/emoncms/input/post?node=emontx&time=1597686335&fulljson={"power1":100,"power2":200,"power3":300}

davefiddes:
This fixes Issue #1621 by URL encoding the Input API example
which uses a text string for the date and time. The date
format used includes the current system timezone which
contains a "+" character which, if not URL encoded, will be
interpreted as a space. This causes a failure in the time
parsing.

The link in the live example will now behave correctly when clicked.


TrystanLea:
Thanks @davefiddes !

PR #1632, opened 2020-09-25
Possible solution to Cumulative totals resetting

borpin:
I have seen this happen to me; i.e. the cumulative process resetting to zero.

Community discussion

https://community.openenergymonitor.org/t/correcting-cumulative-kwh-reset-with-post-process-module/4631/12

nicsergio suggested a fix to the get_timevalue() function. I have pulled that in here for discussion.

As it relates to one of your comments @chaveiro perhaps you could look at the logic.  I am **not** suggesting this is correct just the user reports that the issue stopped for him when using this code.

Another user has noted this happening as well. https://community.openenergymonitor.org/t/how-is-post-processing-supposed-to-work/15478

### DIFF
--- a/Modules/feed/feed_model.php
+++ b/Modules/feed/feed_model.php
@@ -542,6 +542,10 @@ class Feed
                     $lastvalue['value'] = (float) $lastvalue['value'];
                 }
                 // CHAVEIRO comment: Can return NULL as a valid number or else processlist logic will be broken
+                // Test for avoid to reset PowerToKWh process feed
+                if (is_null($lastvalue['time']) || is_null($lastvalue['value'])) {
+                    $lastvalue = $this->EngineClass($engine)->lastvalue($id);
+                }
             } else {
                 // if it does not, load it in to redis from the actual feed data because we have no updated data from sql feeds table with redis enabled.
                 $lastvalue = $this->EngineClass($engine)->lastvalue($id);


TrystanLea:
Thanks @borpin and sorry for the slow reply. Just noting that I've seen this at this point. Dont want to add it in without careful consideration, its strange that the redis entry exists and is then incorrect. You would expect the redis entry to not exist and for the last value to then be fetched from the feed data.. but it sounds like it's getting out of sync somehow..

borpin:
I think the point is it can exist, but be an invalid value - if that wasn't true the checks would not already exist.

Trouble is, currently, all the code does is set the time and value to `null`.

This code attempts to then recover the situation.

lechercheur123:
I have this kind of issue when I reboot my Emoncms instance (I host it on a VM). When I reboot, all my accumulators reset.

Capture

If I download the data, it looks like this:
```
1642285800, 58.4
1642286700, 58.4
1642287600, 58.4
1642288500, 58.4
1642289400, 58.4
1642290300, 0.0
1642291200, null
1642292100, null
1642293000, null
```

I will try the proposed modification.

And2IT:
This issue (cumulative totals resetting) happened to me also.
I resolved with the proposed original solution from @nicsergio (as written in the first post), just adding these few lines in feed_model.php file, starting at row 535: 

```
                // CHAVEIRO comment: Can return NULL as a valid number or else processlist logic will be broken

                // Test for avoid to reset PowerToKWh process feed
                if (is_null($lastvalue['time']) || is_null($lastvalue['value']))
                	$lastvalue = $this->EngineClass($engine)->lastvalue($id);

```

I also opened a pull request: #1818 

This solution works very well for me, I haven't found any side effect. I kindly ask to add it to the master branch, to give benefit to all users.

And2IT:
> I have seen this happen to me; i.e. the cumulative process resetting to zero.
> 
> Community discussion
> 
> https://community.openenergymonitor.org/t/correcting-cumulative-kwh-reset-with-post-process-module/4631/12
> 
> nicsergio suggested a fix to the get_timevalue() function. I have pulled that in here for discussion.
> 
> As it relates to one of your comments @chaveiro perhaps you could look at the logic. I am **not** suggesting this is correct just the user reports that the issue stopped for him when using this code.
> 
> Another user has noted this happening as well. https://community.openenergymonitor.org/t/how-is-post-processing-supposed-to-work/15478

Hello @borpin ,
as reported in my pull request #1824 also , @nicsergio 's solution works perfectly, and without this change, every cumulative kWh is going to be reset to zero after each reboot.
What can we do to ease the merge of this code into the main emoncms project? 
It's very annoing to have to deal with this issue. At every emoncms upgrade, I need to manually modify the feed_model.php file.
Thank you,
Andrea

borpin:
@TrystanLea

And2IT:
Hello @borpin 
I have just updated my Raspeberry from Emoncms 11.2.3 to 11.3.0 (I have removed the custom code by @nicsergio) and now cumulative processes are not anymore resetting to zero, also after a reboot !!!
Can you confirm that is not a surprising piece of luck but is a real correction of the issue ?
Thank you,
Andrea

nicsergio:
> > I have seen this happen to me; i.e. the cumulative process resetting to zero.
> > Community discussion
> > https://community.openenergymonitor.org/t/correcting-cumulative-kwh-reset-with-post-process-module/4631/12
> > nicsergio suggested a fix to the get_timevalue() function. I have pulled that in here for discussion.
> > As it relates to one of your comments @chaveiro perhaps you could look at the logic. I am **not** suggesting this is correct just the user reports that the issue stopped for him when using this code.
> > Another user has noted this happening as well. https://community.openenergymonitor.org/t/how-is-post-processing-supposed-to-work/15478
> 
> Hello @borpin , as reported in my pull request #1824 also , @nicsergio 's solution works perfectly, and without this change, every cumulative kWh is going to be reset to zero after each reboot. What can we do to ease the merge of this code into the main emoncms project? It's very annoing to have to deal with this issue. At every emoncms upgrade, I need to manually modify the feed_model.php file. Thank you, Andrea

I'm interested too

nicsergio:
> Hello @borpin I have just updated my Raspeberry from Emoncms 11.2.3 to 11.3.0 (I have removed the custom code by @nicsergio) and now cumulative processes are not anymore resetting to zero, also after a reboot !!! Can you confirm that is not a surprising piece of luck but is a real correction of the issue ? Thank you, Andrea

I'm interested too

borpin:
@nicsergio @And2IT 
Are you sure the file was updated? If you didn't manually do the git update, git will not update the local changes.

nicsergio:
> @nicsergio @And2IT Are you sure the file was updated? If you didn't manually do the git update, git will not update the local changes.

I haven't updated yet, I was following what was reported by @And2IT

And2IT:
Sorry to all, I did a mistake saying that version 11.3.0 solved the issue. I confirm that 11.3.0 HAS the issue.
I need to add the @nicsergio mod if I want not to loose the cumulative values.
I think too may years have passed since today, I believe it is important to fix this wrong behaviour that affect each cumulative value (e.g. Power-to-kWh).
Please, merge the "historical" mod from @nicsergio into master project. I have been used it since months with huge benefit and no drawback. Until the mod is not added to master project, I need to manually modify the file feed_model.php each time there is an update in the emoncms, and every other user has to do this if doesn't want to loose data.
Thank you very much for undestanding this need and thank you for who of you works hard on emoncms.
Andrea

lechercheur123:
> I confirm that 11.3.0 HAS the issue.

I also confirm this.

@nicsergio fix is still needed and is still working

borpin:
@TrystanLea Bump.

lechercheur123:
Hello,
This seems fixed in 11.3.22

And2IT:
@borpin 
I have installed Emoncms version 1feb24 just now, and it seems to have solved this problem.
Any other that can confirm ?
Thank you

PR #1637, opened 2020-10-12
Empower EmonCMS with Redis storage of forecasting datas coming from various API

alexandrecuer:
This PR is just a proposition....

My idea was to extend the Redis Engine in order to be able to store "small" feeds in Redis, which would become something else than a buffer... I have a solution which works fine and is not introducing complexity into the code, from my point of view

My actual usecase is retrieving meteo forecasting coming from openweathermap or others API, in order to be able to predict the evolution of temperature in a building using a model (electrical or statistical)

graph_w_prev

on the image you can see the "cold" feed in blue, coming from a PHPFINA feed and the "hot" forecast in orange, stored as a temporary redis feed, which you can drop when you like using the emoncms feed control panel.....

everything is explained with details here : 
https://github.com/dromotherm/sandbox/blob/master/enhancingEmonCMS.md
and [here](https://github.com/dromotherm/favorites/blob/master/images/redis%20feed%20in%20emonCMS.pdf)

for me, it is interesting as you will attract people coming from machine learning
only one dependancy : [numpy](https://numpy.org/) which you can easily install via apt-get

I've integrated a small python script in the commit, which yu can use to sniff the openweathermap API
https://github.com/alexandrecuer/emoncms/commit/5e2fa44a32ca1d99e7d18df4ea12969b7d8a5f57
all you have to do is to to fill the OWM key in a conf file...

in command line, fill the OWM key
```
cd /var/www/emoncms/scripts/examples
nano forecast.conf 
```
then launch the sinffing process : 
```
./get_meteo_forecast.py
```
I will fully understand that this may not at all be in the roadmap, but I wanted to share, because I think it is very interesting....I know some of you are maybe doing this with nodered but here the advantage is that the feed is not managed as a historical timeserie such PHPFINA, and that you will benefit from the power of numpy
 
@TrystanLea : if you find this not suitable, I will remove the PR and keep the evolution on my themis fork

best

Alex

### DIFF
--- a/Modules/feed/engine/RedisBuffer.php
+++ b/Modules/feed/engine/RedisBuffer.php
@@ -12,7 +12,7 @@ class RedisBuffer implements engine_methods
     private $log;
     private $redis;
     private $feed;
-    
+
     /**
      * Constructor.
      *
@@ -34,7 +34,13 @@ class RedisBuffer implements engine_methods
     public function delete($feedid)
     {
         $this->redis->srem("feed:bufferactive",$feedid); // remove from feedlist
-        $this->redis->zRemRangeByRank('feed:$feedid:buffer', 0, -1); // remove buffer
+        $bkey="feed:$feedid:buffer";
+        // remove buffer
+        if ($this->redis->type($bkey)==1){
+            $this->redis->del($bkey);
+        } else {
+            $this->redis->zRemRangeByRank($bkey, 0, -1);
+        }
     }
 
     public function get_meta($feedid)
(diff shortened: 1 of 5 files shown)


TrystanLea:
Thanks for putting this up here as a proposition and for time that you have evidently put into documenting the feature @alexandrecuer. The link to numpy sounds interesting but Im still not quite understanding what the batch idea is about. Can the numpy batch not be converted to the existing redis buffer format and back again? 

I'm keen to focus for a period on making the core of emoncms more lean, ideally I would like to avoid different API end points that only work for specific feed engines and some of the hacks I've been adding in to skip the buffers etc for updating existing feed data. If you can see a way this feature could be introduced in a modular way that might be worth considering?

alexandrecuer:
OK I will keep that on my fork :-)

Issue #1624, opened 2020-08-20
JSON input with a timestamp is stored as a value along with the actual input values


davefiddes:
Discovered as part of investigating Issue #1621. When a JSON input is supplied with a specific timestamp such as:
`{"power1":100,"power2":200,"power3":300,"time":"2020-08-20T12%3A10%3A15%2B01%3A00"}` the system stores 4 values power1, power2, power4 and time with the timestamp indicated by the "time" field. This doesn't make a lot of sense to me as it clutters the input process with spurious data values.

Steps to reproduce (assuming fix for #1621 is merged):
1. Navigate to the Inputs view
2. Delete the "emontx" node if present
3. Navigate to the Inputs | Input API Help page
4. Click on the "JSON format - with time (as a string in this example)" example link
5. Navigate back to the Inputs view

Expected Outcome:
1. A new input node "emontx" has been created
2. Three data values have been created: power1, power 2, power3 with a data timestamp of a few seconds previously from when the Input API Help page was displayed and values 100, 200 and 300 respectively

Actual Outcome:
1. A new input node "emontx" has been created
2. Four data values have been created: power1, power2, power4 and time with the values and timestamps as the expected outcome
3. The "time" data value is "NULL"

Alternative steps:
 - For Step 3 replace the time value string with a UNIX time_t value such as  `{"power1":100,"power2":200,"power3":300,"time":1597923351}`
 - The outcome is that the "time" value stored is `1597923351` rather than `NULL`. The behaviour is otherwise identical

borpin:
Yes, `time` element should be stripped from the input data

https://github.com/emoncms/emoncms/blob/611fd01be39052b6aa5688ce9215aae0d428f795/Modules/input/input_methods.php#L109-L109

I think 

```
unset($jsondataLC['time']);
```
will do it. Not got time to test right now.

borpin:
I'd missed you had done PR #1623. Thanks.

42Network:
Any idea why that bit of new code for the unset fails now for me?


```
[Thu Jan 28 15:05:46.578649 2021] [php7:notice] [pid 50986] [client 10.1.1.6:63906] PHP Notice:  Undefined variable: jsondataLC in /var/www/emoncms/Modules/input/input_methods.php on line 60
[...]

PR #1638, opened 2020-10-20
Remove Time from input array

borpin:
*Breaking change for anyone logging time*

re #1624 

Remove the time element from the array so no Input is created.

If the time is a string, then a N/A appears in the Inputs

If a user wants to log time it can be added as an additional JSON field but not called `time`.


TrystanLea:
Thanks @borpin I imagine this will not affect many users, and a feed with the updated time is not likely that critical in terms of other parts of the processing chain anyway

PR #1639, opened 2020-10-21
Add support for secure mqtt connections

f4grx:
Hello,

this commit adds support for secure mqtt broker connections.

Thanks for merging.

### DIFF
--- a/Lib/process_old_settings.php
+++ b/Lib/process_old_settings.php
@@ -47,7 +47,11 @@ $settings = array(
     'user'      => isset($mqtt_server["user"])?$mqtt_server["user"]:'',
     'password'  => isset($mqtt_server["password"])?$mqtt_server["password"]:'',
     'basetopic' => isset($mqtt_server["basetopic"])?$mqtt_server["basetopic"]:'emon',
-    'client_id' => isset($mqtt_server["client_id"])?$mqtt_server["client_id"]:'emoncms'
+    'client_id' => isset($mqtt_server["client_id"])?$mqtt_server["client_id"]:'emoncms',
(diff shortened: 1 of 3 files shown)


TrystanLea:
Thanks @f4grx thats great!

TrystanLea:
I made a couple of small modifications after testing.

borpin:
I would have expected an example setup in the `default.settings.ini` file. How will anyone know how to use it other wise?

slorquet:
Sure. Do you require a pull request or can you just apply this?
```
diff --git a/default-settings.ini b/default-settings.ini
index 14b85e40..0b7201bc 100644
--- a/default-settings.ini
+++ b/default-settings.ini
@@ -54,6 +54,18 @@ user = ''
 password = ''
 basetopic = 'emon'
 client_id = 'emoncms'
+; secure MQTT parameters
+; capath must be defined to enable SSL usage.
+; This is the only setting required to connect to a SSL mqtt broker.
+;capath = '/etc/ssl/certs'
+; The following settigs are related to your CLIENT certificate, if your broker requires client authentication.
+; certpath must point to your client certificate
+;certpath = '/etc/letsencrypt/live/yoursite/fullkey.pem'
+; keypath must point to your client private key
+;keypath = '/etc/letsencrypt/live/yoursite/privatekey.pem'
+; keypw is the client private key password.
+; do not define if your private key is not password protected
+;keypw = ''


 ; Input
```

TrystanLea:
Thanks @slorquet a pull request would be super helpful if possible, thanks

slorquet:
Here it is
https://github.com/emoncms/emoncms/pull/1649

PR #1641, opened 2020-10-24
Fixes #1640

### LINKED ISSUE #1640: Admin page MySQL host is incorrect
The Admin page's MySQL host value is incorrect.  Instead of the MySQL hostname and IP, the page lists the redis server's hostname and IP.  Offending code is in Modules/admin/admin_main_view.php, line 371 in current master:

`<?php echo row(_('Host'), $system['redis_server'] . ' (' . $system['redis_ip'] . ')'); ?>`

References to `$system['redis_server']` and `$system['redis_ip']` should be `$system['db_server']` and `$system['db_ip']` respectively.

I've tested this on my own and the above seems to fix the issue.

Joe

joemcool:
Admin page MySQL server info area was displaying Redis host as MySQL
host.  This changes the key read from $system from 'redis_server' and
'redis_ip' to 'db_server' and 'db_ip'.


TrystanLea:
Thanks @joemcool

PR #1644, opened 2020-11-17
Update core.php

mikbund:
When using docker and apache as reverse proxy
 RequestHeader set HTTP_X_FORWARDED_PROTO https
is not passed on to $_SERVER['HTTP_X_FORWARDED_PROTO'] hence emoncms remains in state HTTP with $proto as default set to https.

This fix resolves this issue by adding a check for reading request_headers and not $_SERVER variables.


TrystanLea:
Thanks @mikbund

PR #1647, opened 2020-12-29
Create scale input process

nev7n:
Create new input process that scales current value by 10^input.  

I needed this for processing input retrieved from modbus over TCP from a Solaredge inverter.  It returns values along with a scale factor that has to be applied to each value, for example:

AC Power value: 6090
Scale factor: -1
AC Power (W) = 6090 x 10^-1 = 609

Not sure how common a requirement this is but thought would make a pull request in case other people would find it useful.

### DIFF
--- a/Modules/process/process_processlist.php
+++ b/Modules/process/process_processlist.php
@@ -891,7 +891,19 @@ class Process_ProcessList
               "unit"=>"",
               "group"=>_("Input"),
               "description"=>_("<p>Scales the current value by ten to the power of the other input.</p>")
-           )
+           ),
+           array(
+            "id_num"=>61,
+            "name"=>_("value and scale"),
+            "short"=>"v+s",
+            "argtype"=>ProcessArg::NONE,
+            "function"=>"value_and_scale",
+            "datafields"=>0,
+            "datatype"=>DataType::UNDEFINED,
+            "unit"=>"",
+            "group"=>_("Calibration"),
+            "description"=>_("<p>Unpack combined value and scale factor, applying the scale factor.</p>")
+         ),
         );
         return $list;
     }
@@ -979,6 +991,21 @@ class Process_ProcessList
     public function abs_value($arg, $time, $value)
     {
         return abs($value);
+    }
+
+    public function value_and_scale($arg, $time, $value) {
+        $scale = $value & 0xFFFF; // Right 16 bits are the scale
+        $val = $value >> 16; // Left 16 bits are the value
+
+        if($scale >= 0x8000) { // Decode twos complement if negative
+          $scale = -(($scale ^ 0xFFFF)+1);		
+        }
+        
+        $val = $value >> 16; // Left 16 vits are the value
+        
+        $result = $val * pow(10, $scale);        
+
+        return $val * pow(10, $scale);
     }
 
     //---------------------------------------------------------------------------------------
--- a/Modules/process/process_processlist.php
+++ b/Modules/process/process_processlist.php
@@ -882,10 +882,10 @@ class Process_ProcessList
            ),
            array(
               "id_num"=>60,
-              "name"=>_("scale input"),
+              "name"=>_("pow10 input"),
               "short"=>"s inp",
               "argtype"=>ProcessArg::INPUTID,
-              "function"=>"scale_input",
+              "function"=>"pow10_input",
               "datafields"=>0,
               "datatype"=>DataType::UNDEFINED,
               "unit"=>"",
@@ -993,6 +993,12 @@ class Process_ProcessList
         return abs($value);
     }
 
+    //---------------------------------------------------------------------------------------
+    // Combines a value and scale input, multiplies the value by 10^scale
+    //
+    // These split value and scale inputs are output from the Modbus interface of solar edge
+    // inverters.
+    //---------------------------------------------------------------------------------------
     public function value_and_scale($arg, $time, $value) 
     {
         $scale = $value & 0xFFFF; // Right 16 bits are the scale
@@ -1000,7 +1006,7 @@ class Process_ProcessList
           $scale = -(($scale ^ 0xFFFF)+1);		
         }
         
-        $val = $value >> 16; // Left 16 vits are the value
+        $val = $value >> 16; // Left 16 bits are the value
 
         // Apply scale factor to the value
         return $val * pow(10, $scale);
@@ -1015,9 +1021,9 @@ class Process_ProcessList
     }
 
     //---------------------------------------------------------------------------------------
-    // Scales current value using another input.  Current value is multiplied by 10^input
+    // Scales current value by 10 to the power of another input.
     //---------------------------------------------------------------------------------------
-    public function scale_input($id, $time, $value)
+    public function pow10_input($id, $time, $value)
     {
         $input_value = $this->input->get_last_value($id);
         return $value * pow(10, $input_value);
--- a/Modules/process/process_processlist.php
+++ b/Modules/process/process_processlist.php
@@ -993,18 +993,16 @@ class Process_ProcessList
         return abs($value);
     }
 
-    public function value_and_scale($arg, $time, $value) {
+    public function value_and_scale($arg, $time, $value) 
+    {
         $scale = $value & 0xFFFF; // Right 16 bits are the scale
-        $val = $value >> 16; // Left 16 bits are the value
-
         if($scale >= 0x8000) { // Decode twos complement if negative
           $scale = -(($scale ^ 0xFFFF)+1);		
         }
         
         $val = $value >> 16; // Left 16 vits are the value
-        
-        $result = $val * pow(10, $scale);        
 
+        // Apply scale factor to the value
         return $val * pow(10, $scale);
     }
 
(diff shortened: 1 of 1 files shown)


borpin:
Why not just use the multiply process?

nev7n:
Hi @borpin 
I didn't just need to multiply the value by another input, but 10 to the power of another input. I don't think the multiply process would let me do that?

I modified my solution this morning because I occasionally the scale and value would get out of sync resulting in the wrong scale being applied.  Instead of sending two inputs - one with a int16 scale factor and one with a int16 value I now have one input with a int32 combining the two.  I now unpack the scale and value from this and apply the scale in one function which fixes the problem:
```
    public function value_and_scale($arg, $time, $value) {
        $scale = $value & 0xFFFF; // Right 16 bits are the scale
        if($scale >= 0x8000) { // Decode twos complement if negative
          $scale = -(($scale ^ 0xFFFF)+1);		
        }

        $val = $value >> 16; // Left 16 bits are the value

        return $val * pow(10, $scale);
    }
```
This forum post documents what I was doing and why:
https://community.openenergymonitor.org/t/logging-solaredge-inverter-data-using-modbus-over-tcp/16341/3

Realise this is probably a fairly uncommon thing to want to do so please reject the pull request if not likely to be useful for other people.  I know solar edge inverters are fairly popular, not sure how many other systems use separate value and scale figures like this.

alexandrecuer:
@nev7n : are you using emonhub ? if so I think you can use the scale parameter in emonhub : 
 example with the emonhub modbusTCP interfacer :
```
[[12]]
    nodename = sofrel_Text
    [[[rx]]]
       names = Text_nord,Text_cellule,Text_sud,Text_hallssol,Text_est,Text_ouest
       registers = 54632,54604,54662,54686,54722,54746
       unitId = 3
       datacode = f
       scale = 0.1
```

nev7n:
@alexandrecuer thanks for the suggestion, however the scale I need to use varies depending on the value of another register so I don't think I can configure emonhub like that.  These are the registers I'm pulling from the inverter using ModBusTCP:
```
[[ModbusTCP]]
    # this interfacer retrieves register information from modbusTCP clients 
    # retrieve register information from modbus TCP documentation for your inverter.
    # Information here is designed for Fronius Symo 3 phase inverter.
    Type = EmonModbusTcpInterfacer
    [[[init_settings]]]
        modbus_IP = 192.168.178.36   # ip address of client to retrieve data from
        modbus_port = 502          # Portclient listens on
    [[[runtimesettings]]]
        # list of names of items being retrieved
#        rName = DC_Current_1,DC_Voltage_1,DC_Current_2,DC_Voltage_2,Power,AC_Current,Frequency,Temperature,Residual_Current,Total_Yield,Daily_Yield
        rName = I_AC_Current,I_AC_Current_SF,I_AC_Power,I_AC_Power_SF,I_AC_VA,I_AC_VA_SF,I_AC_VAR,I_AC_VAR_SF,I_AC_PF,I_AC_PF_SF,I_DC_Current,I_DC_Current_SF,I_DC_Voltage,I_DC_Voltage_SF,I_DC_Power,I_DC_Power_SF,I_Temp_Sink,I_Temp_Sink_SF,I_AC_Power_Combined
        # List of starting registers for items listed above
#        register = 30769,30771,30957,30959,30775,30795,30803,30953,31247,30513,30517
        register = 40072,40076,40084,40085,40088,40089,40090,40091,40092,40093,40097,40098,40099,40100,40101,40102,40104,40107,40084
        # List of # of registers used for each item 
#        nReg = 1,1,1,1,1,1,1,1,1,1,1
        nReg = 1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,2
        # Data type for each item above
#        rType = float32,float32,float32,float32,float32,unint32,unint32,float32,float32,uint64,uint64
        rType = uint16,int16,int16,int16,int16,int16,int16,int16,int16,int16,uint16,int16,uint16,int16,int16,int16,int16,int16,uint32
        # nodeid used to match with node definition in nodes section below. Can be set to any integer value not previously used.
        nodeId = 27
        # Channel to publish data to should leave as ToEmonCMS
        pubchannels = ToEmonCMS,
        # time in seconds between checks, This is in addition to emonhub_interfacer.run() sleep time of .01
        # use this value to set the frequency of data retrieval from modbus client
        interval = 1
```

The power value is in the I_AC_Power register, and the scale factor is in register I_AC_Power_SF (both 16 bit integers).  So for example if

I_AC_Power = 6090
I_AC_Power_SF = -1
Power (W) = 6090 x 10^-1 = 609

But you might also see:

I_AC_Power = 60900
I_AC_Power_SF = -2
Power (W) = 60900 x 10^-2 = 609

alexandrecuer:
normally you can use scales instead of scale

here is an example

```
[[12]]
        nodename = fronius
        [[[rx]]]
            # list of names of items being retrieved
            # This example retrieves the Inverter status, AC power in watts being produced, AC Lifetime KWh produced,
            # KWh produced for current day,....
            names = Inverter_status,AC_power_watts,AC_LifetimekWh,DayWh,mppt1,mppt2,Vmppt1,Vmppt2,PhVphA,PhVphB,PhVphC
            datacodes = H,f,f,Q,H,H,H,H,f,f,f
            scales = 1,0.1,0.1,1,1,1,1,1,0.1,0.1,0.1
            units = V,W,kWh,Wh,W,W,V,V,V,V,V
```

which version of the modbus interfacer do you use ? I think you should upgrade because the last interfacer conf section is not so verbose....

cf https://github.com/openenergymonitor/emonhub/tree/master/conf/interfacer_examples/modbus

best

nev7n:
@alexandrecuer I am using scales later in the file when defining the node:
```
[[27]]
    nodename = inverter
    [[[rx]]]
        names = ac_total_current,ac_current_scale_factor,ac_power_value,ac_power_scale_factor,apparent_power,apparent_power_scale_factor,reactive_power,reactive_power_scale_factor,power_factor,power_factor_scale_factor,dc_current,dc_current_scale_factor,dc_voltage,dc_voltage_scale_factor,dc_power,dc_power_scale_factor,heat_sink_temperature,heat_sink_temperature_scale_factor,ac_power_combined
        datacodes = H,h,h,h,h,h,h,h,h,h,H,h,H,h,h,h,h,h,i
        scales = 1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1
        units = A,n,W,n,VA,n,VAR,n,%,n,A,n,V,n,W,n,C,n,n
```

Not sure how to check which version of modbus interfacer I have, I would assume fairly recent since I'm running on an emonpi I bought from openenergymonitor.org a couple of weeks ago and I've updated the software so I'm on emonSD version: emonSD-24Jul20

Comparing my config to the example https://github.com/openenergymonitor/emonhub/tree/master/conf/interfacer_examples/modbus it looks like I'm specifying rName, nReg and rType in addition to the fields in that example.  rName is probably redundant, but I think I need to specify nReg and rType since they vary by register.

alexandrecuer:
You dont have to specify nReg and rType...
everything is in the datacodes

cf 

https://github.com/openenergymonitor/emonhub/blob/2baf836fc4e75b2f158aa3ac722178dd864f5677/src/interfacers/EmonModbusTcpInterfacer.py#L81

and 

https://github.com/openenergymonitor/emonhub/blob/2baf836fc4e75b2f158aa3ac722178dd864f5677/src/interfacers/EmonModbusTcpInterfacer.py#L179

I confirm also Name is also redundant

nev7n:
@alexandrecuer thanks - that's useful to know

TrystanLea:
@nev7n were you able to use emonhub to achieve the goal of these input processes? I think naming the input process pow10_input might be better, as scale is used elsewhere to mean just multiply by a value.

nev7n:
@TrystanLea agree pow10_input is a better name, I've updated it. 

The pow10_input function was working most of the time for me, but occasionally due to the timing of the input feed being processed occasionally the scale from one point in time would be applied to the input from another, causing the result to be out of scale by a factor of 10 for that one tick.

To work around this I sent the 16 bit value integer and 16 bit scale integer packed together as a 32 bit value, then used the combined value_and_scale function to unpack the two values and apply the scale, this has been working great now:
```
    //---------------------------------------------------------------------------------------
    // Combines a value and scale input, multiplies the value by 10^scale
    //
    // These split value and scale inputs are output from the Modbus interface of solar edge
    // inverters.
    //---------------------------------------------------------------------------------------
    public function value_and_scale($arg, $time, $value) 
    {
        $scale = $value & 0xFFFF; // Right 16 bits are the scale
        if($scale >= 0x8000) { // Decode twos complement if negative
          $scale = -(($scale ^ 0xFFFF)+1);		
        }
        
        $val = $value >> 16; // Left 16 bits are the value

        // Apply scale factor to the value
        return $val * pow(10, $scale);
    }
```

TrystanLea:
Thanks @nev7n I will try and take a closer look at this soon, I just want to think through the option of including this in the emonhub interfacer to understand if that is an option

TrystanLea:
I've decided to put these processes in a new emoncms module called processext for now, making use of the modular implementation of input processing developed by chaveiro. This should hopefully get Richard up and running with this a bit faster given that I've not had a chance yet to review how the modbus interfacer could be modified. Here's the new module: https://github.com/emoncms/processext

TrystanLea:
Thanks for your efforts on this @nev7n sorry that this contribution was left without action on my part for so long!

PR #1653, opened 2021-01-28
Allow user themes to style multigraph.

cmcgerty:
My dark theme requires all page elements to pick up the black background and white text. Currently, multigraph does not process the user theme.


cmcgerty:
This is unnecessary. I had not realised that the new theme colour needed adding to both theme.php and embed.php.

PR #1654, opened 2021-02-06
Change how timestamp is removed from Inputs

borpin:
As noted https://github.com/emoncms/emoncms/commit/c8c88dd0bb2a7f2df8aab5977e32d89bc4966b25#r46773477

The `unset` was in the wrong place.

Moved as a standalone condition to remove from the `$inputs` if it exists.

NEEDS TESTING


borpin:
Tested locally and all seems OK.

borpin:
I have realised what is wrong, I `unset` the LC version of the array and return the original array. Needs more work.

PR #1657, opened 2021-02-16
Remove Unset of time parameter

borpin:
Remove `unset($jsondataLC['time']);` as it was causing unexpect results

https://community.openenergymonitor.org/t/anyone-importing-data-from-weewx-weather-station-into-emoncms-its-stopped-working/16916/2


TrystanLea:
thanks, strange that this was placed where it is, as jsondataLC is not created until lower down in the code

PR #1658, opened 2021-02-25
Emoncms feedlist refactor (Still in development)

TrystanLea:
Reverts feeds table to use a html table again, not sure why I went down the rabbit hole of using JavaScript to auto size html elements into columns!, when most of the functionality is possible out of the box.. 

This also uses handlebarsjs to build the feedlist rather than lots of html inside javascript code..

I've also split the modals out into separate files

Still work in progress.

Note this pull request includes the menu_v3 branch changes


TrystanLea:
I'm interested in any tips on how this could be improved.

Does separating out the modal files help with readability?

How could they be improved further?

TrystanLea:
I've hit a crucial problem with this implementation in that table tbody elements do not support slideUp and slideDown animations.
It's a really nice visual element to the feed and input lists to have this feature and so I think Im going to have to revert my proposed change back to tables and keep with the custom javascript implementation

chaveiro:
Thinking about the new table for input and feeds that visualy is okish, the code behind got unmanegeable.. 
Your idea to revert back to the old table logic and add the new visual features makes some sense here.

PR #1659, opened 2021-02-27
Bugfixing 10.3

chaveiro:
Fix mysql DMY problem
- Improvement of zoom graph viz
- Improvement of Input list on small devices and with big descriptions.


chaveiro:
Please merge, more to come later.

TrystanLea:
Hello @chaveiro thanks for this, would you mind breaking this down into each individual feature, I find the mix of different things really difficult to evaluate. 

I think this would work well:

- Pull request 1: Mysql engine bug fix 
- Pull request 2: input view changes (include screenshots of what the changes achieve and why they are needed).
- Pull request 3: vis api.js change (i think its just a change of order?)
- Pull request 4: multigraph changes (im not sure whats going on in those changes, a bit of a description on that would be great)
- Pull request 5: zoom interface changes (screenshots and explanation again of what's changed would be really helpful)

I feel like my past practice of merging pull request blind has made the emoncms code base something that I no longer understand and makes it really hard to maintain and continue to develop. Some of it is also my own doing of course but really keen to try to get back to something more maintainable.

TrystanLea:
Could you also give me some visibility as to your proposed up and coming changes? 

It would be good to check that we are not planning parallel efforts and to find that we get to a point were we are both working on the same part and need to merge two different streams of work together.

One example is the feed list refactor (https://github.com/emoncms/emoncms/pull/1658) which I plan to replicate on the inputs list which would remove your changes above alongside the rest of the js I introduced there to create the responsive view.

chaveiro:
Hi, this are just minor fixes to 'in my face' bugs.
No real plan for future features for now.
I see the code got a really mess since the last 4 years... Ex the inputs and feeds views are totally different in implementation with redundant code.
For the mysql engine i was comparing to my 4 years version and just fix a simple issue that were returning a null on the DMY result and  also refactor to the template that exists so i can compare the files side by side. 
Zoom fix was to add kwh/m /y /d and price where relevant. also fix some NaN issues that were appearing.
Please accept this commit and i will take be more careful breaking context on the next ones, all tested.

TrystanLea:
Ok I've merged this one, I notice there is a gap opened on the feeds page, just under the controls:

image

do you see the same?

chaveiro:
Its caused by https://github.com/emoncms/emoncms/blob/815a630d7227cf7a71ba31b42e2c1de18d581545/Modules/feed/Views/feedlist_view.php#L80 

Why is there, can be removed ?

PR #1667, opened 2021-03-02
virtual feed null value fix

TrystanLea:
Fix for bug were null values in either source feeds being used in a virtual feed process were converted to zeros in the output as seen in the graph.

@chaveiro

### DIFF
--- a/Modules/process/process_processlist.php
+++ b/Modules/process/process_processlist.php
@@ -1638,6 +1638,8 @@ class Process_ProcessList
     public function add_source_feed($feedid, $time, $value, $options)
     {
         $last = $this->source_feed_data_time($feedid, $time, $value, $options);
+        
+        if ($value==null || $last==null) return null;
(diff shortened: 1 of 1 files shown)


TrystanLea:
Before screenshot:

image

After screenshot:

image

chaveiro:
Not sure if this change fixed this already:  https://github.com/emoncms/emoncms/pull/1659/files#diff-0fb56f6565d5f7802636d882b2ba010ec9b4b21fd7c4d7b000d61091a0f93959R550

TrystanLea:
Sorry @chaveiro I cant see were in your changes a fix that would have effected this was made? Testing here before merging this pull request showed that the issue was still present. I've merged this pull request to fix.

If this bug was not present in v9 (it looks like the last commits to these processes where yours) perhaps it is was introduced by a change in the way php handles null values in php 7+? and therefore neither of our faults?

PR #1669, opened 2021-03-02
Fix inputlist resize behaviour

TrystanLea:
Fixes bug when resizing input lists. 

**What the bug looked like:**
image

**Description:**
When resizing to small screen the processlist element retained it's original height even though the element is meant to be hidden by width = 0px. Adding another property to set the element height (col_h) and switching between auto and 0 height depending on whether it should be hidden or not fixes most of the issue.

There was an associated second issue to do with conflicting window resize events registered in Lib/emoncms.js and the input_list.js. This pull request switches to using the trigger created by the emoncms.js event rather than having a duplicate handler.

**Result**

image

### DIFF
--- a/Lib/emoncms.js
+++ b/Lib/emoncms.js
@@ -70,7 +70,7 @@ $(function(){
         clearTimeout(resizeTimeout)
         resizeTimeout = setTimeout(function() {
             $.event.trigger("window.resized");
-        }, 200);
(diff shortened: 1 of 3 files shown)


chaveiro:
i think that the bug got fixed on https://github.com/emoncms/emoncms/pull/1666 but please try to merge the 2

PR #1677, opened 2021-03-22
remove support for histogram feed engine and associated visualisations

TrystanLea:
This pull request removes support for the histogram feed engine which like phpfiwa has not really been supported for a long time (feed creation with this engine has already been disabled for probably 5 years +). 

This cleans up code that is no longer maintained making it easier to maintain and focus on emoncms core features.


chaveiro:
Hi, is there a newer alternative to this histogram functionality?

TrystanLea:
Yes, at least part of it is built into the graph module. https://guide.openenergymonitor.org/emoncms/histograms/
Do you use the histogram engine and how do you use it?

Ultimately it could be possible to added an API end point to all standard time series feed engines to calculate this kind of output, in the same way that we calculate averages etc on the fly.

alexandrecuer:
I agree : for me the graph module does clearly the job and so I think it is a viable alternative.

TrystanLea:
Thanks @alexandrecuer

chaveiro:
Nice to know that histogram feature of graph module. It kind of works, a little slow for many years of data but useful. Is there a way to put juat the graph on a dashboard?

TrystanLea:
Thanks @chaveiro no it's not possible to put this information in the dashboard yet unfortunately but it would be something to consider for future development.  At some point I would like to revisit the dashboard side of things and bring some of the app functionality across into dashboards

PR #1679, opened 2021-04-04
changed max power threshold for accumulators form 25 kW to 60 kW

mr15:
25 kW seems to be an arbitrary limit for omitting bogus pulses. Residential gas appliances can go up to say 50 kW, therefore I propose to increase the threshold to 60 kW.
I hope I found all places where these 25 kW are mentioned, please double check.


mr15:
Actually I'd like to have this change also on emoncms.org (which I am using for years now). Do you want me to do another pull request there (https://github.com/emoncms/emoncmsorg/blob/751ebfa2851ca6b8f955252fff5320a2d5f78438/emoncms/Modules/input/process_model.php) or how are changes propagated to the emoncms.org instance?
CC: @TrystanLea

TrystanLea:
Thanks @mr15 this seems reasonable, I will merge to emoncms.org separately, the process is a bit slower

PR #1681, opened 2021-04-17
Add emoncms-cli

inverse:
While working on the [Home Assistant extension](https://github.com/inverse/hassio-addon-emoncms) for emcoms. An issue was raised that required database migrations to be manually executed.

See: https://github.com/inverse/hassio-addon-emoncms/issues/21

I was thinking would it be good to have a basic CLI tool that could be used to perform such migrations? So that I could hook into this on the service starting up. e.g. when upgrades happen or just general life-cycle changes of the container.

What are your thoughts on providing such a thing?

Some of the common init logic could be extracted into a shared file.


TrystanLea:
Hello @inverse thanks for working on the home assistant extension for emoncms! I keep meaning to spend some time getting up to speed with home assistant! 

In principle yes happy to merge this, how do you see this CLI expanding further beyond updating the database? 
Would it be worth calling this command: dbupdate or mysqlupdate to be more specific?

TrystanLea:
Is there a way that you could perform these actions via the emoncms http api? is it an authentication issue that gets in the way?

inverse:
You really should give it a shot! If you're comfortable with docker I'd highly suggest just spinning it up with that and playing around with it.

I didn't envision anything yet, but wanted to abstract it in a way that it could be extended to add additional commands in a backwards compatible way.

But it could include things like: flushing caches, dumping data, triggering cron related things

Authentication would get in the way of that as the extension wouldn't know any user information on startup.

I will rename the command and add some documentation.

TrystanLea:
Ok thanks, so is this specifically for those operations on startup, or do you see it being extended with things like accessing feed data etc?

inverse:
It could be extended for all kinds of CLI related tasks as the community see's fit.

If so would you suggest moving this `dbupdate` command under another sub-command to allow better organisation? 

```bash
./emoncms-cli admin dbupdate
```

or

```bash
./emoncms-cli maintenance dbupdate
```

Open to suggestions :)

TrystanLea:
Yes I think the nested arg is better, admin:dbupdate
Lets go with that to start with, great!

TrystanLea:
Sorry one last change, could we move this to the scripts directory?

inverse:
Done!

TrystanLea:
Sorry for the delay!!

inverse:
No need to apologise - we're all working on OSS in our free time ;)

PR #1686, opened 2021-05-08
Add a delayed / smoothed kWh to power input processor.

rmn30:
This is useful to reduce quantisation in instantaneous power measurement derived from pulse counting.

The emonPi reports pulse counts every 5 seconds, which leads to a problem if trying to derive instantaneous power using the `kWh to power` input processor: every 1 Wh pulse recorded in a 5 second interval equates to 720W of power so the power level appears to jump between a small number of bands. This doesn't produce nice graphs and can lead to artefacts like apparent negative household use when the power level reads lower than it should (there's nowhere to put a CT clip on my PV generation so I am using pulse counting to measure PV and a CT on the grid meter tail then deriving household use as PV generation + grid import / export).

The best solution would be for the emonPi firmware to time the interval between the two most recent pulses and report this as well as the pulse count. It looks like this would be pretty easy to do in the firmware but I don't have a programmer at the moment so I made this emoncms input processor to estimate power over a longer period which improves the resolution. It works by storing the last N inputs in a redis list (hardcoded to 10 at the moment because I don't think it's possible to specify more than one argument for a processor?) then calculating the power level based on the difference between current and oldest value. The effect is like a low pass filter: removing the high frequency oscillations caused by aliasing / quantisation but delaying and smearing the output slightly.

### DIFF
--- a/Modules/process/process_processlist.php
+++ b/Modules/process/process_processlist.php
@@ -760,6 +760,20 @@ class Process_ProcessList
               "description"=>_("<p>Return the reciprical of the specified feed. Returns NULL for zero values.</p>")
            ),
            array(
+              "id_num"=>60,
+              "name"=>_("kWh to Power (smoothed)"),
+              "short"=>"kwhpwr_smooth",
+              "argtype"=>ProcessArg::FEEDID,
+              "function"=>"kwh_to_power_smoothed",
+              "datafields"=>1,
+              "datatype"=>DataType::REALTIME,
+              "unit"=>"W",
+              "group"=>_("Power & Energy"),
+              "engines"=>array(Engine::PHPFINA,Engine::PHPTIMESERIES),
+              "requireredis"=>true,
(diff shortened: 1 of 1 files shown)


TrystanLea:
Thanks @rmn30, Keeping this open for now, trying not to add too many new processes in the core at the moment but I do like the idea of this one. I agree we should be timing the pulses on the pi/emontx ideally. Thanks for making the pull request.

rmn30:
OK, I've been using it for a while and it's not perfect but better. Unfortunately I see big spikes when rebooting and haven't spent the time to diagnose and fix yet. I will have a go at modifying the emonpi firmware. I didn't realise it is possible to program it directly from the pi.

TrystanLea:
Hello Robert, I've included your input process in a new emoncms module intended to contain additional input processors that can be optionally installed to extend emoncms input processing, it's available here: https://github.com/emoncms/processext
Thankyou for contributing this!

PR #1691, opened 2021-05-19
Admin display of masked services

joemcool:
The Admin page did not handle a situation where a service was loaded in systemd but purposefully masked so it wouldn't run (for example, when running the MQTT broker on another machine).  Masked services were reported as errors since they weren't running, whether they were supposed to or not.  Such services are now displayed in a new 'Masked' state with a grey status badge (instead of the original red/green badges).

See further discussion of this change at https://community.openenergymonitor.org/t/admin-page-services-status-doesnt-handle-disabled-services/15793/2


TrystanLea:
Thanks @joemcool

PR #1692, opened 2021-05-21
Work around lack of GLOB_BRACE properly 

inverse:
My attempt to fix this in #1685 didn't work and only hid the problem. This solution removes the need for the glob brace expansion by calling glob multiple times.

Tested on the addon on a branch and it works.

p.s. with pull requests when you use squash instead of merge the git history of the repo looks cleaner as you remove all the noise within the PR itself :)

### DIFF
--- a/locale.php
+++ b/locale.php
@@ -12,30 +12,28 @@
 // no direct access
 defined('EMONCMS_EXEC') or die('Restricted access');
 
-
-// Work around for Alpine Linux
-if (!defined('GLOB_BRACE')) define('GLOB_BRACE', 0);
-
-
 // Return all locale directory from all modules.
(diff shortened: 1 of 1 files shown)


locale.php
```diff
@@ -12,30 +12,28 @@
...
-
-// Work around for Alpine Linux
-if (!defined('GLOB_BRACE')) define('GLOB_BRACE', 0);
-
-
 // Return all locale directory from all modules.
 // If one module has a language it will be detected
 function directoryLocaleScan($dir)
 {
+    $directoryList = array();
```
inverse, line 19: Ensures that a an array is returned and doesn't blow up later down the line.

TrystanLea:
Thanks @inverse !

PR #1697, opened 2021-06-15
Psychrographs in the vis module

alexandrecuer:
@TrystanLea : I would like these commits to merge to emoncms, if possible. Hope you will be interested :-)

This permits to create psychrometric diagrams to appreciate summer confort in a house.

commits are only located in the vis Module

To create a pshychrometric diagram, you need to build a multigraph with a temperature feed first and a relative humidity feed after. 

image

The multiphrap above is for one area one  but to compare multiple areas of your home, you may stack as many "TRH" pairs as you want in the multigraph and the resulting psychrometric diagram should be something like that : 

image

the blue trapeze is the confort area. So if you are here, no problem, everything is fine.

 best

Alex


TrystanLea:
This looks great @alexandrecuer! Thankyou for contributing this!

alexandrecuer:
@TrystanLea thanks for merging :-)
I will produce a documentation with how to shape the multigraph in order to use it in the pshychrograph...

TrystanLea:
Thanks @alexandrecuer yes that would be great, how does it know to pick up a temperature feed vs a humidity feed?

alexandrecuer:
You have to stack the feeds on a specific order : temperature first then humidity, fix the color, then another temperature and the correspondant humidity, and so on...

Issue #1701, opened 2021-07-06
MySQL no longer needs length of int variables to be specified


brucemiranda:
The schema.php files have table declaration arrays. These arrays have lengths for the int datatypes. However mySQL now gives a warning when trying to apply a length to an int datatype. However because of the way the comparison of the table description and the schema arrays works, emoncms marks these as Updates to be applied. But these can never be applied and so you land up in a loop. 
This is true atleast on mySQL Version 8.0.25-0ubuntu0.20.04.1

Issue and Solution discussed here. https://community.openenergymonitor.org/t/update-database-changes-not-being-applied/18049

Either need to update the schema definitions to remove the int lengths or create a more intelligent table description comparison

PR #1702, opened 2021-07-06
Remove all the int variable declaration lengths which are no longer required

brucemiranda:


brucemiranda:
Fix for https://github.com/emoncms/emoncms/issues/1701

TrystanLea:
Thanks @brucemiranda I will get to this one soon

PR #1709, opened 2021-07-13
Update theme.php

thib66:
Add localizations to the theme
before : 
image
after : 
image


thib66:
Also update French translations / localisations for the other modules

PR #1713, opened 2021-07-18
Add localization 18 07 2021

thib66:
Before : 
image

After : 
image


TrystanLea:
Thanks again @thib66 I appreciate your continued work on the translations like this!

PR #1716, opened 2021-07-22
secure session cookie, use samesite=strict

TrystanLea:
Protect against Cross-Site Request Forgery by restricting use of session cookie to emoncms domain

### DIFF
--- a/Modules/user/rememberme_model.php
+++ b/Modules/user/rememberme_model.php
@@ -40,8 +40,12 @@ class Rememberme {
     // ---------------------------------------------------------------------------------------------------------
     public function __construct($mysqli)
     {
-            $this->mysqli = $mysqli;
-            $this->log = new EmonLogger(__FILE__);
+        $this->mysqli = $mysqli;
+        $this->log = new EmonLogger(__FILE__);
+        
+        if (is_https()) {
+            $this->secure = true;
+        }
     }
 
     // ---------------------------------------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


TrystanLea:
@inverse @chaveiro would any of you have any view on this by any chance? Seems to work fine here, am I restricting a useful auth pathway by setting to strict?

Modules/user/rememberme_model.php
```diff
@@ -40,16 +40,29 @@ class Rememberme {
...
-        setcookie($this->cookieName,$content,$expire,$this->path,$this->domain,$this->secure,$this->httpOnly);
+        // setcookie($this->cookieName,$content,$expire,$this->path,$this->domain,$this->secure,$this->httpOnly);
+        // May be limited to PHP7.3
+        setcookie($this->cookieName,$content, [
+            'expires' => $expire,
+            'path' => $this->path,
+            'domain' => $this->domain,
+            'secure' => $this->secure,
+            'httponly' => $this->httpOnly,
+            'samesite' => 'Strict'
```
inverse, line 64: Could also be defined like this https://stackoverflow.com/a/46971326/50913

Which could be a way to make it work for lower versions

core.php
```diff
@@ -15,20 +15,26 @@
 // no direct access
 defined('EMONCMS_EXEC') or die('Restricted access');
 
-function get_application_path()
-{
-    // Default to http protocol
-    $proto = "http";
-
+function is_https() {
```
inverse, line 18: Nice refactoring here 👍🏼

inverse (approved):
Looks good for me and tested in the emon docker-compose setup.

Really think considerations need to be made about the minimum supported PHP version. E.g. inline with EOL etc.

inverse:
Regarding your question. From my understanding and [refreshing](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie)

> The cookie is not sent on cross-site requests, such as calls to load images or frames, but is sent when a user is navigating to the origin site from an external site (e.g. if following a link).
This is the default behavior if the SameSite attribute is not specified.


Would this imply it'll be blocked if someone wanted to embed emoncms in an iframe or such?

TrystanLea:
Thanks @inverse, I think emoncms iframes embedded in 3rd party sites will work fine because they are either public or authenticated with an apikey. I will do some testing around that just in case.

Your right we should target a minimum PHP version now. 
Like you say we should move along with the official security timelines https://www.php.net/supported-versions.php

Reviewing emonSD image history, emonSD-17Oct19 was the first image of the new EmonScripts installation system (and can still be updated), it had version 7.3.9-1~deb10u1 (Zend Version 3.3.9), all emonSD images before that should not update via the emoncms updater, so it seems that setting 7.3 as our minimum version should be fine.

PR #1724, opened 2021-08-21
Update process_processlist.php

CidiRome:
Add Wh Accumulator that does not have a limit of 60Kw.

### DIFF
--- a/Modules/process/process_processlist.php
+++ b/Modules/process/process_processlist.php
@@ -760,6 +760,20 @@ class Process_ProcessList
               "description"=>_("<p>Return the reciprical of the specified feed. Returns NULL for zero values.</p>")
            ),
            array(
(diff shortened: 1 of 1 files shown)


TrystanLea:
Keeping this open for now as it's a nice addition but also keen not to add too many new processes at this point. 
Id like to review options to include process parameters separately somehow.

chaveiro:
Fyi processes are modular, see scheduler module as example. 
For a module with processes just create a new module file at Module/modulename/modulename_process.php

TrystanLea:
Good idea @chaveiro I've created a new module called processext and included this input process in that module. Thankyou @CidiRome for the contribution!

TrystanLea:
Here's the link to the new module: https://github.com/emoncms/processext

PR #1730, opened 2021-09-17
Fix admin: log and component

chaveiro:
- Avoid info log spam
- Fix admin log bootstrap buttons
- Components update support reading available components from github stable repo
- Components update need redis check
- Components update remove hardcoded location
- Components update reporting path


chaveiro:
Hi,
Location is used for what? No problem keeping it, just that  for better support it shouldn't be an absolute path.
I can move back the log level to the bottom right side of the the log text box.   The funny part was that I started to implement a text message with the log level just to notice that it was there already. Never noticed it at that 'hidden' side of the screen. Also the syntax for bootstrap code was wrong. And the drop down with the loglevels (if you switch the if reverse) was showing out of placement at the left side of the page. If this is an aborted development, maybe just remove the if all ltogether.

TrystanLea:
Thanks @chaveiro 

Location is the target install location as defined in module.json. It will be used to install the modules in the right places.

Your right the log level selector is currently aborted development, I have removed it for now and added a title property to suggest changing in the settings file if needed.

chaveiro:
Thanks, i've in the meantime Push another commit.

PR #1734, opened 2021-10-03
Menu revamp

chaveiro:
- Fix left menu system quirks
- Fix left scroll bar
- Add auto hide on first page load if menu is overlaying content
- Fix content centering / fitting on resize / screen rotation
- Add shadows to menus
- Improve readability contrast on menu mouse over
- Add auto hide top menu on scroll if no lateral menus are open
- Fix menu-l2-controls abnormal movement

### DIFF
--- a/Lib/menu/menu.css
+++ b/Lib/menu/menu.css
@@ -193,6 +193,7 @@ text-decoration:none;
   z-index:999;
   overflow-y: auto;
   overflow-x: hidden;
+  white-space: nowrap;
   scrollbar-width: none; /* can also be: thin */
   -ms-overflow-style: none;
   box-shadow: 0 2px 4px 1px rgb(0 0 0 / 25%);
--- a/Lib/menu/menu.css
+++ b/Lib/menu/menu.css
@@ -96,6 +96,7 @@
   position:fixed;
   top:46px;
   z-index:1000;
+  box-shadow: 0 2px 4px 1px rgb(0 0 0 / 25%);
 }
 
 .menu-l2-inner {
@@ -155,7 +156,8 @@ text-decoration:none;
 
 #menu-l2-controls {
   background-color:var(--bg-l2);
-  position:absolute;
+  position:fixed;
+  bottom: 0;
   text-decoration: none;
   color: var(--l2-text);
   display: block;
@@ -164,7 +166,7 @@ text-decoration:none;
 }
 
 .ctrl-exp {
-  padding: 15px 110px 15px 110px;
+  padding: 15px 112px 15px 112px;
   /*transition: all 0.1s ease-out;*/
 }
 
@@ -191,6 +193,7 @@ text-decoration:none;
   overflow-x: hidden;
   scrollbar-width: none; /* can also be: thin */
   -ms-overflow-style: none;
+  box-shadow: 0 2px 4px 1px rgb(0 0 0 / 25%);
 }
 
 .menu-l3 ul {
--- a/Lib/menu/menu.css
+++ b/Lib/menu/menu.css
@@ -90,18 +90,18 @@
 
 .menu-l2 {
   width:0px;
+  height: 100%;
   background-color:var(--bg-l2);
   display:inline-block;
   position:fixed;
   top:46px;
-  height: 100%;
   z-index:1000;
   transition: all 0.3s ease-out;
   box-shadow: 0 2px 4px 1px rgb(0 0 0 / 25%);
 }
 
 .menu-l2-inner {
-  height: calc(100vh - 96px);
+  height: calc(100% - 96px);
   overflow-y: auto;
   overflow-x: hidden;
   scrollbar-width: none; /* can also be: thin */
--- a/Lib/menu/menu.css
+++ b/Lib/menu/menu.css
@@ -90,18 +90,17 @@
 
 .menu-l2 {
   width:0px;
-  height: 100%;
   background-color:var(--bg-l2);
   display:inline-block;
   position:fixed;
   top:46px;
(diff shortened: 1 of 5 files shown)


TrystanLea:
Thanks for creating this in a separate feature branch @chaveiro I will review soon.

chaveiro:
The menu appearing is due to https://github.com/emoncms/graph/pull/137/files#diff-024df4e202173453d097d453bc377cfa4ecf8e22ac1009fd8fca3a7fa3e06d15 .
You have to segregate responsibilities and the way menu was done had lot of work around's.
I understand you had lot of work on it but lots of corner cases were not taken care, i've covered them all.
Please trust my judgment on this and try it for a few days, you should not get any unexpected behavior, but if you do, just point out and i will fix.

TrystanLea:
Hey Chaveiro

The graph page should not appear on the login page but it should be able to appear in the public context, it allows public access to shared feeds e.g https://emoncms.org/graph?userid=8862 

Then when I login, the sidebar incorrectly overlays the feeds list:

image

and when I navigate between the L1 menu's the animation sort of pops in and out in a strange way that doesnt work as well as the original implementation:

animated

I prefer the original behaviour:

original

TrystanLea:
Here's the issue on login, notice how the menu covers the feed list in my expanded view, it should open with the feedlist margin moved to the right so that you see both the menu and the feed list: 
login_issue

```
From a user experience perspective, how many times do you switch menus like that ? :)
When you switch menu, it collapse the previous and expands the new. Makes sense, don't you think?
```
Ha, no I dont agree! :) I really prefer the way I had it already! sorry

chaveiro:
@TrystanLea  have reverted back to the original style animations on the last push that also simplified much more the code, less differences from where we started. I like it better now :)

Auto hide only does is the follow:
1 -If there is an overlay of l2-exp (screen < 1150px)  -> auto will make l2-min
2 -If there is an overlay of l3 (screen < 1150px)   -> auto will close l3
(similar to menu.mode = 'manual' that you have for resize)

Is enabled when:
- page is refreshed, resized or l2 is manually expanded 

Is disabled when:
- user changes l2 context from l1 menu.
- after hiding one time (example: if the user expands l3 after it closes it will not close again)

TrystanLea:
Ok, can you help me try to understand what you were seeing, what was the original problem, I'm not yet following. 
What is the screen width that l2-min at page load is intrusive at? Can you do a gif? 

I cant see the top menu hiding with you branch, how does that work?

chaveiro:
This is a screenshot from my mobile,  the top menu and l2 min takes space by default.

To hide top menu, hide l2 and scroll down.

Screenshot_20211014-200528_Chrome

TrystanLea:
Thanks @chaveiro!

TrystanLea:
Hey @chaveiro I've just got a lot on at the moment, I have times when I can give emoncms development more time and other times where I may have 30 mins in a day. Believe me I have a lot of emoncms development work myself that is still in open branches waiting for the point at which I can continue that work and bring it to release. This last month I've had less time for emoncms development than usual. I can do fast merges on pull requests that are either fully self contained e.g your mysql to phptimeseries script (thanks for that!) or pull requests that are maybe ~10-20 lines of code changed at once but not 100's of lines in parts of emoncms that I particularly want to be careful about, the menu system being one of those.

I tested the feature that hides the top menu when l2 is hidden quickly this morning and I like it.
I will continue with this when I get time in the coming week, please don't rush me with this.

chaveiro:
> please don't rush me with this

Sorry I understand,  my time is also limited, sometimes we just need to prioritise or delegate to get some releave. I also saw some old pulls back to one year still open, some more challenging than others, maybe try to plan for some cleaning later.

> can you explain what you mean here, does this relate to existing changes in this pull request?

I may have been more clear, i meant that for example show-l2 and hide-l2 will be the single responsible for setting visible-l2 variable. No other function should set it as it relates to l2 only.
Draw methods are used for building the obj var.
Show and hide for display the divs and setting page content margins.
And resize for all the menu visible logic. 
Also other modules should not set variables on the menu, just call show/hide and not set page margins, leave that single responsibility to the menu class.

> looks like that one might be a cache issue

happens a lot.

>  secondary top menu on the feeds page that sometimes has an offset

That is one of the reasons I want to close this push and move on. That is on the feeds page responsibility.

chaveiro:
Have tried Meld as you suggested and didn't liked it for the reason it makes too much coloring on top of the comparing files ,wrongly highlighted some lines that are not changes, just spaced differently and different scroll positions for each file.
Maybe I'm not used to that style as I've been using Beyond Compare (payed) for many years and it works really great to compare changes while ignoring unmeaningful parts, like indentation changes that does not break code.
See same comparison on the different programs:

image

chaveiro:
I dont see that, i think you forgot to submit the review so it's only visible to you?

TrystanLea:
Thanks @chaveiro

chaveiro:
Fireworks! :)
Thank you @TrystanLea for trusting my work.

TrystanLea:
Thanks @chaveiro 

I did make a couple of changes to get a particular behaviour I was after with some of the animations, I've carefully committed these changes in discrete commits that only contain each particular change at a time so that they are hopefully a little clearer to understand https://github.com/emoncms/emoncms/commits/master

There is still one animation I'm not quite happy about, when we have say the admin l3 menu open and navigate between sub menu pages, the l2 menu animates each time which you don't want, but you do need that animation the first time when you open an admin page.. so it really needs a way of remembering the previous state of the menu to stop the animation for subsequent page changes.. I've merged it for now with the view to fixing this later.

PR #1738, opened 2021-10-04
php7.0 compatibility mode for rememberme

mabi:
Rememberme setcookie is using PHP 7.3 signature, check PHP version and fallback to older signature if needed.


TrystanLea:
Thanks @mabi

PR #1746, opened 2021-11-01
Virtual feeds: modification to use in-built feed engine averaging and daily, weekly, monthly timezone aligned requests

TrystanLea:
Virtual feeds: modification to use in-built feed engine averaging and daily, weekly, monthly timezone aligned requests

Implementation is also faster. E.g: loading one month of data at an hourly interval took ~215ms with the original implementation and is now about 12ms.

Results for a virtual feed (single source feed only) and the original feed are consistent (graph shows both plots overlayed):

image

With averaging on both and hourly interval:

image

Daily timezone aligned interval with averaging:

image


TrystanLea:
Here's another good example of it working well dividing heat output by electrical input to a heat pump, averaging the data at 60s and returning a result across the local network in 64ms down from about 460ms.

image

PR #1748, opened 2021-11-03
new virtual feed implementation, makes use of in-built feed engine av…

TrystanLea:
…eraging and DMY timezone aligned requests


TrystanLea:
Here's another good example of it working well dividing heat output by electrical input to a heat pump, averaging the data at 60s and returning a result across the local network in 64ms down from about 460ms.

image

PR #1732, opened 2021-09-23
Revamp Admin Module

chaveiro:
Fixs or add the follow:
-  Full components manager
-- Add component uninstall
-- Add component install
-- Add components available list
-- Notify if updates available
- added disks read load statistic
- improved disks write load statistic to use (if installed) iostats, remove hardcoded partitions to support non raspberypi devices
- added php native support for components update with git 
- added cpu info
- added machine brand model and bios description
- show all services states and added ssh
- Added support for running services from PHP when redis not available.
- Move service-runner trigger run responsibility to the model class runService() function.
- User list view return link
- Use model class the right way: remove static usage and global 'nightmare'
- Support disk write load stats without redis
- Preparation for future script execution / components update without redis
- Remove absolute paths from script files (use config base path instead)
- Uniform code across all modules
- Minor UI fixs
- Fix state machine of update logs when needed so that errors can be seen 
- Deal with lack of redis
- Make use of json result codes for all errors across admin views (components, log, firmware and update)
- Move views to ./Views/ folder
- Check symbolic link exists when deleting a module
- Added current user info to main admin page

PR #1765, opened 2021-11-30
Chaveiro component manager

TrystanLea:
See original pull request: https://github.com/emoncms/emoncms/pull/1732
This pull request fixes the conflicts that arose from my original partial merge.
Review still in progress

### DIFF
--- a/Modules/admin/Views/admin_main_view.php
+++ b/Modules/admin/Views/admin_main_view.php
@@ -213,6 +213,7 @@ listItem;
         <h4 class="text-info text-uppercase border-top pt-2 mt-0 px-1"><?php echo _('PHP'); ?></h4>
         <dl class="row">
         <?php echo row(_('Version'), $system['php'] . ' (' . "Zend Version" . ' ' . $system['zend'] . ')'); ?>
+        <?php echo row(_('Run user'), _('User') . ": " . $system['run_user'] . " " . _('Group') . ": " .  $system['run_group'] . " " . _('Script Owner') . ": " . $system['script_owner'] ); ?>
         <?php echo row(_('Modules'), "<ul id=\"php-modules\"><li>".str_replace("v".$system['php'],"", implode('</li><li>', $php_modules)).'</li></ul>', '', 'overflow-hidden'); ?>
         </dl>
 
@@ -508,6 +509,7 @@ $(window).resize(function() {
 function getBufferSize() {
   $.ajax({ url: path+"feed/buffersize.json", async: true, dataType: "json", success: function(result)
     {
+      if (result.reauth == true) { window.location.reload(true); }
       $("#bufferused").html( result + " feed points pending write");
     }
   });
@@ -518,6 +520,7 @@ function getBufferSize() {
 $("#redisflush").click(function() {
   $.ajax({ url: path+"admin/redisflush", async: true, dataType: "json", success: function(result)
     {
+      if (result.reauth == true) { window.location.reload(true); }
       $("#redisused").html(result.dbsize+" keys ("+result.used+")");
     }
   });
@@ -568,6 +571,7 @@ $(".service-action").click(function() {
     console.log(action+" "+name)
     
     $.ajax({ url: path+"admin/service/"+action+"?name="+name, async: true, dataType: "json", success: function(result) {
+        if (result.reauth == true) { window.location.reload(true); }
         setTimeout(function() {
             location.reload();
         },1000);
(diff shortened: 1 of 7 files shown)


TrystanLea:
Hello @chaveiro 

I'm holding off on merging this for now as I would really prefer to only maintain one way of installing, updating, uninstalling modules rather than having both the script calls and the php implementation to maintain. Your php approach is a good suggestion and I will definitely consider it carefully when I bring my focus back to this side of emoncms development. I have merged everything else and given you a pull request here to fix the conflicts that were created https://github.com/chaveiro/emoncms/pull/3. 

A couple of questions I have that I'm still thinking about are:

- Im not sure if giving the www-data user write permissions to the module directories is a good idea? Is that normal for other projects e.g wordpress, drupal? What are the implications of this that we need to think about?

- Some modules require running an install.sh script, the php implementation so far suggests that the user runs this manually if required. This is not ideal as it would be good if the full install completes as per install script requirements.

Id say that my general feeling on this, even though I do quite like your php implementation, is that the EmonScripts install scripts approach is there already to give users a consistent way of installing the full emoncms stack, reusing that code from within emoncms for installing and updating components is an advantage as it avoids code duplication and maintaining multiple versions of the same thing. To add this additional code to support environments without the ability to run the EmonScripts is Im not sure worth it. Likewise with things like the option to use emoncms without redis. There are cheap VPS services out there now that can provide all the requirements that emoncms needs, Emoncms is free and development time is limited , I would really prefer to target a specific environment and have it work well in that environment, rather than try to maintain in perpetuity support for a lot of different environments.

chaveiro:
Hi Trystan,

> Im not sure if giving the www-data user write 
permissions to the module directories

The approuch is to have a 'emoncms' owning the files and  running httpd so not to be www-data that may be running other websites that in case of a security compromise all is restricted to that user only.

> Some modules require running an install.sh script

Some may require and some don't, usualy the ones that require are not related to the Web interface but are the core backend services, not possible to install on shared servers anyway. Even for those, php can run sh scripts with the right permissions. Maybe just need to warn the user when it can't be done due to permitions restrictions.
The end goal is to have a self service update area that works for most schenarios on this product.

>Emoncms is free and development time is limited , I would really prefer to target a specific environment and have it work well in that environment

Wouldn't be better to think on an unique solution that targets the many environments in decrement of a solution that targets only one?


If we look at current trends, shared hosting is not popular anymore (but still the cheapest) future is aligned with automation and  having docker images as virtualization. 
Have a core docker emoncms image, mysql and redis as other images and plugins installed on a shared volume via the core imagem may make more sense as vps support and even raspberry deployment, be decouplin the host OS from the application.
See https://hub.docker.com/_/wordpress for inspiration.
This would work for raspberry Pi also.
So you update core docker image and then install / update plugins via the admin interface.

chaveiro:
As a side note, do you have statistics of how many users are using raspberrypi, shared hosting, emoncms.org or vps currently?
That may help decide on the support to prioritise.

TrystanLea:
Thanks @chaveiro Yes I think docker would be the other environment that would be good to target. My understanding from chatting to others is that the application should not really be able to modify the components installed in a docker image, it should be modified from the docker file itself.. it would mean of course restricting all component install, update, uninstall actions on the docker image.. Im not convinced that we have to adhere to this principle, if it is a principle at all? Are you familiar with how wordpress handles plugins within it's docker container? I guess it must allow modification of the installation?

TrystanLea:
I don't unfortunately have stats as to how many users use emoncms in different environments,  all I can say is that we havent actively tried to support shared, wamp, mamp beyond those initial install guide written maybe 6-7 years ago, and when users ask about support for issues on custom installs we generally suggest installing using EmonScripts as there's always a so many different variables that can affect an installation otherwise.

My guess would be:

1. Raspberrypi environment (emonSD image built using EmonScripts)
2. Emoncms.org 
3. Custom install on VPS 
4. Install using EmonScripts on RaspberryPi or VPS 
5. Docker
5. Shared hosting
6. WAMP & MAMP

TrystanLea:
`Some may require and some don't, usualy the ones that require are not related to the Web interface but are the core backend services, not possible to install on shared servers anyway.`

Handling these background services is more complicated on docker as well, from my understanding, in principle each service should have it's own container, but it does all get quite complicated fast when you need a service for emoncms_mqtt, feedwriter, demandshaper, service-runner etc ... and then the web frontend is designed to work with services running as systemd which is not usually an option in a docker container.. or at least in the containers we've developed so far.

Is there a better way (or at least more practical way) to run these background scripts and services within a docker environment and then manage them from the frontend I wonder?

TrystanLea:
I probably need to get some time to look at the docker side of things again...

PR #1774, opened 2021-12-26
Fix failure in MQTT import when Redis is not enabled

jamesfidell:
Prevent use of Redis for obtaining data to publish to MQTT when Redis is not enabled

Also, add comments to phpmqtt_input.php to indicate that it is deprecated in favour of emoncms_mqtt.php


TrystanLea:
Thanks!

PR #1775, opened 2022-01-09
Suggest the user asks for less data

MyForest:
I believe we should encourage the user to request less data points by reducing the time window.


TrystanLea:
Thanks @MyForest

PR #1777, opened 2022-01-14
Php 8.1 fix   part1

thib66:
Changes have been done to : 
route.php
locale.php
core.php
user_controller.php
test if variables are defined to null before calling functions using those variables as parameters (Passing null to parameter #1 ($string) of type string is deprecated in php 8.1)
replace  FILTER_SANITIZE_STRING with htmlspecialchars() (Constant FILTER_SANITIZE_STRING is deprecated )

### DIFF
--- a/Lib/EmonLogger.php
+++ b/Lib/EmonLogger.php
@@ -92,7 +92,7 @@ class EmonLogger
         }
 
         $now = microtime(true);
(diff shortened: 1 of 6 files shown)


borpin:
Does this merge work with existing installs?

TrystanLea:
Yes seems fine

PR #1779, opened 2022-02-04
mqtt: support nested JSON values

ramcq:
JSON values from ebusd are published on MQTT with values "boxed" in a
format like: { name: xxx, value: xxx }

Support this structure by allowing deeper-nested JSON structures, and
unpacking the value and adding the name if present.

Ensure non-scalar values (nested JSON, etc) are not recorded, or logged
without being converted back to JSON.

### DIFF
--- a/scripts/services/emoncms_mqtt/emoncms_mqtt.php
+++ b/scripts/services/emoncms_mqtt/emoncms_mqtt.php
@@ -252,7 +252,7 @@
             $topic = str_replace(":","",$topic);
 
             //Check and see if the input is a valid JSON and when decoded is an array. A single number is valid JSON.
-            $jsondata = json_decode($value,true,2);
+            $jsondata = json_decode($value,true,3);
             if ((json_last_error() === JSON_ERROR_NONE) && is_array($jsondata)) {
                 // JSON is valid - is it an array
                 $jsoninput = true;
@@ -279,7 +279,7 @@
                             $time = $timestamp;
                         }
                     } else {
-                        $log->warn("Time value not valid ".$inputtime);
+                        $log->warn("Time value not valid ".json_encode($inputtime));
                         $time = time();
                     }
                 } else {
@@ -291,7 +291,7 @@
                 $time = time();
             }
 
-            $log->info($topic." ".$value);
+            $log->info($topic." ".($jsoninput ? json_encode($jsondata) : $value));
             $count ++;
             
             $inputs = array();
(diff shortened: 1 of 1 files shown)


scripts/services/emoncms_mqtt/emoncms_mqtt.php
```diff
@@ -279,7 +279,7 @@ function message($message)
                             $time = $timestamp;
                         }
                     } else {
-                        $log->warn("Time value not valid ".$inputtime);
+                        $log->warn("Time value not valid ".json_encode($inputtime));
```
borpin, line 282: What difference does the `.json_encode` make?

ramcq: Because $inputtime is derived from json_decoded values it can be a non-scalar type, an array or whatever, and attempting to print it was killing the whole script.

scripts/services/emoncms_mqtt/emoncms_mqtt.php
```diff
@@ -334,7 +334,19 @@ function message($message)
...
                     foreach ($jsondata as $key=>$value) {
-                        $inputs[] = array("userid"=>$userid, "time"=>$time, "nodeid"=>$nodeid, "name"=>$input_name.$key, "value"=>$value);
+                        // Unbox { name: xxx, value: xxx }
+                        if (is_array($value) && array_key_exists("value", $value)) {
+                            if (array_key_exists("name", $value) && $value["name"])
+                                $key .= '_'.$value["name"];
+                            $value = $value["value"];
+                        }
+
+                        if (is_scalar($value)) {
```
borpin, line 344: I quite like this check, but might I suggest it should check for a number instead?

borpin:
I am trying to think where the acceptable JSON structure is documented. It may not be, as it is currently the same for the HTTP interface as well as the MQTT interface.

To be consistent, you may need to update the HTTP interface for `full_json` as well.

I did some work on this a while back as originally, it did not accept valid JSON.

borpin:
Can we presume you have tested this with 2 level JSON so it doesn't break anything :)

borpin:
Yup, input API here

https://github.com/emoncms/emoncms/blob/d9264eca2ba3181168eee95d35a28c38cb682436/Modules/input/input_methods.php#L98

ramcq:
> Can we presume you have tested this with 2 level JSON so it doesn't break anything :)

*ahem* **taptaptaptap** Yes that appears to work :)

scripts/services/emoncms_mqtt/emoncms_mqtt.php
borpin: Yes true and yes sort of dropped later, but often create bad inputs as a result so ensuring only numbers are sent on would be a really good thing (but could be proposed as a separate PR).

borpin:
> _ahem_ **taptaptaptap** Yes that appears to work :)

My life resolves around "Assume nothing"...

borpin:
> I am trying to think where the acceptable JSON structure is documented

Opened an issue #1780

scripts/services/emoncms_mqtt/emoncms_mqtt.php
ramcq: Would be nice to know if there's a possible approach for keeping the strings in the pipeline, if we could add a feed function to do matching or mapping somehow...?

ramcq:
> Yup, input API here
> 
> https://github.com/emoncms/emoncms/blob/d9264eca2ba3181168eee95d35a28c38cb682436/Modules/input/input_methods.php#L98

Happy to update this but (and perhaps giving us one place to decode "truthy" string values and rejecting others) these two close-but-no-cigar functions are ideally looking for a couple of utility functions for fetching time and numeric values from submitted strings. Is there somewhere sensible to factor those out to?

borpin:
> Is there somewhere sensible to factor those out to?

@TrystanLea will need to comment on that.

ramcq:
Ping @TrystanLea - looking at this again (trying to remember and make it actually work...!) - is there somewhere I can move the JSON-to-time-and-number functions to so they are not duplicated between the MQTT code and the HTTP input API?

TrystanLea:
Finally merged 3 years later! ChatGPT helped me add some notes and extend the test script 

https://github.com/emoncms/emoncms/blob/master/scripts/services/emoncms_mqtt/emoncms_mqtt.php#L26

https://github.com/emoncms/emoncms/blob/master/scripts/examples/mqtt_input_publisher.php

image

PR #1809, opened 2022-08-12
Update .htaccess

CidiRome:
Correction to file to support Let's Encrypt Certificate generation using getssl script.


borpin:
Can you explain & provide instructions on the script and how to use it, please?

CidiRome:
The .htaccess file is not a script.
This file contains configurations for the web server to interpret.
The first line I made changes was a correction because the regular expression, as it was, would only include that exact string and we need it to include any address started by that string.
And the second line was to exclude that folded from the address exclusions that are made in that part of the configuration.

PR #1824, opened 2022-11-10
Update feed_model.php: For avoid to reset PowerToKWh process feed

And2IT:
Here is the original @nicsergio proposed modification:
https://community.openenergymonitor.org/t/correcting-cumulative-kwh-reset-with-post-process-module/4631/11

to resolve issue of loosing cumulative values in PowerToKwh feed.


borpin:
Duplicate of #1632 - please close

PR #1826, opened 2022-12-02
divide added to the process list

Chrischi-:
Divide added to the process list – Sometimes it can be useful to use divide directly instead of multiplication


Bra1nK:
Might as well add subtract as well if we're going down this road

Chrischi-:
We should also check all the functions in the file `"Modules/process/process_processlist.php"`.
Here, many of them are taking `$time` as an argument, but there is no other usage in the function 
or in the return of it. I think that this is a copy-and-paste leftover.

PR #1827, opened 2022-12-15
Use floor/ceil for datastart/end on SimpleZoom khwd graph

tstabrawa:
* Based on calculations done in bargraph.php
* Fixes ability to draw graphs with delta = On.

Tested locally.  Without fix, kwhd graph is blank when delta = On.  With fix, graph is shown correctly.


TrystanLea:
Cheers

PR #1829, opened 2022-12-17
Fix false to array deprecation in feed/fetch.json

townsen:
* This occurs using PHP 8.1 when the returned JSON begins with: *
* <br />
* <b>Deprecated</b>:  Automatic conversion of false to array is deprecated in
* <b>/var/www/emoncms/Modules/feed/feed_controller.php</b> on line <b>94</b>
* <br /> *


TrystanLea:
Thanks Nick!

PR #1838, opened 2023-02-04
Update upgrade SD Card information

borpin:
Update the upgrade information add photo of emonPi SD Card location.

Make clearer this is the route to Upgrade the SD Card.


TrystanLea:
Thanks !!

PR #1849, opened 2023-03-05
Fix timepicker not working outside graph module 

chaveiro:
Fix timepicker not working on all modules except graph.

Please don't change global function names that affect other modules without testing.  

The original function name used across emoncms vis and tables date time picker is *parse_timepicker_time* not *parseTimepickerTime*. Made this work for all but please correct the graph module to use the correct (historic) function name.


TrystanLea:
thanks!

PR #1850, opened 2023-03-06
Fix deprecation warnings from PHP8+

mattheworres:
Found that these deprecated string interpolations tend to cause some rather scary looking errors re:headers output when running on PHP 8.2, and didn't think it would harm anyone still running on PHP7.


TrystanLea:
Cheers

PR #1851, opened 2023-03-06
Remove deprecated FILTER_SANITIZE_STRING in feed_model and user_controller

mattheworres:
Another set of issues from deprecations I stumbled upon while running my own instance of Emoncms on PHP 8.2 -- seems `FILTER_SANITIZE_STRING` was ambiguous and it is thus frowned upon from using it now, `htmlspecialchars()` seems to be the best replacement for most instances.

There was one other instance of `FILTER_SANITIZE_STRING` in the codebase, but it seemed to be far more complicated than these cases - multiple other flags were likely passed to `filter_var` and thus, I wasn't confident `htmlspecialchars()` was a drop-in replacement.

I was able to test the user controller change on my instance (its in the logout action), but I don't have any feed data just yet, so I'm flying a bit blind there.

### DIFF
--- a/Modules/feed/feed_model.php
+++ b/Modules/feed/feed_model.php
@@ -848,7 +848,7 @@ class Feed
 
         if (isset($fields->name)) {
             //remove illegal characters
-            $fields->name = trim(filter_var($fields->name, FILTER_SANITIZE_STRING));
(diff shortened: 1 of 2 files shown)


TrystanLea:
Thanks again!

PR #1852, opened 2023-03-10
Fix inputs screen for PHP8 re:utf8_encode

mattheworres:
Inputs screen spins forever because of a deprecation warning about `utf8_encode`. On my local running instance of emoncms on PHP8.2, this fixed the issue, but it's based off of the Docker devbox. Still, I assume `mbstring` should be available for most if not all installs.


Tifaifai (approved):

TrystanLea:
Thanks!

PR #1860, opened 2023-08-17
Minor code cleanup

reedy:
* else if -> elseif
* Add some braces
* Remove no-op ;
* Use ?: where possible
* Swap join() for implode()
* Call implode() with correct parameter order
* Simplify some conditionals
* Use PHP_VERSION rather than phpversion()
* Trim trailing whitespace from files edited
* Use https:// for relevant urls


TrystanLea:
Thanks @reedy !

PR #1862, opened 2023-08-17
Update composer.json

reedy:
* Add ext-mbstring and ext-zlib to require
* Add cassandra, mosquitto-php and redis extensions to suggest

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,11 +16,18 @@
     },
     "license": "AGPL-3.0-or-later",
     "require": {
-        "ext-mysqli": "*",
+        "ext-ctype": "*",
+        "ext-curl": "*",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -16,11 +16,18 @@
...
+        "ext-ctype": "*",
+        "ext-curl": "*",
         "ext-gettext": "*",
         "ext-json": "*",
-        "ext-curl": "*",
+        "ext-mbstring": "*",
+        "ext-mysqli": "*",
         "ext-openssl": "*",
-        "ext-ctype": "*"
+        "ext-zlib": ""
```
reedy, line 26: Bah. Missing '*' for the version.

PR #1876, opened 2024-04-04
Use more common -B for specifyng block size for df

teadur:
fix emoncms_1  | df: unrecognized option: block-size=1 when running emoncms on alpine.


TrystanLea:
thanks

PR #1878, opened 2024-04-13
to support home-assistant ingress mode :-)

alexandrecuer:
core.php has to be modified when the web server is running in ingress mode, for example inside [homeassistant](https://www.home-assistant.io/blog/2019/04/15/hassio-ingress/)

when running in ingress mode, apache has got some specific headers we have to use in the path

image

when running out of ingress mode, these headers are no more present

image

### DIFF
--- a/core.php
+++ b/core.php
@@ -45,7 +45,13 @@ function get_application_path($manual_domain=false)
     }
 
     if (isset($_SERVER['HTTP_X_FORWARDED_HOST'])) {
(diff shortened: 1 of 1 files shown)


alexandrecuer:
emoncms inside homeassistant in ingress mode 

image

image

alexandrecuer:
emoncms running in ingress mode inside HA

image

alexandrecuer:
the same running container opened in the classic way (exposing a port)

image

TrystanLea:
Thanks @alexandrecuer

PR #1880, opened 2024-04-18
Fix last update time for inputs and feeds

chaveiro:
- Fix report of last updated input and feed.
- Refactor to use same logic for both in resposive-linked-tables.js shared lib.
- Added new feature to report month/day of last update time when update was more than 1 month ago in gray color. Useful for easy understanding time for disabled inputs or feeds.
- Fix consider server time out of sync with timeServerLocalOffset var.
- Fix input table first line offset.

Screenshot:
image

### DIFF
--- a/Lib/responsive-linked-tables.js
+++ b/Lib/responsive-linked-tables.js
@@ -40,7 +40,7 @@ $(function() {
     //data-toggle="collapse"
 
     $(document).on("shown hidden", '#table .tbody.collapse', function(e){
-    	$header = $('[data-target="#'+e.target.id+'"]')
+        $header = $('[data-target="#'+e.target.id+'"]')
         if (e.type == 'hidden'){
             $header.addClass('collapsed')
             $header.find('.icon-indicator').removeClass('icon-chevron-down').addClass('icon-chevron-right')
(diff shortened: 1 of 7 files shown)


TrystanLea:
Hello @chaveiro

Really like the grey date replacement for inactive! that's a nice touch.

It's a shame to loose the 'feedListItemIntervalClass' functionality with this update.

I have as an example a whole load of 60s temperature and humidity feeds and then lots of 10s feeds. Previously both of these would remain green as their update rate is within the feed interval but trying your branch the 60s update feeds now appear yellow to suggest they are a bit slow coming in. 

What's the best way to test the server time offset fix?

TrystanLea:
I will have a look at this and reinstate the feed colour based on interval in a different way whilst keeping all of your changes

chaveiro:
Hi Trystan, i was thinking at this today. Didn't understood the feeds implementation but it's clear now with your comment.

How about when it's a feed, consider the colors severity as an offset of the feed interval?

Something like this: ex. if last update was 30s but the feed interval is 60s, it only considers not green after feed interval + 30s (90s)
That way it can all be done on the same list_format_updated_obj function by adding a new optional arguments that is the feed interval.

TrystanLea:
Thanks @chaveiro yes agreed re adding this into same list_format_updated_obj, that's what I've done. Now all merged in. Here's the commit change that I did https://github.com/emoncms/emoncms/commit/24a7d4734b7ec92f65a165e0604253823c4d4684

I've moved things around a bit in feedlist_view.php, the html for the node row is now built after building the section for the node's feeds. The colour for the node,  last updated value and sum of feed sizes can then be selected from the feeds without addition loops and calculations.. 

The colour code for the node level is based on the highest colour code in the feeds... this is a slightly different approach. Lets say there are two feeds in the node, one is 10s data and has not been updated in 24 intervals (colour code red). The other is daily data that is only running behind 2 intervals (colour code green). The node colour and time will be based on the 10s feed that is more unusually behind than the daily feed.. 

There might be a better way of doing that Im not sure..

chaveiro:
Minor fixes here: https://github.com/emoncms/emoncms/pull/1882

PR #1881, opened 2024-04-30
Publish 5 minute mqtt msg process count to MQTT 

alandpearson:
Publish 5 minute mqtt msg process count to mqtt so can be seen as input in emoncms
This allows users to easily see performance and any variances in mqtt volume. Also changed the log level for this to WARN so it can easily be seen in the logfile without drowning in debug data.
This is a very useful stat to see (and log optionally log as a feed) and helped me debug several problems with my installation (run away rfm69 module for example).

### DIFF
--- a/scripts/services/emoncms_mqtt/emoncms_mqtt.php
+++ b/scripts/services/emoncms_mqtt/emoncms_mqtt.php
@@ -199,7 +199,9 @@
 
         if ((time()-$last_heartbeat)>300) {
             $last_heartbeat = time();
-            $log->info("$count Messages processed in last 5 minutes");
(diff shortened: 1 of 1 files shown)


chaveiro:
As its not a warning, why log it as a warning?

alandpearson:
There only seems to be 3 levels of logging supported, and  'INFO' drowns the log file in mqtt messages meaning this useful message gets lost.

TrystanLea:
Thanks @alandpearson I can see the useful intention of this, perhaps it would be good to have the option to enable this as a setting in settings.ini? I can see some users finding this input and being confused even if useful for those who know.
Cheers

PR #1885, opened 2024-06-02
mac address broadcast through API admin/getmac route

alexandrecuer:
It is vendor responsibility to provide a unique identifier

It is asked for example by home-assistant when using the emoncms module synchronizing an existing emoncms installation to home-assistant core

broadcasting the mac address through the API is an accepted solution : cf https://developers.home-assistant.io/docs/entity_registry_index/

### DIFF
--- a/Modules/admin/admin_controller.php
+++ b/Modules/admin/admin_controller.php
@@ -134,6 +134,12 @@ function admin_controller()
     // ----------------------------------------------------------------------------------------
     // System info page actions
     // ----------------------------------------------------------------------------------------
+    if ($route->action == 'getmac') {
+        // return server mac address
+        $route->format = 'text';
+        $cmd = "ifconfig -a | grep -o -E '([[:xdigit:]]{1,2}:){5}[[:xdigit:]]{1,2}'";
+        return str_replace("\n", "", shell_exec($cmd));
+    }
 
     if ($route->action == 'service') {
         $route->format = 'json';
--- a/Modules/admin/admin_controller.php
+++ b/Modules/admin/admin_controller.php
@@ -138,7 +138,9 @@ function admin_controller()
         // return server mac address
         $route->format = 'text';
         $cmd = "ifconfig -a | grep -o -E '([[:xdigit:]]{1,2}:){5}[[:xdigit:]]{1,2}'";
-        return str_replace("\n", "", shell_exec($cmd));
+        $result=shell_exec($cmd);
+        if (!isset($result)) return "UNKNOWN_DEVICE";
+        return str_replace("\n", "", $result);
     }
 
     if ($route->action == 'service') {


alexandrecuer:
Complementary approach could be to include a unique identifier based on the mac adress for each feed in the feed/list.json route :

`{MAC_ADDRESS}_{USER_ID}_{FEED_ID}`

borpin:
Does a Docker install have a unique MAC address? i.e. if you have 2 Docker installs on a single machine, do they return different UIDs?

This could be expanded to a health/system status via API.

alexandrecuer:
Yes a docker install shares the MAC Address with the host. So if you share 2 Docker installations on the same machine, you should have the same MAC address for the 2 installs, and so the same UID except if you have a server with different cards.

But I dont see why would you want to have 2 docker containers running emoncms on the same machine ?

borpin:
You might be doing some testing and have 2 instances running?

> and so the same UID

My point is the MAC address is not *guaranteed* to be a UUID. I wonder if the install should generate something that is more likely to be a UUID?

alexandrecuer:
Good point :-) you are very right,  there is indeed a big problem in my approach

docker start assigning always the same mac 02:42:ac:11:00:02 for the first container and then is increasing by one each mac for each different container.

image

```
alexandrecuer@alexandrecuer-ThinkPad-X13-Gen-1:~$ sudo docker exec -it 69aecfe063fa sh
[sudo] Mot de passe de alexandrecuer : 
/opt/openenergymonitor # ifconfig
eth0      Link encap:Ethernet  HWaddr 02:42:AC:11:00:03  
          inet addr:172.17.0.3  Bcast:172.17.255.255  Mask:255.255.0.0
          UP BROADCAST RUNNING MULTICAST  MTU:1500  Metric:1
          RX packets:208 errors:0 dropped:0 overruns:0 frame:0
          TX packets:161 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:0 
          RX bytes:53949 (52.6 KiB)  TX bytes:366634 (358.0 KiB)

lo        Link encap:Local Loopback  
          inet addr:127.0.0.1  Mask:255.0.0.0
          UP LOOPBACK RUNNING  MTU:65536  Metric:1
          RX packets:2847 errors:0 dropped:0 overruns:0 frame:0
          TX packets:2847 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:1000 
          RX bytes:170421 (166.4 KiB)  TX bytes:170421 (166.4 KiB)
```


image
```
alexandrecuer@alexandrecuer-ThinkPad-X13-Gen-1:~$ sudo docker container exec -it dac4946a7f4f sh
/opt/openenergymonitor # ifconfig
eth0      Link encap:Ethernet  HWaddr 02:42:AC:11:00:02  
          inet addr:172.17.0.2  Bcast:172.17.255.255  Mask:255.255.0.0
          UP BROADCAST RUNNING MULTICAST  MTU:1500  Metric:1
          RX packets:49731 errors:0 dropped:0 overruns:0 frame:0
          TX packets:45983 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:0 
          RX bytes:18885699 (18.0 MiB)  TX bytes:142426846 (135.8 MiB)

lo        Link encap:Local Loopback  
          inet addr:127.0.0.1  Mask:255.0.0.0
          UP LOOPBACK RUNNING  MTU:65536  Metric:1
          RX packets:400393 errors:0 dropped:0 overruns:0 frame:0
          TX packets:400393 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:1000 
          RX bytes:264790548 (252.5 MiB)  TX bytes:264790548 (252.5 MiB)

```

borpin:
> there is indeed a big problem in my approach

I know nothing about Docker, so are you going to get the same UUID every time?

Secondly, If I move an instance to a different server/machine, the UUID will change which is not what we want!

alexandrecuer:
> My point is the MAC address is not _guaranteed_ to be a UUID. I wonder if the install should generate something that is more likely to be a UUID?

Proposition : 
- for containers, create at startup an ENV var named UUID with the bash command `uuidgen` or another way  cf https://www.uuidgenerator.net/dev-corner/bash
- if the install is not containerized, the ENV var does not exist and we go back to the mac address

alexandrecuer:
> I know nothing about Docker, so are you going to get the same UUID every time?

On the same machine, docker start assigning always the same mac 02:42:ac:11:00:02 for the first container and then is increasing by one each mac for each different container. So the second container will be 02:42:ac:11:00:03, the third 02:42:ac:11:00:04

So all running containers will have different UUID on the machine

But if you install another machine and run containers, the first container running will get the mac 02:42:ac:11:00:02, and so on....

When generating the MAC, docker always start with `02:42` and the remaining four octets represent the IP address of the container in hex

alexandrecuer:
> Secondly, If I move an instance to a different server/machine, the UUID will change which is not what we want!

Agree, you mean moving containers from a host to another ? I was not aware this was possible

borpin:
> you mean moving containers from a host to another

No I mean moving via an export/import. If the UUID is in the SQL database, then you can move to a new Pi (or any platform) and HA will treat it the same (as it should).

alexandrecuer:
closing for something more clever

PR #1892, opened 2024-07-01
Add documentation related to unique identifier

alexandrecuer:
Describe the process to update the database in order to produce the unique identifier


alexandrecuer:
Thanks Trystan, 

Do i have something else to do so that it appears online on https://docs.openenergymonitor.org/emoncms/update.html

Maybe change the commit number on https://github.com/openenergymonitor/docs ?

TrystanLea:
It's manual process I'm afraid, just updated it for you

PR #1894, opened 2024-07-30
Spelling correction of the word visualization

cypherm0nk:
I corrected the spelling of some words in the comments.


alexandrecuer (approved):

reedy:
The primary developers are based in the UK... So it's not a surprise that they're using en-gb rather than en-us.

So changing the comments wholesale doesn't make sense, neither does the usages in line.

If these are wanted (and that's fair), appropriate `locale/en_US/LC_MESSAGES` files should be created with the translations.

reedy (changes requested):
See comment

PR #1896, opened 2024-08-20
Add missing MQTT settings for old config

Dash:
I found that these two settings are critical for MQTT support in EmonCMS, but if running on a legacy config format, they don't exist.

* userid
* multiuser

I don't believe these are used in a local installation, so have defaulted them to 1, and false respectively.


borpin:
The old PHP settings should be removed completely IMHO.

PR #1897, opened 2024-09-28
Minor bug fixes for feed module

MrTimbones:
A couple minor fixes found when upgrading to php8.2

1. if a user has a mix of MySQL TimeSeries and Memory feeds, emoncms can sometimes load the Memory engine before the main MySQL engine, resulting in an "this class already loaded" error. Fixed with `require_once`.

2. if a PHPTimeSeries somehow manages to have a _negative_ length, `fread` would error.


reedy (approved):

TrystanLea:
Thanks @MrTimbones

PR #1900, opened 2024-10-20
Fixed matching of settings in serial config view

mdr78:


reedy:
Do we have a min req browser verison etc? As much as they're old versions...

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith

TrystanLea:
Thanks @mdr78

Issue #1903, opened 2024-11-07
PHP Warning:  Undefined array key "hw" in /var/www/emoncms/Modules/admin/Views/admin_main_view.php on line 235


reedy:
https://github.com/emoncms/emoncms/blob/aee5e2f/Modules/admin/Views/admin_main_view.php#L235

borpin:
This is probably related to your previous issues I guess.

reedy:
It's not.

https://github.com/emoncms/emoncms/commit/6b4d93dc230c75da97d9c9c008c57a263f0d1c85 stopped it being set.

```
<!-- <?php echo row(_('SoC'), $rpi_info['hw']) ?> -->
```

^ HTML commenting it doesn't stop PHP executing it.

PR #1909, opened 2024-11-24
admin_main_view.php: Remove echo of $rpi_info['hw']

reedy:
It's not set anymore since https://github.com/emoncms/emoncms/commit/6b4d93dc230c75da97d9c9c008c57a263f0d1c85

Fixes https://github.com/emoncms/emoncms/issues/1903


Modules/admin/admin_model.php
```diff
@@ -617,8 +617,8 @@ public function is_Pi() {
     /**
      * get an array of raspberry pi properites
      *
-     * @see: SoC 'hw' now not required - https://github.com/emoncms/emoncms/issues/1364
-     * @return array has keys [hw,]rev,sn,model     * @return array has keys hw,rev,sn,model
```
reedy, line 621: The `@return` was duplicated and isn't helpful.

`now not required` is completely different to `not used ever`

alexandrecuer (approved):

TrystanLea:
Thanks!

PR #1911, opened 2024-12-23
Enable PI info, reboot, and shutdown for late model RPis

davidski:
Late model RaspberryPIs are using a new mac OUI (see https://maclookup.app/macaddress/2ccf67). 
These devices (Raspberry Pi 4B sourced from AdaFruit in my case) do not have the 
PI section enabled in admin and therefore can't reboot/shutdown.

This PR enabled this OUI. Reboot/shutdown works, though video temps do not (yet) 
work.

### DIFF
--- a/Modules/admin/admin_model.php
+++ b/Modules/admin/admin_model.php
@@ -611,7 +611,7 @@ class Admin {
      * @return bool
      */
     public function is_Pi() {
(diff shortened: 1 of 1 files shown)


TrystanLea:
Thanks @davidski

PR #1913, opened 2024-12-24
Adjust how serial, revision, and hardware are parsed out.

davidski:
Under 01Feb-24 versions of EmonSD, the serial and model information fail to parse 
for current RPi models. This PR adjusts the match logic to be a bit more flexible 
in how the regex gets assigned. Tested locally on a RPi4B.


TrystanLea:
Thanks @davidski

PR #1917, opened 2025-02-06
Adapt admin model for a correct behaviour in container

alexandrecuer:
This commits permits the admin module to work friendly in container without having to deal with systemctl commands or df, iostat or grep options (-x squashfs, -o JSON, -oP) that it cannot achieve

image
image

Still in containers, this also permits the sync module not to complain not to find emoncms_sync
image


TrystanLea:
Thanks @alexandrecuer

alexandrecuer:
Thanks @TrystanLea : if you can also check the prs in the postprocess module, it could be nice
Best

TrystanLea:
Thanks @alexandrecuer already merged, cheers!

PR #1920, opened 2025-03-18
Update .htaccess so that export can take more than 30secs

vdesmedt:
When exporting large feed, the request timeout after default value of 30secs.  This update extend the maximum execution time to 300secs


TrystanLea:
Thanks @vdesmedt

PR #1922, opened 2025-04-22
Enhanced processlist editor, new processes and minor fixs

chaveiro:
Enhanced processlist editor
- new process description for multi arguments processors
- copy, past
- dynamic vars form
- document vars
- small bugfixes

- Convert a power value in Watts to a feed that contains an entry for the total energy used every x min (starting mid night) (kWh/ x min)

Demo of the new processlist editor:
edit dash rezise


chaveiro:
@TrystanLea can you pleace accept this MR to move on?

TrystanLea:
Sorry, trying to work my way over to this. Hopefully will update soon!

TrystanLea:
First I would like to say that I agree that multi arg input processing has been a needed feature for emoncms for a while so thanks that!

I've added one of the new processes you have added and looking through the resulting processList field value in the mysql db:  E.g:

```
+-----------------------------------------------------------------------------------------------------------------------------------+
| processList                                                                                                                       |
+-----------------------------------------------------------------------------------------------------------------------------------+
|                                                                                                                                   |
| process__power_to_kwh_custom:eyJpbnRlcnZhbCI6IjE1IiwiZmVlZCI6IjE1MzgifQ==,process__log_to_feed_join:2095,4:2096,29:1564,47:10,24: |
|                                                                                                                                   |
|                                                                                                                                   |
+-----------------------------------------------------------------------------------------------------------------------------------+
```

When I first decided on the format for the processList I tried to keep it as short as possible with the view that the less characters to process the faster these processlists would get processed, especially at scale, but this of course created several key limitations: 

1. Needing to know the process id, making it difficult to set unique id's across modules
2. Only supporting a single argument

Later you introduced adding full process names in the field and that's been working fine, on emoncms.org as well. 

Now as you propose expanding on this with base64 encoded json. I wonder if we should switch to a single consistent format across the board. Store the whole processList as a human readable JSON. 

```
[
  {"process": "scale", "args": {"value": 100}},
  {"process": "log_to_feed", "args": {"feed": 2097}},
  {"process": "power_to_kwh_custom", "args": {"interval":"15", "feed":"1538}}
]
```

or: 

```
[
  {"process": "scale", "value": 100},
  {"process": "log_to_feed", "feed": 2097}
  {"process": "power_to_kwh_custom", "interval":"15", "feed":"1538}
]
```

Migration could be handled by supporting both formats for existing processes up to this point and then every time someone adds new input processing or edits an existing process list, the whole process list gets re-written to the new format at that point.

Format detection can be done by checking if the whole processList field is parsable JSON.

chaveiro:
Hi, 

2 - > apple-mobile-web-app-capable with mobile-web-app-capable 
There were some warning on chrome about this being deprecated hence the change, followed some guidelines i found on google.

3 -  Though about that but a change in the format involves changing a lot of places on the java script and UI, this was the best approach i could get on the time I could afford to do this. You can always decode the base64 and see the json that is there for troubleshooting.

4 - There are 2 for historical reasons only, to not break compatibility.

Like to add that all is tested, you can just merge it all and i will be welcome to help if anything arises from that later.

chaveiro:
Like to add also that the only code you didn't comment on flot lib is a fix to allow selecting and deselecting feeds with touch, that were not working.

TrystanLea:
I've just emailed you @chaveiro , I have some wider ideas on this, it would be great to catch up over video call as well!

Will update here for others benefit soon as well.

PR #1928, opened 2025-07-09
Improve PHPCS config, start some cleanup

reedy:
More pinging of https://github.com/emoncms/emoncms/issues/1336

Follows up https://github.com/emoncms/emoncms/pull/1585


Modules/feed/engine/MysqlMemory.php
```diff
@@ -2,12 +2,12 @@
 
 class MysqlMemory extends MysqlTimeSeries
 {
-    public function create($feedid, $options)
+    public function create($feedid)
```
TrystanLea, line 5: will removal of 2nd param here cause an error?

scripts/process_langjs_builder.php
```diff
@@ -24,26 +24,26 @@ function get_js_file($dir){
...
-            $tr[]=$rr[0];
+         $lines = explode("\n", file_get_contents($file));
+         $tr=array();
+         foreach ($lines as $line){
+         $pos = strpos($line, '_Tr(');
+          if ($pos !== false) {
+                 $r = explode('_Tr("', $line);
+                 unset($r[0]);
+                 foreach ($r as $key=>$val){
+                $rr= explode('")', $val);
```
TrystanLea, line 35: indentation does not look right here, but is it just on github?

scripts/vis_langjs_builder.php
```diff
@@ -17,26 +17,26 @@ function get_js_file($dir){
...
-   if (count($tr)>0) {
-    $a=explode('/', $file);
-    $name=array_pop($a);
-    natcasesort($tr);
-    $translation[$name] = $tr;
-   }
+             }
+         $tr= array_unique($tr);
+         if (count($tr)>0) {
+     $a=explode('/', $file);
```
TrystanLea, line 35: again indentation is not quite right

TrystanLea:
Thanks @reedy I've added a few inline comments. 

Are you just running this command but on a small number of files for the above changes?

    phpcbf -p .

reedy:
I'm just running 'composer fix' which runs on everything, excluding vendor and Lib.

reedy:
The indenting issues is probably from files not having consistent (tabs or spaces) indenting, not to consistent levels. And without a specific config setting that, it may just be getting confused as part of doing other changes...

Seeing as it's not doing weird things on most of the files in the repo, we can probably just fix that by hand one time and should be good going forward.

Like I say, if we can decide whether we actually want tabs or spaces, getting that fixed should be easier.

Modules/feed/engine/MysqlMemory.php
reedy: `Generic.CodeAnalysis.UnusedFunctionParameter.FoundInExtendedClassAfterLastUsed` is the rule being flagged here...

TrystanLea:
Before we go any further with this, lets get json_i18b merged in first, seeing lots of merge conflicts

PR #1938, opened 2025-07-23
Add support to update device IP automaticaly

chaveiro:
Update to support add ip to device module on https://github.com/emoncms/device/pull/86


chaveiro:
Update to support add ip to device module.

Issue #1944, opened 2025-09-25
all tests are failing in th CI


alexandrecuer:
<img width="1844" height="774" alt="Image" src="https://github.com/user-attachments/assets/1fcfce97-0d2e-471a-8625-bcdabf065640" />

PR #1945, opened 2025-09-25
to solve ci failure

alexandrecuer:
should solve https://github.com/emoncms/emoncms/issues/1944


TrystanLea:
Thanks @alexandrecuer

Issue #1948, opened 2025-10-13
admin/setuser?id=1 fatal


reedy:
Using [11.9.10](https://github.com/emoncms/emoncms/releases/tag/11.9.10)

```
Warning: Undefined variable $user in /var/www/emoncms/Modules/admin/admin_controller.php on line 518

Fatal error: Uncaught Error: Call to a member function get() on null in /var/www/emoncms/Modules/admin/admin_controller.php:518 Stack trace: #0 /var/www/emoncms/core.php(81): admin_controller() #1 /var/www/emoncms/index.php(257): controller() #2 {main} thrown in /var/www/emoncms/Modules/admin/admin_controller.php on line 518
```

Last seemingly touched in https://github.com/emoncms/emoncms/commit/27cc63e439175dd36d2a74b31ba51dad4684ac3f

PR #1949, opened 2025-10-13
admin_controller: Use global $user

reedy:
Fixes https://github.com/emoncms/emoncms/issues/1948

Caused by 27cc63e439175dd36d2a74b31ba51dad4684ac3f


reedy:
This has been broken for nearly 3 months now...

Issue #1965, opened 2026-01-03
Only All/Space Heating go blue when selected


reedy:
If I select All/Space Heating, they go blue.

<img width="276" height="106" alt="Image" src="https://github.com/user-attachments/assets/8c672c30-07fd-4c7f-833d-80efd66054da" />
<img width="246" height="101" alt="Image" src="https://github.com/user-attachments/assets/db05b787-46ed-4765-8668-f7cd25d53433" />

When running and Water heating don't.

FF 146.0.1 on macOS 15.7.3

reedy:
https://github.com/user-attachments/assets/28303e46-d53e-437f-baaa-40cbe96bc7f0

PR #1966, opened 2026-01-03
svg_icons.svg: Remove fill from play, shower and snowflake icons

reedy:
Fixes https://github.com/emoncms/emoncms/issues/1965


TrystanLea:
Thanks @reedy

PR #1969, opened 2026-01-06
Document input name constraints in API documentation

swwilshub:
Add a prominent description box to the input API page that documents:
- Maximum input name length (64 characters)
- Allowed characters (letters, numbers, underscore, hyphen, period, whitespace)
- Behavior when invalid characters are used (silently removed)

This addresses the issue where input names that are too long or contain invalid characters would silently fail without clear documentation.


TrystanLea:
Thanks for this, it unfortunately also seems to add constraints to the code not just document them?

swwilshub:
Oh, I didn't realise it was just a length limit on .org. maybe just accept the information box?

